Issue #923, opened 2013-12-19
Enable https protocol in custom site binding


davorgrubelic:
It would be great if you could implement possibility to choose between HTTP and HTTPS protocol in custom site binding. This is specially good reason with service applications to increase service security.

Is there any reason why this isn't supported yet?

Happy coding... :)
@gissues:{"order":66.34615384615365,"status":"backlog"}

davidebbo:
Fixed by d42e4fbcbd64ba4f48895520849e8106d453ddcb

PR #1409, opened 2014-12-04
[RFC] Support for HTTPS Protocol, Certificates for IIS 7-8 and SNI for IIS 8

jeme:
This replaces the old PR: https://github.com/projectkudu/kudu/pull/1386

Requirements
----

- [x] Should be able to add custom HTTPS Bindings though the UI
- [x] Should be able to choose a certificate when adding a custom HTTPS Bindings though the UI
- [x] Should be able to choose SNI on IIS 8.x
- [x] Should not be able to choose SNI on IIS 7.x
- [x] Should not not be able to set a host name on unless the Certificate is a wildcard cert (*.) or if SNI is enabled

The UI has changed significantly, I ended up choosing to mimic the UI in the IIS manager instead of the "Syntax" thing, this is because I had to add some additional fields anyways, like certificate and SNI on IIS 8, so I ended up remaking it completely for familiarity sake. (IIS Administrators should now be familiar with the fields)

- [x] Should be able to add define default HTTPS bindings
- [x] Should be able to configure certificate for default HTTPS bindings
- [x] Should be able to configure SNI for default bindings
- [x] Should be able to configure Certificate stores (relevant for IIS 8 as it allows you to choose multiple stores when adding certs, we could state that the cert HAS to be in the personal store for KUDU as an alternative)

I Changed the whole configuration concept to be a Configuration section, and added an abstraction on top of that that to keep backwards compatibility.

Default bindings are now quite flexible in the way that you can add any number of default bindings, currently only with http or https, but it should now be significantly easier to add support for other protocols if needed.

Tests for components:
----

If at all possible, there should be unit tests for the components mentioned below.

- [x] Kudu.SiteManagement.Certificates.CertificateLookup
- [x] Kudu.SiteManagement.Certificates.CertificateSearcher 
- [x] Kudu.SiteManagement.Configuration.KuduConfiguration
- [x] Kudu.SiteManagement.PathResolver

There is still a few missing things in this PR:
----

- [x] Move IP's and IISVersion to a IKuduContext or redesign the KuduEnvironment class to support those. They where added to the IKuduConfiguration for now as it was the easy route, but they don't belong there as they are not configuration values.
- [x] Cleanup the default `app.config` and refresh the documentation.
- [x] CLA Signed
- [x] Squash all commits into a single for easy of review.
- [x] Fix usage of Apps definition (seems broken).
- [x] Add explicitly named parameters where applicable (was removed)
- [x] Run code analysis and fix any issues.
- [x] Fix SNI Flag visibility in the UI (currently hard-coded to be visible for debugging, but should only be visible for IIS 8)

Please feel free to note other things missing.

CC: @n3rd, @davidebbo, @ahmelsayed, @suwatch (Those that participated in the old PR)

----
From the old PR:

We need to be able to define HTTPS bindings for our app's deployed with KUDU, so since I found #923 and that didn't seem to have much traction, I thought I would look into how this could possibly be added.

So this is a work in progress, but I wanted to see if I could get some feedback midway as I choose to change the configuration concept as well.

I am still figuring out how to deal wit certificates, as those become a thing to deal with when we suddenly wan't to do https.


Kudu.SiteManagement/SiteManager.cs
```diff
@@ -374,92 +390,93 @@ private static int GetRandomPort(IIS.ServerManager iis)
...
+                {
+                    switch (binding.Scheme)
+                    {
+                        case UriScheme.Http:
+                            site.Bindings.Add(binding.Binding, "http");
+                            break;
+
+                        case UriScheme.Https:
+                            Certificate cert = _certificateSearcher.Lookup(binding.Certificate).ByFriendlyName();
+                            Binding bind = site.Bindings.Add(binding.Binding, cert.GetCertHash(), cert.StoreName);
```
jeme: Agreed, so for now I changed it to throw an ConfigurationErrorsException with a more meaning-full message, it feels a bit odd to throw that here, but it's the most meaning-full thing I could come up with at the moment.

Kudu.SiteManagement.Test/Kudu.SiteManagement.Test.csproj
```diff
@@ -0,0 +1,72 @@
...
+  <ItemGroup>
+    <None Include="app.config" />
+    <None Include="packages.config" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\Kudu.SiteManagement\Kudu.SiteManagement.csproj">
+      <Project>{D5669C1D-3408-4CEE-8C1B-D86D03D27EE2}</Project>
+      <Name>Kudu.SiteManagement</Name>
+    </ProjectReference>
+  </ItemGroup>
```
jeme: Perhaps this is related to this: http://www.alteridem.net/2014/10/16/visual-studio-2013-modifies-nunit-project-files/

And since I am using ReSharper instead of the built-in test explorer I believe that is why I don't get that.

davidebbo:
Running build.cmd at the root on a clean repo, I get a bunch of errors from Code Analysis:

```
c:\code\GitHub\kudu\Kudu.SiteManagement\SiteManager.cs(184): error CA1811: Microsoft.Performance : 'SiteManager.BindingInformation.SiteType.get()' appears to have no upstream public or protected callers. [C:\code\GitHub\kudu\Kudu.SiteManagement\Kudu.SiteManagement.csproj]
c:\code\GitHub\kudu\Kudu.SiteManagement\SiteManager.cs(182): error CA1811: Microsoft.Performance : 'SiteManager.BindingInformation.Url.get()' appears to have no upstream public or protected callers. [C:\code\GitHub\kudu\Kudu.SiteManagement\Kudu.SiteManagement.csproj]
MSBUILD : error CA1726: Microsoft.Naming : Replace the term 'Flags' in type name 'SslFlags' with an appropriate alternate or remove it entirely. [C:\code\GitHub\kudu\Kudu.SiteManagement\Kudu.SiteManagement.csproj]
MSBUILD : error CA1010: Microsoft.Design : 'NamedElementCollection<T>' appears to be a non-generic, strongly typed collection. Add an implementation to the instance of ICollection<T> where T is the underlying element type of 'NamedElementCollection<T>'. [C:\code\GitHub\kudu\Kudu.SiteManagement\Kudu.SiteManagement.csproj]
MSBUILD : error CA1010: Microsoft.Design : 'BindingsConfigurationElementCollection' appears to be a non-generic, strongly typed collection. Add an implementation to the instance of ICollection<T> where T is the underlying element type of 'BindingsConfigurationElementCollection'. [C:\code\GitHub\kudu\Kudu.SiteManagement\Kudu.SiteManagement.csproj]
MSBUILD : error CA1010: Microsoft.Design : 'CertificateStoresConfigurationElementCollection' appears to be a non-generic, strongly typed collection. Add an implementation to the instance of ICollection<T> where T is the underlying element type of 'CertificateStoresConfigurationElementCollection'. [C:\code\GitHub\kudu\Kudu.SiteManagement\Kudu.SiteManagement.csproj]
```

Can you get it to a clean state? It probably only takes minor changes.

jeme:
Interesting... I am not getting those when running the build.cmd, so had to run the Analysis on the solution from VS to get those... Which seems odd when I don't have any changes to the build file...

But ofc... Will have a look at those.

jeme:
I don't have anything left on this one to do in my mind... So I am now waiting for comments if there is anything lacking.

The SiteManager could use some refactoring (also making it more testable), but that is outside the scope of the PR and would just generate allot of noise as I see it, so that would be better handled as a different PR IMO...

I have left the configuration elements out of testing as I A) don't really know how to make a meaningful test of them, and B) see it mostly as tests of the .NET framework.

davidebbo:
I think we were left on the build errors I showed above, which you said you were going to look at. Were you able to repro that?

jeme:
Those should have been fixed, at least I am not getting any errors when running code Analysis.

jeme:
I have updated the commit, I found out that I had set the "enableCustomHostNames" to true by default, which was not the original behavior.

davidebbo:
Ok, the change is finally in! I rebased it on latest, and I updated BCL packages to match the rest of the solution.

jeme:
Super. That should close the issue: https://github.com/projectkudu/kudu/issues/923 as I see it.

davidebbo:
Yep, just closed it.

PR #1428, opened 2015-01-06
Fixed WebPi feed

shrimpy:
Fixed WebPi feed dependencies; 
Update gitignore file to ignore WebPi log.

### DIFF
--- a/Setup/KuduSetupCustomWebPiFeed.xml
+++ b/Setup/KuduSetupCustomWebPiFeed.xml
@@ -1,8 +1,8 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <feed xmlns="http://www.w3.org/2005/Atom">
-  <version>1.0.0</version>
+  <version>1.0.1</version>
   <title>Feed to setup Kudu for various scenarios.</title>
-  <updated>2013-5-06T01:00:00Z</updated>
+  <updated>2015-1-05T01:00:00Z</updated>
   <author>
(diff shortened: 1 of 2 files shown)


suwatch:
Looks good - I don't have a new VM to actually try it.   BTW, we have 3 envs in play.

- local machine
- kudu-ci
- antares (Kudu2 and Kudu3)

Ideally we would love to keep them the same - so what we test locally will highly likely work the same on Antares.   However, it seems hard given Antares moves at slower pace than any.   At least, we should try to keep local machine and Kudu-CI the same.  Thought?

davidebbo:
Agreed that local machine and Kudu-CI should be the same. I think the script probably works for both.

Note that this script has been pretty broken for a while, so the bar to make changes to it is not very high!

Setup/KuduSetupCustomWebPiFeed.xml
```diff
@@ -212,4 +208,123 @@
       </installer>
     </installers>
   </entry>
-</feed>
\ No newline at end of file
+  <!-- entry for iisnode locked to a known version -->
+  <entry>
+    <productId>iisnode_0_2_11</productId>
```
davidebbo: So the reason for this is that the standard WebPI feed doesn't have those newer entries?

shrimpy: correct. it has the old version from 2012, and the old version is depends on Node.js 0.8

PR #1430, opened 2015-01-06
fix #1415 suppress spamming traces

### LINKED ISSUE #1415: Suppress certain APIs that return 200 from trace.
While investigating other issues, I noticed that the below traces spamming the log causing the more useful ones to be purged out.   These traces were from Portal polling requests.

```
2014-12-09T13-05-34_35a1b7_003_GET_api-webjobs_200.xml
2014-12-09T13-05-39_35a1b7_004_GET_api-siteextensions_200.xml
2014-12-09T13-06-36_35a1b7_006_GET_api-webjobs_200.xml
2014-12-09T13-06-41_35a1b7_007_GET_api-siteextensions_200.xml
2014-12-09T13-07-36_35a1b7_009_GET_api-webjobs_200.xml
2014-12-09T13-07-43_35a1b7_010_GET_api-siteextensions_200.xml
```

The ultimate fix is to implement etag on Kudu and fix the portal to consume etag e2e (like `/deployments`).   

The short term would be to suppress these specific APIs HTTP Get if 200 from the trace.   The list of api are ...

```
/api/webjobs
/api/siteextensions
/api/processes
```

There are a couple of twists we can add.

- Only suppress if call latency is less than 10s.    We want see traces for those taking longer time.
- Only suppress if call from Portal (based on User-Agent).   However, for `/api/webjobs` called was done thru CSM and Geo via promoted route and Geo do not propagate User-Agent!   Hence, it might not be possible till that is fixed.

suwatch:
Implemented per #1415 discussion.

### DIFF
--- a/Kudu.Services.Web/App_Start/NinjectServices.cs
+++ b/Kudu.Services.Web/App_Start/NinjectServices.cs
@@ -543,6 +543,11 @@ namespace Kudu.Services.Web.App_Start
             {
                 string appData = System.Environment.GetEnvironmentVariable("APPDATA");
                 FileSystemHelpers.EnsureDirectory(Path.Combine(appData, "npm"));
+
+                // discovered while adding npm 2.1.17
+                // this is to work around below issue with the very first npm install 
+                // npm ERR! uid must be an unsigned int
+                FileSystemHelpers.EnsureDirectory(Path.Combine(appData, "npm-cache"));
             }
             catch
             {


Kudu.Core/Tracing/XmlTracer.cs
```diff
@@ -175,6 +187,24 @@ private void WriteEndTrace()
...
+            }
+
+            // for api taking long time (>10s), always trace.
+            if (elapsed >= ElapsedThreshold)
+            {
+                return true;
+            }
+
+            // filter out those that may be polled by portal.
+            return !TraceFilters.Any(filter => file.IndexOf(filter, StringComparison.OrdinalIgnoreCase) > 0);
```
davidebbo: Seems a bit fragile to be filtering based on the complex file naming pattern. e.g. if we decide to tweak our file names, all the filters would need to be updated.

davidebbo:
Nice. :shipit:!

PR #1433, opened 2015-01-09
Improve setup and build script to minimize manual action.

shrimpy:
Improve setup and build script, so that for a new machine, we just need to manually install web platform install and has kudu code clone, the rest will be handled by script.


davidebbo:
Seems your editor is using tab characters instead of spaces. We use spaces everywhere in Kudu. Could you convert and push new commit? Thanks!

shrimpy:
oh.. didn`t realize that. will fix it. thanks

Issue #1441, opened 2015-01-21
[Documentation] Deploying to a server page, incorrect path for Kudu.Services.Web


juanpescador:
On the [Deploying to a server](https://github.com/projectkudu/kudu/wiki/Deploying-to-a-server) page, the path for Kudu.Services.Web is missing wwwroot\\. It currently says C:\inetpub\Kudu.Services.Web, but it should be C:\inetpub\wwwroot\Kudu.Services.Web.

I'd also suggest an edit to make it clearer that the directory has to be created/renamed. Something along the lines of:

> Copy the `artifacts\Release\SiteExtensions\Kudu` directory to `C:\inetpub\wwwroot` and rename it from `Kudu` to `Kudu.Services.Web`

davidebbo:
`\inetpub\Kudu.Services.Web` is actually correct, because by default, Kudu.Web's web.config has (not the `..`):

    <add key="serviceSitePath" value="..\Kudu.Services.Web" />

I did reword the sentence to clarify the rename.

juanpescador:
Sorry, I didn't give enough context. I initially placed the directory in C:\inetpub as per the documentation and the Admin Settings page reported an error:

![image](https://cloud.githubusercontent.com/assets/1808374/5853198/33fef52e-a221-11e4-8a20-71fd8c81f96b.png)

Moving Kudu.Services.Web to wwwroot\ and refreshing the Admin Settings page fixed the error, which is what brought me to suggest the change.

Giving it some thought, I realised maybe it's an issue of development vs deployment. For development (F5 in Visual Studio) it works: each project has a directory under the repo's root, so ..\Kudu.Services.Web from Kudu.Web\ is correct.

For deployment, though, following the documentation results in:

    C:\inetpub\
      - Kudu.Web.Services\
      [...]
      - wwwroot\
        - KuduWeb\

So, from inside KuduWeb\, traversing up one level takes you to wwwroot\ rather than inetpub\\.

juanpescador:
This pull request transforms Kudu.Web's web.config when executing build.cmd so the paths for Kudu.Web.Services\ and apps\ match those described in the documentation. Tested locally and works, all tests passing except four in Kudu.FunctionalTests:

[...]

PR #1444, opened 2015-01-22
Fix serviceSitePath and sitesPath for Release configuration.

juanpescador:
As discussed in #1441 directory structure for deploying to a server is different from developing/debugging with Visual Studio.

Use XML transformation to have the keys in Web.config point to the correct paths. When building with Release configuration they are one level higher. This is because Release is used when deploying Kudu to a server, so Kudu.Services.Web\ and apps\ are located in e.g. C:\inetpub\ while KuduWeb\ is in e.g. C:\inetpub\wwwroot\.


juanpescador:
Was actually a misunderstanding on my part of the directory structure proposed by documentation.

PR #1445, opened 2015-01-28
Add ARM support for WebJobs api.

amitapl:
Also add settings property to WebJob.

### DIFF
--- a/Kudu.Services/Arm/ArmActionSelector.cs
+++ b/Kudu.Services/Arm/ArmActionSelector.cs
@@ -1,0 +1,25 @@
+using System.Net.Http;
+using System.Web.Http.Controllers;
+using System.Web.UI;
+
+namespace Kudu.Services.Arm
+{
+    public class ArmActionSelector : ApiControllerActionSelector
+    {
+        public override HttpActionDescriptor SelectAction(HttpControllerContext controllerContext)
+        {
+            // This method selects which action method on the controller to invoke for any specific request.
+            // In this case if the request is a PUT and it is an ARM request we will direct it to the intended
+            // method name adding an "Arm" to it.
+            // For example if the route "/api/continuouswebjobs/myjob" is routed to action/method "CreateContinuousJob"
+            // then the ARM version of that method will be "CreateContinuousJobArm" and that is what is invoked when
+            // the condition applies.
+            if (controllerContext.Request.Method == HttpMethod.Put && ArmUtils.IsArmRequest(controllerContext.Request))
+            {
+                controllerContext.RouteData.Values["action"] += "Arm";
+            }
+
+            return base.SelectAction(controllerContext);
+        }
+    }
+}
--- a/Kudu.Services/Arm/ArmControllerConfiguration.cs
+++ b/Kudu.Services/Arm/ArmControllerConfiguration.cs
@@ -1,0 +1,25 @@
+using System;
+using System.Web.Http.Controllers;
+
+namespace Kudu.Services.Arm
+{
+    /// <summary>
+    /// Use this attribute on controllers that support the ARM api.
+    /// In these controllers, for each method that serves a PUT request you'll need to also implement
+    /// a method with the same name + an "Arm" postfix and the input will have an Arm envelope.
+    /// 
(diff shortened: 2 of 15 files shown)


Kudu.Services/Arm/ArmActionSelector.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Kudu.Services.Arm
+{
+    public class ArmActionSelector : ApiControllerActionSelector
+    {
+        public override HttpActionDescriptor SelectAction(HttpControllerContext controllerContext)
+        {
+            if (controllerContext.Request.Method == HttpMethod.Put && ArmUtils.IsArmRequest(controllerContext.Request))
+            {
+                controllerContext.RouteData.Values["action"] += "Arm";
```
davidebbo: Comment would be good here is this looks pretty cryptic without context.

Kudu.Services/Arm/ArmControllerConfiguration.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Web.Http.Controllers;
+
+namespace Kudu.Services.Arm
+{
+    [AttributeUsage(AttributeTargets.Class)]
+    public sealed class ArmControllerConfiguration : Attribute, IControllerConfiguration
+    {
+        public void Initialize(HttpControllerSettings controllerSettings, HttpControllerDescriptor controllerDescriptor)
+        {
+            controllerSettings.Services.Replace(typeof(IHttpActionSelector), new ArmActionSelector());
```
davidebbo: Should that use a singleton instance?

davidebbo: Well, object looks cheap so it may not matter.

Kudu.Services/Arm/ArmUtils.cs
```diff
@@ -0,0 +1,108 @@
...
+
+                armEntry.Name = String.Empty;
+                for (int i = 8; i < idTokens.Length; i += 2)
+                {
+                    armEntry.Name += "/" + idTokens[i];
+                }
+            }
+
+            IEnumerable<string> values;
+            if (request.Headers.TryGetValues(GeoLocationHeaderKey, out values))
```
suwatch: it gets this far, it must have GeoLocationHeader.   I would skip TryGetValue.

Kudu.Services/Arm/ArmUtils.cs
```diff
@@ -0,0 +1,108 @@
...
+        private static ArmEntry<T> Create<T>(T o, HttpRequestMessage request, bool isChild = false) where T : INamedObject
+        {
+            var armEntry = new ArmEntry<T>()
+            {
+                Properties = o
+            };
+
+            // In Azure ARM requests, the referrer is the current id
+            Uri referrer = request.Headers.Referrer;
+            if (referrer != null)
```
suwatch: This is one of the contract that it came from CSM.   Maybe we should add this check in IsArmRequest - and no need to check here.

Kudu.Services/Arm/ArmActionSelector.cs
suwatch: Especially, where it is being consumed.  I somehow can't find it.

Kudu.Services/Arm/ArmControllerConfiguration.cs
davidebbo: Yes, it's tricky. It modifies the name of the action method that we're binding to. So foo --> FooArm. That's why you don't find "arm" literal anywhere else. Definitely calls for comment!

Kudu.Services/Arm/ArmUtils.cs
amitapl: I would really rather be safe here

Kudu.Services/Arm/ArmUtils.cs
amitapl: Again I really rather be safe

amitapl:
Added some comments

amitapl:
shipit?

davidebbo:
Yep.

PR #1449, opened 2015-01-30
fix #1063 support SCM_REPOSITORY_PATH and SCM_TARGET_PATH absolute path

### LINKED ISSUE #1063: SCM_REPOSITORY_PATH value should be consistent
The current value is `repository` or `wwwroot` which is relative under site folder.  However, other apis take path relative to root.  For instance, `site/wwwroot` or `site/repository`.   We should strike for consistency.  To minimize breaking change, we handle `wwwroot` as `site\wwwroot` (common initial clone scenarios).

suwatch:


davidebbo:
:shipit:

PR #1455, opened 2015-02-09
check if subdirectory is LocalSiteRoot first before SystemDrive since th...

ahmelsayed:
...at breaks 'cd d:\local'


davidebbo:
:shipit:

PR #1462, opened 2015-02-12
add last site config modified time to environment api

suwatch:
this would allow a better polling to see if the new config is effective than polluting with dummy appSettings.


amitapl:
looks good

PR #1464, opened 2015-02-12
Make site extension install/update idempotent; Added recycle header after successfully install site extension

shrimpy:
Make site extension install/update idempotent; Added recycle header after successfully install site extension

### DIFF
--- a/Common/Constants.cs
+++ b/Common/Constants.cs
@@ -63,5 +63,8 @@ namespace Kudu
         {
             get { return _maxAllowedExectionTime; }
         }
+
+        public const string SiteOperationHeaderKey = "X-MS-SITE-OPERATION";
+        public const string SiteOperationRestart = "restart";
     }
 }
--- a/Kudu.Client/Infrastructure/HttpClientExtensions.cs
+++ b/Kudu.Client/Infrastructure/HttpClientExtensions.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Net.Http;
 using System.Threading.Tasks;
+using Kudu.Contracts.SiteExtensions;
 using Newtonsoft.Json;
 
 namespace Kudu.Client.Infrastructure
@@ -87,9 +88,23 @@ namespace Kudu.Client.Infrastructure
 
         public static async Task<TOutput> PutJsonAsync<TInput, TOutput>(this HttpClient client, string url, TInput param)
         {
-            HttpResponseMessage result = await client.PutAsJsonAsync(url, param);
+            HttpResponseMessage response = await client.PutAsJsonAsync(url, param);
+            string content = await response.EnsureSuccessful().Content.ReadAsStringAsync();
 
-            string content = await result.EnsureSuccessful().Content.ReadAsStringAsync();
+            var outputType = typeof(TOutput);
+            if (HttpResponseResultUtils.IsTypeOfHttpResponseRresult(outputType))
+            {
+                Type bodyType = outputType.GenericTypeArguments[0]; // HttpResponseResult<T> takes one generic type
+                var bodyObject = JsonConvert.DeserializeObject(value: content, type: bodyType);
+                var headerDict = new Dictionary<string, IEnumerable<string>>();
+
+                foreach (var item in response.Headers)
+                {
+                    headerDict.Add(item.Key, item.Value);
+                }
+
+                return (TOutput)HttpResponseResultUtils.CreateHttpResponseResultInstance(outputType, headerDict, bodyObject);
+            }
 
             return JsonConvert.DeserializeObject<TOutput>(content);
         }
--- a/Kudu.Client/Infrastructure/HttpResponseResult.cs
+++ b/Kudu.Client/Infrastructure/HttpResponseResult.cs
@@ -1,0 +1,50 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace Kudu.Client.Infrastructure
+{
+    public class HttpResponseResult<T>
+    {
+        public IDictionary<string, IEnumerable<string>> Headers { get; set; }
+        public T Body { get; set; }
+
+        private HttpResponseResult(IDictionary<string, IEnumerable<string>> headers, T body)
+        {
+            Headers = headers;
+            Body = body;
+        }
+    }
+
+    /// <summary>
+    /// <para>CA1000: Do not declare static members on generic types</para>
+    /// <para>https://msdn.microsoft.com/en-us/library/ms182139.aspx</para>
+    /// </summary>
+    internal static class HttpResponseResultUtils
+    {
+        public static bool IsTypeOfHttpResponseRresult(Type t)
+        {
+            var expectedType = typeof(HttpResponseResult<>);
+            return t.IsGenericType
+                && t.GenericTypeArguments.Length == 1
+                && t.GetGenericTypeDefinition() == expectedType;
+        }
+
+        public static object CreateHttpResponseResultInstance(Type httpResponseResultType, IDictionary<string, IEnumerable<string>> headers, object body)
+        {
+            var expectedType = typeof(HttpResponseResult<>);
+            if (!IsTypeOfHttpResponseRresult(httpResponseResultType))
+            {
+                throw new ArgumentException(string.Format("Unexpected HttpResponseResult type: '{0}'", httpResponseResultType.FullName));
+            }
+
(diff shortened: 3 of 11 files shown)


Kudu.Core/SiteExtensions/SiteExtensionManager.cs
```diff
@@ -306,6 +327,40 @@ private SiteExtensionInfo GetPreInstalledExtension(string id)
         }
 
         /// <summary>
+        /// <para> Return true if any of below cases is satisfied:</para>
+        /// <para>      1) Package with same version and from same feed already exist in local repo</para>
+        /// <para>      2) If given feedUrl is null</para>
+        /// <para>              Package with same version from default feed already exist in local repo</para>
+        /// <para>      3) If given version and feedUrl are null</para>
+        /// <para>              Same version after query default feed for latest package</para>
```
shrimpy: i think i am missing one more condition.

if feed is null, i should try the feed from the installed package, only if the feed from local package is also null, then fallback to default feed

Common/Constants.cs
```diff
@@ -63,5 +63,8 @@ public static TimeSpan MaxAllowedExecutionTime
         {
             get { return _maxAllowedExectionTime; }
         }
+
+        public const string SiteOperationHeaderKey = "X-MS-SITE-OPERATION";
+        public const string SiteOperationRecycle = "recycle";
```
suwatch: let call it restart.

shrimpy: sure.. will update the const and value to "restart"

Kudu.Services/SiteExtensions/SiteExtensionController.cs
```diff
@@ -55,18 +57,18 @@ public SiteExtensionController(ISiteExtensionManager manager)
...
-        public async Task<SiteExtensionInfo> InstallExtension(string id, SiteExtensionInfo requestInfo)
+        public async Task<HttpResponseMessage> InstallExtension(string id, SiteExtensionInfo requestInfo)
         {
             if (requestInfo == null)
             {
                 requestInfo = new SiteExtensionInfo();
             }
 
-            SiteExtensionInfo extension;
+            Tuple<bool, SiteExtensionInfo> result;
```
suwatch: Is it possible to get away from tuple (hard to read)?   for instance, can we deduct this from SiteExtensionInfo.
  - call from ARM
  - info indicates success
  - info indicate installed time > request start time

Kudu.Core/SiteExtensions/SiteExtensionManager.cs
```diff
@@ -297,12 +312,60 @@ private SiteExtensionInfo GetPreInstalledExtension(string id)
...
+        /// <para> Return true if any of below cases is satisfied:</para>
+        /// <para>      1) Package with same version and from same feed already exist in local repo</para>
+        /// <para>      2) If given feedUrl is null</para>
+        /// <para>              Try to use feed from local package, if feed from local package also null, fallback to default feed</para>
+        /// <para>              Check if version from query is same as local package</para>
+        /// <para>      3) If given version and feedUrl are null</para>
+        /// <para>              Try to use feed from local package, if feed from local package also null, fallback to default feed</para>
+        /// <para>              Check if version from query is same as local package</para>
+        /// </summary>
+        private async Task<bool> IsSiteExtensionInstalled(string id, string version, string feedUrl)
```
suwatch: This is called exclusively by InstallSiteExtension.  Meaning feedUrl should have a value (never null - either user provided or default), no?

Kudu.Client/Infrastructure/HttpClientExtensions.cs
```diff
@@ -93,5 +93,22 @@ public static T GetJson<T>(this HttpClient client, string url)
 
             return JsonConvert.DeserializeObject<TOutput>(content);
         }
+
+        public static async Task<HttpResponseResult<TOutput>> RichPutJsonAsync<TInput, TOutput>(this HttpClient client, string url, TInput param)
```
suwatch: I would just do it in the existing PutJsonAsync.  If we detect that TOutput type is HttpResponseResult (or certain interface), we always add response headers.   It is transparent to callers.

Kudu.Core/SiteExtensions/SiteExtensionManager.cs
shrimpy: right now actually most of the case feedUrl is null.
when install site extension from SCM site, feedUrl is always null and relying on code to fallback to "siteextensions.net"

Kudu.Client/Infrastructure/HttpClientExtensions.cs
shrimpy: yep sure.. i was actually thinking about it at the very beginning, didn`t go with it was worry about people who call it might now know the feature. 

now look back seems like i was over-worry . 

will update it.

Kudu.Services/SiteExtensions/SiteExtensionController.cs
shrimpy: ah.. good idea. let`s do that.

Kudu.Client/Infrastructure/HttpClientExtensions.cs
```diff
@@ -87,9 +88,26 @@ public static T GetJson<T>(this HttpClient client, string url)
...
+            string content = await response.EnsureSuccessful().Content.ReadAsStringAsync();
 
-            string content = await result.EnsureSuccessful().Content.ReadAsStringAsync();
+            var outputType = typeof(TOutput);
+            if (outputType.Name.Equals("HttpResponseResult`1"))
+            {
+                Type bodyType = outputType.GenericTypeArguments[0]; // HttpResponseResult<T> takes one generic type
+                var bodyObject = JsonConvert.DeserializeObject(value: content, type: bodyType);
+                var tmpBody = JsonConvert.DeserializeObject<SiteExtensionInfo>(content);
+                var tmpBody2 = JsonConvert.DeserializeObject(content, type: typeof(SiteExtensionInfo));
```
suwatch: please remove

Kudu.Client/Infrastructure/HttpClientExtensions.cs
```diff
@@ -87,9 +88,26 @@ public static T GetJson<T>(this HttpClient client, string url)
 
         public static async Task<TOutput> PutJsonAsync<TInput, TOutput>(this HttpClient client, string url, TInput param)
         {
-            HttpResponseMessage result = await client.PutAsJsonAsync(url, param);
+            HttpResponseMessage response = await client.PutAsJsonAsync(url, param);
+            string content = await response.EnsureSuccessful().Content.ReadAsStringAsync();
 
-            string content = await result.EnsureSuccessful().Content.ReadAsStringAsync();
+            var outputType = typeof(TOutput);
+            if (outputType.Name.Equals("HttpResponseResult`1"))
```
suwatch: Use GetGenericTypeDefinition for comparison as opposed to string.

Kudu.Client/Infrastructure/HttpResponseResult.cs
```diff
@@ -0,0 +1,50 @@
...
+            Headers = headers;
+            Body = body;
+        }
+    }
+
+    /// <summary>
+    /// <para>CA1000: Do not declare static members on generic types</para>
+    /// <para>https://msdn.microsoft.com/en-us/library/ms182139.aspx</para>
+    /// </summary>
+    public class HttpResponseResultUtils
```
shrimpy: thinking to change HttpResponseResultUtils to be "internal" instead of "public"

Kudu.Services/SiteExtensions/SiteExtensionController.cs
```diff
@@ -79,12 +81,26 @@ public SiteExtensionController(ISiteExtensionManager manager)
...
 
-            return extension;
+            // TODO: xiaowu, when update to real csm async, should return "Accepted" instead of "OK"
+            var responseMessage = Request.CreateResponse(HttpStatusCode.OK, ArmUtils.AddEnvelopeOnArmRequest<SiteExtensionInfo>(result, Request));
+
+            if (result != null
+                && result.InstalledDateTime.HasValue
+                && result.InstalledDateTime.Value > HttpContext.Current.Timestamp.ToUniversalTime()
+                && ArmUtils.IsArmRequest(Request))
+            {
```
suwatch: This should include SUCCESS status check, correct?

shrimpy: when result is not null indicate success. i will put more comments here.

PR #1471, opened 2015-02-20
Update NuGet V3 client to 3.0.0-pre-20150220053309

shrimpy:
all site extension test case passed
also verified from portal, see below screenshot

image

### DIFF
--- a/Kudu.Contracts/SiteExtensions/SiteExtensionInfo.cs
+++ b/Kudu.Contracts/SiteExtensions/SiteExtensionInfo.cs
@@ -44,7 +44,7 @@ namespace Kudu.Contracts.SiteExtensions
         public SiteExtensionInfo(UIPackageMetadata data)
         {
             Id = data.Identity.Id;
-            Title = data.Identity.Id;
+            Title = data.Title;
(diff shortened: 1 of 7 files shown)


Kudu.Contracts/SiteExtensions/SiteExtensionInfo.cs
```diff
@@ -61,7 +61,7 @@ public SiteExtensionInfo(UIPackageMetadata data)
         public SiteExtensionInfo(UISearchMetadata data)
```
ahmelsayed: any reason we can't call the `SiteExtensionInfo(UIPackageMetadata)` constructor with `data.LatestPackageMetadata` ? they seem to be doing the same thing

any reason this constructor can't just be 

```
```

shrimpy: good call. only some fields need to be override. let me udpate it.

shrimpy: actually can remove it now after update search to use SearchLatestPackageResource.

ahmelsayed:
:shipit:

Issue #1470, opened 2015-02-20
Fix support for GitLab payload


davidebbo:
I just tried to use it from gitlab.com by manually setting up the hook. It seems our logic for gitlabhq is no longer working, possibly because the payload changed since it was written. You can find the code here.
 
Focusing only on the relevant parts, I get a payload that looks like this:

```json 
{  
   "ref":"refs/heads/master",
   "user_id":12345,
   "user_name":"David Ebbo",
   "repository":{  
      "name":"Mvc51ApplicationFx451",
      "url":"git@gitlab.com:david-ebbo/Mvc51ApplicationFx451.git",
      "description":"Test MVC project",
      "homepage":"https://gitlab.com/david-ebbo/Mvc51ApplicationFx451"
   }
}
```
 
And for a public repo (as is the case for https://gitlab.com/david-ebbo/Mvc51ApplicationFx451), we’re expecting an http URL and not ssh. So that blows up in there.

davidebbo:
Gitlab folks wrote:

>The repository['url'] field has returned the ssh url for long time (maybe from beginning of GitLab). You can use the repository['homepage'] field and add `.git` to it to get the git over HTTP url.

randx:
@davidebbo hey. You can follow advice and use `homepage` for HTTP url. In future we are planning to explicitly add `git_http_url` and `git_ssh_url` for hook payload

suwatch:
As you mentioned, it was missing the is_private (or public) field as part of hook payload.   One workaround is to Http GET on  repository['homepage'] url.   For public, it returns 200.  For private, it redirect to login (302).   Of course, it is best they include this info in the playload.

```json
{
  "before": "20fb361885bc66358b7d73e1bd11eda08c643b49",
  "after": "1ba391f4d100ab8dde08e8bf77355a67325182d6",
  "ref": "refs/heads/master",
  "checkout_sha": "1ba391f4d100ab8dde08e8bf77355a67325182d6",
  "user_id": 99630,
  "user_name": "Suwat Bodin",
  "project_id": 168565,
  "repository": {
    "name": "HelloPrivate",
    "url": "git@gitlab.com:suwatch01/HelloPrivate.git",
    "description": "",
    "homepage": "https://gitlab.com/suwatch01/HelloPrivate"
  },
  "commits": [
    {
      "id": "1ba391f4d100ab8dde08e8bf77355a67325182d6",
      "message": "testing gitlab2\n",
[...]

PR #1475, opened 2015-02-24
Fix support for GitLab payload #1470

suwatch:


Kudu.Services/ServiceHookHandlers/GitlabHqHandler.cs
```diff
@@ -7,68 +7,81 @@ namespace Kudu.Services.ServiceHookHandlers
...
             if (repository == null || userid == null || username == null)
             {
                 // doesn't look like GitlabHQ
                 return null;
             }
 
-            string newRef = payload.Value<string>("after");
-            if (IsDeleteCommit(newRef))
+            var url = repository.Value<string>("url");
+            if (url == null || url.IndexOf("gitlab.com", StringComparison.OrdinalIgnoreCase) < 0)
```
davidebbo: I think we want to avoid testing the domain name, as this will block and other gitlab deployment (i.e. like the GitHub enterprise scenario). So it's best to sniff it in some other way.

suwatch: sure.   I wonder why the old code did not test it.

PR #1480, opened 2015-02-26
Support ASP.NET 5 beta3

ahmelsayed:
This goes with this change in KuduScript https://github.com/projectkudu/KuduScript/commit/642274789013a85b66b1eb3bf305562191735137

I have a question about [`IBuildPropertyProvider`](https://github.com/projectkudu/kudu/blob/master/Kudu.Core/Deployment/IBuildPropertyProvider.cs) it seems to have only 1 implementer [`BuildPropertyProvider`](https://github.com/projectkudu/kudu/blob/master/Kudu.Core/Deployment/BuildPropertyProvider.cs) yet no one ever uses it! The only place it's used is in [this loop](https://github.com/projectkudu/kudu/blob/master/Kudu.Core/Deployment/Generator/ExternalCommandBuilder.cs#L66) but it's always going to be empty!
I'm asking because I am not sure what it was intended for but [I am using it to add environment variable for the correct runtime version for asp.net](https://github.com/projectkudu/kudu/blob/16b44a6347fc4d9617217b31a8db6b83121e3fd2/Kudu.Core/Deployment/Generator/AspNet5Builder.cs) Do you see any concerns?


Kudu.Core/Deployment/Generator/ExternalCommandFactory.cs
```diff
@@ -168,9 +167,9 @@ private static string KreBitness
                 var bitness = System.Environment.GetEnvironmentVariable(WellKnownEnvironmentVariables.SiteBitness);
                 if (bitness == null)
                 {
-                    return System.Environment.Is64BitProcess ? "amd64" : "x86";
+                    return System.Environment.Is64BitProcess ? "X64" : "x86";
```
davidfowl: The bitness of the kudu process is the bitness of the deployment target?

davidfowl: Also, lowercase x64

ahmelsayed: Only if `WellKnownEnvironmentVariables.SiteBitness` is null. kudu's `w3wp.exe` will be the same bitness as the main site's process, but that's not true for `kudu.exe`. If deployment is happening in `kudu.exe`, then that environment variable won't be null, else we are running deployment in `w3wp.exe`

But this actually reminds me I should fix that environment variable as well to be X64 rather than Amd64

davidebbo:
:shipit:

PR #1489, opened 2015-03-07
Micro-Service support

shrimpy:
if given type is "WebRoot", it would indicate we will need to copy site extension content to "wwwroot"

install:
   package content will be extra to wwwroot
   nupkg will still site in siteextenion/{packageid}
   not to check xdt file
   not to deploy webjob
   ignore install.cmd

Uninstall:
   remove content from wwwroot
   remove folder siteextenion/{packageid}
   not to undeploy webjob
   ignore uninstall.cmd

all site extension test case are passed

### DIFF
--- a/Kudu.Client/SiteExtensions/RemoteSiteExtensionManager.cs
+++ b/Kudu.Client/SiteExtensions/RemoteSiteExtensionManager.cs
@@ -5,6 +5,7 @@ using System.Text;
 using System.Threading.Tasks;
 using System.Web;
 using Kudu.Client.Infrastructure;
+using Kudu.Contracts.SiteExtensions;
 using Kudu.Services.Arm;
 using Newtonsoft.Json.Linq;
 
@@ -114,11 +115,15 @@ namespace Kudu.Client.SiteExtensions
             return (await Client.GetAsync(url.ToString())).EnsureSuccessful();
         }
 
-        public async Task<HttpResponseMessage> InstallExtension(string id, string version = null, string feedUrl = null)
+        public async Task<HttpResponseMessage> InstallExtension(string id, string version = null, string feedUrl = null, SiteExtensionInfo.SiteExtensionType? type = null)
         {
             var json = new JObject();
             json["version"] = version;
             json["feed_url"] = feedUrl;
+            if (type.HasValue)
+            {
+                json["type"] = Enum.GetName(typeof(SiteExtensionInfo.SiteExtensionType), type.Value);
+            }
 
             return (await Client.PutAsJsonAsync("siteextensions/" + id, json)).EnsureSuccessful();
         }
@@ -128,7 +133,6 @@ namespace Kudu.Client.SiteExtensions
             var url = new StringBuilder(ServiceUrl);
             url.Append("siteextensions/");
             url.Append(id);
-
             return (await Client.DeleteAsync(url.ToString())).EnsureSuccessful();
         }
     }
--- a/Kudu.Contracts/SiteExtensions/SiteExtensionInfo.cs
+++ b/Kudu.Contracts/SiteExtensions/SiteExtensionInfo.cs
@@ -1,8 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Net;
 using Kudu.Contracts.Infrastructure;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
 using NuGet.Client.VisualStudio;
 
 namespace Kudu.Contracts.SiteExtensions
@@ -14,7 +14,8 @@ namespace Kudu.Contracts.SiteExtensions
         {
             Gallery,
             PreInstalledMonaco,
-            PreInstalledEnabled
+            PreInstalledEnabled,
+            WebRoot
         }
 
         public SiteExtensionInfo()
@@ -73,7 +74,8 @@ namespace Kudu.Contracts.SiteExtensions
             set;
         }
 
-        [JsonIgnore]
+        [JsonProperty(PropertyName = "type")]
+        [JsonConverter(typeof(StringEnumConverter))]
         public SiteExtensionType Type
         {
             get;
--- a/Kudu.Core/SiteExtensions/SiteExtensionManager.cs
+++ b/Kudu.Core/SiteExtensions/SiteExtensionManager.cs
@@ -41,6 +41,7 @@ namespace Kudu.Core.SiteExtensions
         private const string _settingsFileName = "SiteExtensionSettings.json";
         private const string _feedUrlSetting = "feed_url";
         private const string _installUtcTimestampSetting = "install_timestamp_utc";
+        private const string _packageType = "package_type";
         private const string _versionSetting = "version";
 
         private readonly string _rootPath;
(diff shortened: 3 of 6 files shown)


Kudu.Client/SiteExtensions/RemoteSiteExtensionManager.cs
```diff
@@ -114,21 +115,31 @@ public RemoteSiteExtensionManager(string serviceUrl, ICredentials credentials =
...
+            if (type.HasValue)
+            {
+                json["type"] = (int)type.Value;
+            }
 
             return (await Client.PutAsJsonAsync("siteextensions/" + id, json)).EnsureSuccessful();
         }
 
-        public async Task<HttpResponseMessage> UninstallExtension(string id)
+        public async Task<HttpResponseMessage> UninstallExtension(string id, SiteExtensionInfo.SiteExtensionType? type = null)
```
davidebbo: Infer from settings file instead of passing

Kudu.Contracts/SiteExtensions/SiteExtensionInfo.cs
```diff
@@ -73,7 +74,7 @@ public string Title
             set;
         }
 
-        [JsonIgnore]
+        [JsonProperty(PropertyName = "type")]
```
davidebbo: type is coming back as 0 when I tried it

Kudu.Core/SiteExtensions/SiteExtensionManager.cs
```diff
@@ -449,33 +451,44 @@ private SiteExtensionInfo GetPreInstalledExtension(string id)
                 string packageLocalFilePath = GetNuGetPackageFile(package.Identity.Id, package.Identity.Version.ToString());
                 using (tracer.Step("Download site extension: {0}", package.Identity))
                 {
-                    await remoteRepo.DownloadPackageToFolder(package.Identity, installationDirectory, pathToLocalCopyOfNudpk: packageLocalFilePath);
-                }
+                    string extraPath = installationDirectory;
+                    if (type.HasValue && type.Value == SiteExtensionInfo.SiteExtensionType.WebRoot)
+                    {
+                        extraPath = _environment.WebRootPath;
```
davidebbo: `extractPath`

Kudu.Core/SiteExtensions/SiteExtensionManager.cs
```diff
@@ -449,33 +451,44 @@ private SiteExtensionInfo GetPreInstalledExtension(string id)
...
+                    }
 
-                // If there is no xdt file, generate default.
-                using (tracer.Step("Check if applicationhost.xdt file existed."))
-                {
-                    GenerateApplicationHostXdt(installationDirectory, '/' + package.Identity.Id, isPreInstalled: false, tracer: tracer);
+                    await remoteRepo.DownloadPackageToFolder(package.Identity, extraPath, pathToLocalCopyOfNudpk: packageLocalFilePath);
                 }
 
-                using (tracer.Step("Trigger site extension job"))
```
davidebbo: Avoid microservice naming

Kudu.TestHarness/AppDomainHelper.cs
```diff
@@ -24,7 +24,7 @@ static AppDomainHelper()
                 appDomain.DomainUnload += delegate
                 {
                     // wait for all task completion
-                    _threadCompletion.WaitOne(60000);
+                    _threadCompletion.WaitOne(60000*10);
```
shrimpy: revert. this is debug only.

Kudu.Core/SiteExtensions/SiteExtensionManager.cs
```diff
@@ -449,33 +451,46 @@ private SiteExtensionInfo GetPreInstalledExtension(string id)
...
+                        FileSystemHelpers.EnsureDirectory(extractPath);
+                    }
 
-                // If there is no xdt file, generate default.
-                using (tracer.Step("Check if applicationhost.xdt file existed."))
-                {
-                    GenerateApplicationHostXdt(installationDirectory, '/' + package.Identity.Id, isPreInstalled: false, tracer: tracer);
+                    await remoteRepo.DownloadPackageToFolder(package.Identity, extractPath, pathToLocalCopyOfNudpk: packageLocalFilePath);
+                    tracer.Trace("Generate xdt file to map application to run from wwwroot when request for /{0}", package.Identity.Id);
+                    GenerateRemapToWebRootApplicationHostXdt(installationDirectory, string.Format(CultureInfo.InvariantCulture, "/{0}", package.Identity.Id));
```
davidebbo: It looks like this always generate it, even in the non-WebRoot case. Doesn't that break things for the general case of a site extension that has an xdt?

shrimpy: ah, i am too careless, was support to move it to the if statement above

test case pass thought i did the right thing. it is fixed now.

Kudu.Contracts/SiteExtensions/SiteExtensionInfo.cs
```diff
@@ -10,11 +9,12 @@ namespace Kudu.Contracts.SiteExtensions
     // This is equivalent to NuGet.IPackage
     public class SiteExtensionInfo : INamedObject
     {
-        public enum SiteExtensionType
+        public static class SiteExtensionType
```
davidebbo: Can't it still be an Enum and then use `StringEnumConverter`? See [here](http://stackoverflow.com/questions/2441290/json-serialization-of-enum-as-string). The advantage is that it avoids having to deal with strings throughout the code.

shrimpy: didn`t know this exist. the main reason is to update "Type" to return string value to caller.

let me give it a try to see what value we will return if we keep using enum with StringEnumConverter

Kudu.Core/SiteExtensions/SiteExtensionManager.cs
```diff
@@ -449,33 +451,47 @@ private SiteExtensionInfo GetPreInstalledExtension(string id)
                 string packageLocalFilePath = GetNuGetPackageFile(package.Identity.Id, package.Identity.Version.ToString());
                 using (tracer.Step("Download site extension: {0}", package.Identity))
                 {
-                    await remoteRepo.DownloadPackageToFolder(package.Identity, installationDirectory, pathToLocalCopyOfNudpk: packageLocalFilePath);
-                }
+                    string extractPath = installationDirectory;
+                    if (SiteExtensionInfo.SiteExtensionType.WebRoot == type)
```
davidebbo: Not a big deal (and no need to change), but we don't usually use reversed comparisons in C#. These were useful in C/C++ where they helped catch cases where `=` was used by mistake instead of `==`. But C# prevents this.

Kudu.Core/SiteExtensions/SiteExtensionManager.cs
```diff
@@ -597,27 +622,43 @@ private static void GenerateApplicationHostXdt(string installationDirectory, str
                 throw new DirectoryNotFoundException(installationDirectory);
             }
 
-            var externalCommandFactory = new ExternalCommandFactory(_environment, _settings, installationDirectory);
+            JsonSettings siteExtensionSettings = GetSettingManager(id);
+            string packageTypeStr = siteExtensionSettings.GetValue(_packageType);
+            SiteExtensionInfo.SiteExtensionType packageType;
+            if (Enum.TryParse<SiteExtensionInfo.SiteExtensionType>(packageTypeStr, out packageType)
```
davidebbo: Maybe pass ignoreCase to TryParse, to be on safe side

davidebbo: SInce you have the same logic twice, would be good to have a wrapper. e.g. given an ID, is it WebRoot.

shrimpy: sure.. let me update it.

shrimpy:
merge in b0355d8882c48c6d27b5e4dd3670eeb60efd2fbe

Issue #1492, opened 2015-03-10
Unlisted packages are installed


sebastienros:
When no specific version is given for a package, the latest version is taken into account. The issue is that it also looks into the unlisted packages.

shrimpy:
thanks for reporting this. taking a look now.

shrimpy:
Fixed in a0ec7add404d2eba96cd01298fea0fc59fc01b5e

davidebbo:
I think the desired behavior here should be the same as with NuGet.exe:
- if no version is passed in, you the latest that's not unlisted
- if a version is passed in, honor it regardless of whether it's listed or unlisted

Is that how your change works? Looking at the code, I'm not so sure.

PR #1493, opened 2015-03-10
FIXED: #1492, should not see or install unlisted packages

shrimpy:


ahmelsayed:
:shipit:

davidebbo:
Copying my comment from the issue:

I think the desired behavior here should be the same as with NuGet.exe:
- if no version is passed in, you the latest that's not unlisted
- if a version is passed in, honor it regardless of whether it's listed or unlisted

Is that how your change works? Looking at the code, I'm not so sure.

shrimpy:
current code change will not look from unlisted package. 
tried with Nuget.exe, it won`t list any unlisted package, but do allow user to install package that is "unlisted"

updating logic to do similar behavior. will have code update soon.

shrimpy:
@davidebbo sent out another PR to address your comment https://github.com/projectkudu/kudu/pull/1494, please review

PR #1494, opened 2015-03-10
Update code to allow install unlisted package, but not able to search for it

shrimpy:
address comment from PR https://github.com/projectkudu/kudu/pull/1493:

•	if no version is passed in, you the latest that's not unlisted
•	if a version is passed in, honor it regardless of whether it's listed or unlisted


davidebbo:
:shipit:

PR #1497, opened 2015-03-12
C#6 supports using await in the catch blocks.

kernel32uiuc:


amitapl:
We are currently not using C# 6 to build kudu.

Thanks!

PR #1498, opened 2015-03-12
Remove SiteExtensionInfo.Name from JSON payload; Update ArmEntity.Name to not start with '/'

shrimpy:
Remove SiteExtensionInfo.Name from JSON payload; 
Update ArmEntity.Name to not start with '/'


suwatch:
:shipit:

PR #1499, opened 2015-03-14
Webjob as microservice

shobak101:
in this change the swagger file generated does not have any parameters for web jobs and conforms to swagger specs, 

Example:

```json
{
  "swagger": "2.0",
  "info": {
    "version": "v1",
    "title": "WebJobAsMicroService"
  },
  "host": "placeHolder",
  "schemes": [
    "http",
    "https"
  ],
  "paths": {
    "/HelloWorldWebJob/run": {
      "post": {
        "deprecated": false,
        "operationId": "HelloWorldWebJob_Post",
        "consumes": [],
        "produces": [],
        "responses": {
          "200": {
            "description": "Success"
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/Kudu.Services.Web/App_Start/NinjectServices.cs
+++ b/Kudu.Services.Web/App_Start/NinjectServices.cs
@@ -454,6 +454,7 @@ namespace Kudu.Services.Web.App_Start
             // Jobs
             routes.MapHttpWebJobsRoute("list-all-jobs", "", "", new { controller = "Jobs", action = "ListAllJobs" }, new { verb = new HttpMethodConstraint("GET") });
             routes.MapHttpWebJobsRoute("list-triggered-jobs", "triggered", "", new { controller = "Jobs", action = "ListTriggeredJobs" }, new { verb = new HttpMethodConstraint("GET") });
+            routes.MapHttpWebJobsRoute("list-triggered-jobs-swagger", "triggered", "/swagger", new { controller = "Jobs", action = "ListTriggeredJobsInSwaggerFormat" }, new { verb = new HttpMethodConstraint("GET") });
             routes.MapHttpWebJobsRoute("get-triggered-job", "triggered", "/{jobName}", new { controller = "Jobs", action = "GetTriggeredJob" }, new { verb = new HttpMethodConstraint("GET") });
             routes.MapHttpWebJobsRoute("invoke-triggered-job", "triggered", "/{jobName}/run", new { controller = "Jobs", action = "InvokeTriggeredJob" }, new { verb = new HttpMethodConstraint("POST") });
             routes.MapHttpWebJobsRoute("get-triggered-job-history", "triggered", "/{jobName}/history", new { controller = "Jobs", action = "GetTriggeredJobHistory" }, new { verb = new HttpMethodConstraint("GET") });
--- a/Kudu.Services/Jobs/SwaggerApiDef.cs
+++ b/Kudu.Services/Jobs/SwaggerApiDef.cs
@@ -1,0 +1,140 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Kudu.Contracts.Jobs;
+using Newtonsoft.Json;
+
+namespace Kudu.Services.Jobs
+{
+    public class SwaggerApiDef
+    {
+        [JsonProperty(PropertyName = "swagger")]
+        public string Swagger { get; set; }
+
+        [JsonProperty(PropertyName = "info")]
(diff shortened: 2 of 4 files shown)


Kudu.Services/Jobs/SwaggerApiDef.cs
```diff
@@ -0,0 +1,119 @@
...
+        [JsonProperty(PropertyName = "schemes")]
+        public List<String> Schemes { get; private set; }
+        [JsonProperty(PropertyName = "paths")]
+        public Dictionary<String, PathItem> Paths { get; set; }
+        public SwaggerApiDef(IEnumerable<JobBase> triggeredJobs)
+        {
+            Swagger = "2.0";
+            Info = new SwaggerApiDefInfo();
+            Host = "placeHolder";
+            Schemes = new List<String> { "http", "https" };
```
davidebbo: scm api is only https

Kudu.Services/Jobs/SwaggerApiDef.cs
```diff
@@ -0,0 +1,119 @@
...
+    {
+        [JsonProperty(PropertyName = "version")]
+        public String Version { get; set; }
+        [JsonProperty(PropertyName = "title")]
+        public String Title { get; set; }
+
+        public SwaggerApiDefInfo()
+        {
+            Version = "v1";
+            Title = "WebJobAsMicroService";
```
davidebbo: Let's just call it `WebJob`

Kudu.Services/Jobs/SwaggerApiDef.cs
```diff
@@ -0,0 +1,119 @@
...
+        public SwaggerApiDef(IEnumerable<JobBase> triggeredJobs)
+        {
+            Swagger = "2.0";
+            Info = new SwaggerApiDefInfo();
+            Host = "placeHolder";
+            Schemes = new List<String> { "http", "https" };
+            Paths = new Dictionary<string, PathItem>();
+            foreach (var triggeredJob in triggeredJobs)
+            {
+                Paths.Add("/" + triggeredJob.Name + "/run", PathItem.GetDefaultPathItem(triggeredJob.Name + "_Post"));
```
davidebbo: Let's just use the job name, with no suffix.

davidebbo:
3 minor changes, and good to go!

Kudu.Services/Jobs/SwaggerApiDef.cs
```diff
@@ -0,0 +1,135 @@
...
+using System.Threading.Tasks;
+using Kudu.Contracts.Jobs;
+using Newtonsoft.Json;
+
+namespace Kudu.Services.Jobs
+{
+    public class SwaggerApiDef
+    {
+        [JsonProperty(PropertyName = "swagger")]
+        public String Swagger { get; set; }
```
ahmelsayed: I think we normally use lowercase `string` for the datatype and the upper case one for the static `String` functions

Kudu.Services.Web/App_Start/NinjectServices.cs
```diff
@@ -454,6 +454,7 @@ public static void RegisterRoutes(IKernel kernel, RouteCollection routes)
             // Jobs
             routes.MapHttpWebJobsRoute("list-all-jobs", "", "", new { controller = "Jobs", action = "ListAllJobs" }, new { verb = new HttpMethodConstraint("GET") });
             routes.MapHttpWebJobsRoute("list-triggered-jobs", "triggered", "", new { controller = "Jobs", action = "ListTriggeredJobs" }, new { verb = new HttpMethodConstraint("GET") });
+            routes.MapHttpWebJobsRoute("list-triggered-jobs-swagger", "triggered", "/swagger", new { controller = "Jobs", action = "ListTriggeredJobsInSwaggerFormat" }, new { verb = new HttpMethodConstraint("GET") });
```
suwatch: To me, it is simply a different format or representation of the same resources.   Instead of an explicit path, could it be query string (i.e. ?format=swagger) or adhoc content-type?  
  - One can't name their job `swagger`
  - How to think the odd/even CSM Uri path

What do other people use to serve up their swaggers?

shobak101: this is going to be specified in the nuget package for webjobs as a service, where you can not add a query string. 

we discussed the problem of having "swagger" as not being available as a webjob name and i think it was ok 

the usual path for the swagger endpoint is api/swagger/v1 for example but it did not make sense for triggered web jobs to be api/triggeredwebjobs/api/swagger/v1

davidebbo: Are you sure it can't have a query string? That would indeed be cleaner. Need to check with @mikestall

PR #1502, opened 2015-03-17
Fixes #1501, Avoid restart when install to WebRoot with no XDT

### LINKED ISSUE #1501: SiteExtension: Avoid restart when install to WebRoot only
need to handle case that two different installation request, one is install to webroot, one is normal site extension installation. in this case we will still need to return restart header to trigger restart.

shrimpy:
1) if site extension install to site extension folder, we will trigger restart
2) if site extension install to wwwroot, and there is xdt file come with package, we will trigger restart
3) if site extension install to wwwroot and no xdt file come with pakcage, we will NOT generate XDT file and will NOT trigger restart

### DIFF
--- a/Kudu.Core/SiteExtensions/SiteExtensionManager.cs
+++ b/Kudu.Core/SiteExtensions/SiteExtensionManager.cs
@@ -37,7 +37,6 @@ namespace Kudu.Core.SiteExtensions
         private readonly IDeploymentSettingsManager _settings;
         private readonly ITraceFactory _traceFactory;
 
-        private const string _applicationHostFile = "applicationHost.xdt";
         private const string _settingsFileName = "SiteExtensionSettings.json";
         private const string _feedUrlSetting = "feed_url";
         private const string _installUtcTimestampSetting = "install_timestamp_utc";
@@ -464,20 +463,18 @@ namespace Kudu.Core.SiteExtensions
                     {
                         // if install to WebRoot, check if there is any xdt file come with package
                         // if there is one, move it to site extension folder
-                        // if not, generate one to map request to wwwroot
-                        string xdtFile = Path.Combine(extractPath, _applicationHostFile);
+                        string xdtFile = Path.Combine(extractPath, Constants.ApplicationHostXdtFileName);
                         if (File.Exists(xdtFile))
                         {
                             tracer.Trace("Use xdt file from package.");
-                            string newXdtFile = Path.Combine(installationDirectory, _applicationHostFile);
+                            string newXdtFile = Path.Combine(installationDirectory, Constants.ApplicationHostXdtFileName);
 
                             tracer.Trace("Moving {0} to {1}", xdtFile, newXdtFile);
                             FileSystemHelpers.MoveFile(xdtFile, newXdtFile);
                         }
                         else
                         {
-                            tracer.Trace("Generate xdt file to map application to run from wwwroot when request for /{0}", package.Identity.Id);
-                            GenerateRemapToWebRootApplicationHostXdt(installationDirectory, string.Format(CultureInfo.InvariantCulture, "/{0}", package.Identity.Id));
+                            tracer.Trace("No xdt file come with package.");
                         }
                     }
                 }
@@ -604,7 +601,7 @@ namespace Kudu.Core.SiteExtensions
         {
             // If there is no xdt file, generate default.
             FileSystemHelpers.CreateDirectory(installationDirectory);
-            string xdtPath = Path.Combine(installationDirectory, _applicationHostFile);
+            string xdtPath = Path.Combine(installationDirectory, Constants.ApplicationHostXdtFileName);
             if (!FileSystemHelpers.FileExists(xdtPath))
             {
                 if (tracer != null)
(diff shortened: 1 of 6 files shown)


Kudu.Core/SiteExtensions/SiteExtensionManager.cs
```diff
@@ -37,7 +37,7 @@ public class SiteExtensionManager : ISiteExtensionManager
         private readonly IDeploymentSettingsManager _settings;
         private readonly ITraceFactory _traceFactory;
 
-        private const string _applicationHostFile = "applicationHost.xdt";
+        private const string _applicationHostFile = Constants.ApplicationHostXdtFileName;
```
davidebbo: Why set a constant to another constant, instead of using `Constants.ApplicationHostXdtFileName` directly?

shrimpy: sure, let me update it.

Kudu.Core/SiteExtensions/SiteExtensionManager.cs
```diff
@@ -476,8 +475,7 @@ private SiteExtensionInfo GetPreInstalledExtension(string id)
                         }
                         else
                         {
-                            tracer.Trace("Generate xdt file to map application to run from wwwroot when request for /{0}", package.Identity.Id);
-                            GenerateRemapToWebRootApplicationHostXdt(installationDirectory, string.Format(CultureInfo.InvariantCulture, "/{0}", package.Identity.Id));
+                            tracer.Trace("No xdt file come with package, shouldn`t restart site when install finish.");
```
shrimpy: remove "shouldn`t restart site when install finish." from trace, avoid causing miss-understanding.
itself won`t trigger restart, but if there is multiple site extension installation happen at the same time, others might trigger site restart.

davidebbo:
We found that the restart behavior is actually smarter than we thought. i.e. if there are no differences between the current and the new applicationhost.config (after transforming xdt's), then IIS is smart enough to nor restart the site at all. That's with the softRestart mechanism that we use.

So it may be that we don't actually need all this extra logic in Kudu to determine whether a restart needs to happen. We all send back the response that triggers the softRestart, and we let the IIS optimization do its thing.

@suwatch @shrimpy thoughts?

suwatch:
Operation wise, it will shave 6 seconds off Geo for final poll GET SiteExtension api call.  Runtime wise, as you pointed out, no affect.

Kudu.Core/SiteExtensions/SiteExtensionStatus.cs
```diff
@@ -104,6 +117,72 @@ public bool IsTerminalStatus()
...
+                            if (IsSiteExtensionRequireRestart(statusSettings, siteExtensionRoot))
+                            {
+                                return true;
+                            }
+                        }
+                    }
+                    catch(Exception ex)
+                    {
+                        tracer.TraceError(ex);
+                        tracer.TraceError("Failed to query {0} under {1}, continus to check others ...", _statusSettingsFileName, dir);
```
suwatch: This could combine into one tracer.TraceError.   Please also make note to support writing to MDS (#1507) as it is unexpected exception.

davidebbo:
I see. In that case, it may still be worthwhile. Though we also have the potential optimization to do it synchronously if it's fast, which would be yet better.

Kudu.Core/SiteExtensions/SiteExtensionStatus.cs
shrimpy: sure will do.

didn`t aware of the MDS tracer, will update code later today. thanks for pointing this out.

Kudu.Core/SiteExtensions/SiteExtensionStatus.cs
```diff
@@ -104,6 +117,74 @@ public bool IsTerminalStatus()
...
+                            {
+                                return true;
+                            }
+                        }
+                    }
+                    catch(Exception ex)
+                    {
+                        analytics.UnexpectedException(ex, trace: false);
+                        tracer.TraceError(ex);
+                        tracer.TraceError("Failed to query {0} under {1}, continus to check others ...", _statusSettingsFileName, dir);
```
shrimpy: not combining the two "tracer.TraceError".

when perform "tracer.TraceError(ex)", "text" will be "ex.Message"
when perform "tracer.TraceError(message, param)", "text" will be given message
there is no interface for "tracer.TraceError(ex, message)"

if it is necessary, we can create "tracer.TraceError(ex, message, params)"
and assign combine given message and ex.Message into "text" attribute

```C#
        public static void TraceError(this ITracer tracer, Exception ex)
        {
            var attribs = new Dictionary<string, string>
            {
                { "type", "error" },
                { "text", ex.Message },
                { "stackTrace", ex.StackTrace ?? String.Empty }
            };

            if (ex.InnerException != null)
            {
                attribs["innerText"] = ex.InnerException.Message;
                attribs["innerStackTrace"] = ex.InnerException.StackTrace ?? String.Empty;
            }

            tracer.Trace("Error occurred", attribs);
        }

        public static void TraceError(this ITracer tracer, string message)
        {
            tracer.Trace("Error occurred", new Dictionary<string, string>
            {
                { "type", "error" },
                { "text", message }
            });
        }
```

suwatch: I think the ex.StackTrace is a weird way to do exception trace comparing to a simpler ex.ToString().   I suggest you do whatever necessary as long as information is sufficient to investigate issue.

PR #1511, opened 2015-03-24
Perf optimization for site extension installation

shrimpy:
improve perf for the case that installation finished quick and doesn`t need to trigger a restart

this will also benefit for scenario that install an site extension where it is already installed, in this case installation will run synchronously

### DIFF
--- a/Kudu.Core/SiteExtensions/SiteExtensionStatus.cs
+++ b/Kudu.Core/SiteExtensions/SiteExtensionStatus.cs
@@ -117,6 +117,12 @@ namespace Kudu.Core.SiteExtensions
                 || string.Equals(Constants.SiteExtensionProvisioningStateCanceled, ProvisioningState, StringComparison.OrdinalIgnoreCase);
         }
 
+        /// <param name="siteExtensionRoot">should be $ROOT\SiteExtensions</param>
+        public bool IsRestartRequired(string siteExtensionRoot)
+        {
+            return this.IsSiteExtensionRequireRestart(siteExtensionRoot);
+        }
+
         /// <summary>
         /// <para>Loop thru all site extension settings files under 'siteExtensionStatusRoot', check if there is any successful installation</para>
         /// <para>if not install to webroot, will trigger restart; if install to webroot and with applicationHost.xdt file, trigger restart.</para>
@@ -137,7 +143,7 @@ namespace Kudu.Core.SiteExtensions
                         foreach (var file in settingFile)
                         {
                             var statusSettings = new SiteExtensionStatus(siteExtensionStatusRoot, dirInfo.Name, tracer);
-                            if (IsSiteExtensionRequireRestart(statusSettings, siteExtensionRoot))
+                            if (statusSettings.IsSiteExtensionRequireRestart(siteExtensionRoot))
                             {
                                 return true;
                             }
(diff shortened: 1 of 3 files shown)


Kudu.Core/SiteExtensions/SiteExtensionManager.cs
```diff
@@ -308,7 +314,7 @@ private SiteExtensionInfo GetPreInstalledExtension(string id)
             }
         }
 
-        private async Task<SiteExtensionInfo> TryInstallExtension(string id, string version, string feedUrl, SiteExtensionInfo.SiteExtensionType type, ITracer tracer)
+        private async Task<SiteExtensionInfo> TryInstallExtension(string id, string version, string feedUrl, SiteExtensionInfo.SiteExtensionType type, ITracer tracer, AutoResetEvent installationSignal = null)
```
davidebbo: nitpick: `=null` default value not needed here

Kudu.Core/SiteExtensions/SiteExtensionStatus.cs
```diff
@@ -117,6 +119,11 @@ public bool IsTerminalStatus()
                 || string.Equals(Constants.SiteExtensionProvisioningStateCanceled, ProvisioningState, StringComparison.OrdinalIgnoreCase);
         }
 
+        public bool IsRequireRestart()
```
davidebbo: `IsRestartRequired` is a bit clearer :)

Kudu.Core/SiteExtensions/SiteExtensionStatus.cs
```diff
@@ -117,6 +119,11 @@ public bool IsTerminalStatus()
...
+        {
+            return IsSiteExtensionRequireRestart(this, _rootPath);
```
davidebbo: Why is `IsSiteExtensionRequireRestart` static? If it takes the instance in, why not make it instance?

Kudu.Core/SiteExtensions/SiteExtensionManager.cs
shrimpy: this will be removed.

Kudu.Core/SiteExtensions/SiteExtensionStatus.cs
shrimpy: sure. will update it.

Kudu.Core/SiteExtensions/SiteExtensionStatus.cs
shrimpy: "IsSiteExtensionRequireRestart" is for shared the logic with "IsAnyInstallationRequireRestart"

when perform GET, we use "IsAnyInstallationRequireRestart" to check if there is any newly installed extension required to restart, if we find out, we will return header to GEO to trigger restart. and "IsAnyInstallationRequireRestart" is static, that is why "IsSiteExtensionRequireRestart" is static

Kudu.Contracts/SiteExtensions/ISiteExtensionManager.cs
```diff
@@ -1,4 +1,6 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
+using System.Threading;
```
shrimpy: reverted

PR #1522, opened 2015-04-01
Update to xUnit 2.0.0

suwatch:


suwatch:
This is only update from Xunit 1.9 to Xunit 2.0.  I want to do just this first and make sure we are in par with the old Xunit 1.9 run.   Next step (different check in) is to turn on parallel runs.

<b>Note</b>:  Xunit VS runner is nuget based only (vsix deprecated).  In order to see in test explorer, you <b>must</b> first remove the Xunit VS runner from VS extensions (VS -> Tools -> Extension and Updates...).

amitapl:
Looks good

Kudu.Core.Test/Infrastructure/XmlUtilityFacts.cs
```diff
@@ -20,15 +19,17 @@ public void XmlUtilityReturnsValidCharacterStringsUnchanged(string input)
             Assert.Equal(input, output);
         }
 
-        [Theory]
-        [PropertyData("InvalidCharacterSet")]
-        public void XmlUtilityRemovesInvalidCharacters(string input)
+        [Fact]
```
pranavkm: Why change this?

Kudu.TestHarness/Xunit/KuduXunitTestCase.cs
```diff
@@ -0,0 +1,168 @@
...
+                if (TestMethodArguments != null)
+                {
+                    // Work around test arg issue
+                    Write(stream, DisplayName);
+                    // Write(stream, SerializationHelper.Serialize(TestMethodArguments));
+                }
+
+                stream.Position = 0;
+
+                var hash = new System.Security.Cryptography.SHA1Managed().ComputeHash(stream);
```
pranavkm: Dispose this?

pranavkm:
:shipit:

Kudu.Core.Test/Infrastructure/XmlUtilityFacts.cs
suwatch: this is a regression from xunit 1.9 when saving the result.  It tries to put that input value to XML result and choke due to invalid Xml character.

https://github.com/xunit/xunit/issues/380

Kudu.TestHarness/Xunit/KuduXunitTestCase.cs
suwatch: definitely.

PR #1523, opened 2015-04-01
improve mds error log for site extension install/uninstall

shrimpy:


Kudu.Core/Tracing/IAnalytics.cs
```diff
@@ -11,7 +8,9 @@ public interface IAnalytics
 
         void JobStarted(string jobName, string scriptExtension, string jobType, string siteMode, string error);
 
-        void UnexpectedException(Exception ex, bool trace = true);
+        void UnexpectedException(Exception exception, bool trace = true);
+
+        void UnexpectedException(Exception ex, string method, string path, string result, string message, bool trace = true);
```
suwatch: ex or exception, do pick one.

shrimpy: will rename to "ex". thanks for noticing that.

suwatch:
Do share out MDS.  Other than that, :shipit:

shrimpy:
yes. just sent out. thanks for the reminding.

PR #1526, opened 2015-04-04
Added handling for invalid remote feed url; Added more log; Improve perf...

shrimpy:
Added handling for invalid remote feed url; 
Added more logging; 
Improve performance for list extensions.

### DIFF
--- a/Kudu.Contracts/Kudu.Contracts.csproj
+++ b/Kudu.Contracts/Kudu.Contracts.csproj
@@ -142,6 +142,7 @@
     <Compile Include="Settings\DeploymentSettingsExtension.cs" />
     <Compile Include="Settings\IDeploymentSettingsManager.cs" />
     <Compile Include="Settings\SettingsKeys.cs" />
+    <Compile Include="SiteExtensions\InvalidEndpointException.cs" />
     <Compile Include="SiteExtensions\ISiteExtensionManager.cs" />
     <Compile Include="SiteExtensions\SiteExtensionInfo.cs" />
     <Compile Include="SourceControl\IFileFinder.cs" />
--- a/Kudu.Core/SiteExtensions/FeedExtensions.cs
+++ b/Kudu.Core/SiteExtensions/FeedExtensions.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Ionic.Zip;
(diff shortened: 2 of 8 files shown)


Kudu.Contracts/Kudu.Contracts.csproj
```diff
@@ -142,6 +142,7 @@
     <Compile Include="Settings\DeploymentSettingsExtension.cs" />
     <Compile Include="Settings\IDeploymentSettingsManager.cs" />
     <Compile Include="Settings\SettingsKeys.cs" />
+    <Compile Include="SiteExtensions\InvalidEndpointException.cs" />
```
shrimpy: let me know if it is overkill to create a specific type of exception. was reference from "Deployment" code.

suwatch: Unless you plan to catch this explicitly and perform specific task, then I don't think this is necessary.   Simple trace in the code should do.

Kudu.Core/SiteExtensions/FeedExtensions.cs
```diff
@@ -161,5 +162,20 @@ public static class FeedExtensions
                 }
             }
         }
+
+        private static async Task<T> GetResourceAndValidateAsync<T>(this SourceRepository srcRepo) where T : class, INuGetResource
+        {
```
suwatch: another way to do this is TryGetResourceAnd...().   And let the caller make decision.

Kudu.Contracts/Kudu.Contracts.csproj
shrimpy: in update scenario, i do catch this exception and send back specify error message to caller, to let request initiator realize the feed endpoint is invalid.

if we only log it, we might get a live site ticket, and will only find out from log.

Kudu.Core/SiteExtensions/FeedExtensions.cs
shrimpy: if we do TryGetResource, when caller see a null provider return. might either return empty result or throw an exception.

return empty result might not be a good idea, since there is something wrong with data, we will need to let the user know. so either way would prefer to throw an exception, and let SiteExtensionController or SIteExtensionManager to catch the exception and return proper error message to user.

shrimpy:
merged in cf18aa8f4d35a7842d8d0d14ee8717880fa9c28a

PR #1536, opened 2015-04-15
remove gcdump support

suwatch:
VS team stops supporting gcdump, hence so do we.   I have separate bug to track removing it from the  binary from webpi feed.


davidebbo:
:shipit:

albertjan:
How are we supposed to do this now? :smile:

davidebbo:
The feature had very low usage in VS and is getting deprecated. Not that regular dumps are still available.

albertjan:
I understand, we solved it by running our application locally, I never needed the feature before but it would have saved us some time today:smile:.

PR #1550, opened 2015-05-05
Enabling a disabled WebJob resulted in the WebJob starting

### LINKED ISSUE #1549: Deployment while WebJob is stopped causes 'Thread was being aborted' error on restart
Steps to reproduce:

-Stop a Continuous WebJob
-Perform deployment
-Restart the Continuous WebJob

The job will start and begin executing code and then will be aborted and restarted. 

It appears as though the code to stop/start a *running* continuous job for a deployment executes after the job is restarted (despite the fact that the job wasn't running during deployment).

amitapl:
after 5 seconds it was stopped and started again.

This is due to 2 things that were fixed in this change:

1. Enable job deleted the disable.job file but also started the job, deleting the file causes the job to restart after 5 seconds.
Fix is to not start the job and have the file change detection do it.

2. Enable / Disable job causes file change twice, once after creating or deleting the disable.job and a second time it seems the directory of the job is changed but this only happens after the job refreshes which causes it to refresh again.
I could pinpoint exactly why this happens but "change" file notification on the job directory is not important and so we will ignore it.

Also added log line for when WebJob files update causes WebJob refresh.

Fixes #1549

### DIFF
--- a/Kudu.Core/Jobs/ContinuousJobRunner.cs
+++ b/Kudu.Core/Jobs/ContinuousJobRunner.cs
@@ -189,8 +189,13 @@ namespace Kudu.Core.Jobs
             }
         }
 
-        public void RefreshJob(ContinuousJob continuousJob, JobSettings jobSettings)
+        public void RefreshJob(ContinuousJob continuousJob, JobSettings jobSettings, bool logRefresh)
         {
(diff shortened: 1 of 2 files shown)


suwatch:
Looks good.

PR #1557, opened 2015-05-14
ProcessController writing extraneous error logs

mathewc:
When determining ProcessInfo, Process controller was tracing errors when it shouldn't be, based on invalid dictionary lookups being done by ProcessExtensions. This issue was reported by a customer in the below forum thread. Fixed the helpers to use TryGetValue.

https://social.msdn.microsoft.com/Forums/en-US/02d4bc1d-06b5-48e5-8fcd-5bde8a495b07/azure-web-apps-responses-empty?forum=windowsazurewebsitespreview

### DIFF
--- a/Kudu.Core/Infrastructure/ProcessExtensions.cs
+++ b/Kudu.Core/Infrastructure/ProcessExtensions.cs
@@ -244,7 +244,14 @@ namespace Kudu.Core.Infrastructure
 
         public static bool GetIsScmSite(Dictionary<string, string> environment)
         {
-            return environment[WellKnownEnvironmentVariables.ApplicationPoolId].StartsWith("~1", StringComparison.OrdinalIgnoreCase);
+            string appPool = null;
+            if (environment.TryGetValue(WellKnownEnvironmentVariables.ApplicationPoolId, out appPool) &&
+                !string.IsNullOrEmpty(appPool))
+            {
+                return appPool.StartsWith("~1", StringComparison.OrdinalIgnoreCase);
+            }
+
+            return false;
         }
 
         public static bool GetIsWebJob(Dictionary<string, string> environment)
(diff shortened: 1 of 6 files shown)


Kudu.Core/Infrastructure/ProcessExtensions.cs
```diff
@@ -244,7 +244,14 @@ public static string GetCommandLine(this Process process)
 
         public static bool GetIsScmSite(Dictionary<string, string> environment)
         {
-            return environment[WellKnownEnvironmentVariables.ApplicationPoolId].StartsWith("~1", StringComparison.OrdinalIgnoreCase);
+            string appPool = null;
+            if (environment.TryGetValue(WellKnownEnvironmentVariables.ApplicationPoolId, out appPool) &&
+                !string.IsNullOrEmpty(appPool))
```
amitapl: `!string.IsNullOrEmpty(appPool)` is probably not needed

amitapl:
looks good
:shipit:

Kudu.Services.Test/Diagnostics/ProcessControllerFacts.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        [Fact]
+        public void SetEnvironmentInfo_NotWebJob_ReturnsExpectedResults()
+        {
+            Dictionary<string, string> environmentVariables = new Dictionary<string, string>();
+            ProcessInfo processInfo = new ProcessInfo
+            {
+                EnvironmentVariables = environmentVariables
+            };
+            _controller.SetEnvironmentInfo(processInfo);
```
mathewc: These tests repro'ed the issue. Before my fix, this codepath was writing errors to ITracer

Kudu.Core/Infrastructure/ProcessExtensions.cs
mathewc: I'm doing that IsNullOrEmpty check because below we're doing a StartsWith - we don't want to null ref.

amitapl: TryGetValue will return false if appPool is null.

mathewc: The dictionary can contain a key, with a null value, and TryGetValue will return it and return true. I'm guarding against that.

amitapl: if environments contains a key with null value its a bug and we would actually want to break on that.

amitapl: anyway, its fine like this.

PR #1559, opened 2015-05-15
WebJobs: Log warning for Continuous jobs w/o Always On enabled

mathewc:
The goal is help customers figure out why their continuous jobs are stopping. We have to tell users alot "make sure you have Always On turned on". Likely users will be able to self diagnose these issues themselves when they see the warning in the log:

*'Always On' doesn't appear to be enabled for this Web App. To ensure your continuous job doesn't stop running when the SCM host is idle for too long, consider enabling 'Always On' in the configuration settings for your Web App. Note: 'Always On' is available only in Basic and Standard modes.*

This relies on Kudu knowing whether Always On is enabled. This will be a new environment variable **WEBSITE_SCM_ALWAYS_ON_ENABLED**. I'm doing the Antares work to have that environment variable set. I won't check this in until that work is done.


Kudu.Core/Environment.cs
```diff
@@ -101,6 +104,7 @@ public class Environment : IEnvironment
             _dataPath = Path.Combine(rootPath, Constants.DataPath);
             _jobsDataPath = Path.Combine(_dataPath, Constants.JobsPath);
             _jobsBinariesPath = Path.Combine(_webRootPath, Constants.AppDataPath, Constants.JobsPath);
+            _alwaysOnEnabled = IsAlwaysOnEnabled();
```
amitapl: I don't think this is the right place to put it, this class is about the kudu environment.
you can just keep it right where you need it.

Kudu.Core/Jobs/ContinuousJobRunner.cs
```diff
@@ -123,6 +127,17 @@ private void StartJob(ContinuousJob continuousJob)
...
         }
 
+        /// <summary>
+        /// If Always On is not enabled, log a warning to the job log on startup.
+        /// </summary>
+        internal void CheckAlwaysOn()
+        {
+            if (!_environment.AlwaysOnEnabled)
+            {
+                _continuousJobLogger.LogWarning(Resources.Warning_EnableAlwaysOn);
```
amitapl: my fear is its going to pollute the logs.
maybe we write it once per log file?

Kudu.Core/Resources.resx
```diff
@@ -216,4 +216,7 @@
   <data name="Log_MaxJobLogLinesReached" xml:space="preserve">
     <value>Reached maximum allowed output lines for this run, to see all of the job's logs you can enable website application diagnostics</value>
   </data>
+  <data name="Warning_EnableAlwaysOn" xml:space="preserve">
+    <value>'Always On' doesn't appear to be enabled for this Web App. To ensure your continuous job doesn't stop running when the SCM host is idle for too long, consider enabling 'Always On' in the configuration settings for your Web App. Note: 'Always On' is available only in Basic and Standard modes.</value>
```
amitapl: Also Premium site sku.

amitapl: But maybe shorten this message? just to say Always On need to be enabled for reiablility

Kudu.Core/Jobs/ContinuousJobRunner.cs
mathewc: Agreed. Is there already an established way to accomplish that?

Kudu.Core/Environment.cs
mathewc: Ok. There isn't a central place where we manage environmental settings, etc.? That's what this looked like to me.

Kudu.Core/Resources.resx
mathewc: Right, thanks.

Kudu.Core/Jobs/ContinuousJobRunner.cs
amitapl: No that would be a first on that (for webjobs logger)

Kudu.Core/Environment.cs
amitapl: Basically we use it where we need it, for example: https://github.com/projectkudu/kudu/blob/master/Kudu.Core/Infrastructure/InstanceIdUtility.cs#L29
https://github.com/projectkudu/kudu/blob/master/Kudu.Services.Web/App_Start/NinjectServices.cs#L546

Kudu.Core/Environment.cs
```diff
@@ -266,6 +270,18 @@ public string SiteExtensionSettingsPath
...
+        public bool AlwaysOnEnabled
+        {
+            get { return _alwaysOnEnabled; }
+        }
+
+        internal bool IsAlwaysOnEnabled()
+        {
+            string value = System.Environment.GetEnvironmentVariable(WebsiteSCMAlwaysOnEnabledKey);
+
+            return string.Compare("1", value) == 0;
```
amitapl: Also I would return true if the environment variable doesn't exist this way you can integrate it now and not wait for the azure change.

mathewc: Mainly I just want to hold on checking this in until the Antares change is made and signed off on, so there is no churn in the env name. That's all. Once I get the review out for that and the name is settled, this can go in. The warning will then only be active once the Antares change goes live.

I don't want to emit this warning until then, so I don't want to default to true in absence of the env value, since them message says "it looks like you don't have Always On". For people who do, I don't want to write that.

amitapl: but true means no warning is emitted (which is also good for non-azure scenario)

mathewc: Understood, thanks. Good pointing out the non-azure scenario.

mathewc:
Updated, addressing CR feedback :)

Kudu.Core/Jobs/ContinuousJobLogger.cs
```diff
@@ -32,6 +32,8 @@ public ContinuousJobLogger(string jobName, IEnvironment environment, ITraceFacto
             ResetLockedStatusFile();
         }
 
+        public event Action RolledLogFile;
```
amitapl: Maybe use EventHandler instead of Action

amitapl:
only one small comment for your discretion , other than that looks good.

Kudu.Core/Jobs/ContinuousJobLogger.cs
mathewc: I used Action because I'd be ignoring the object/sender callback params of EventHandler

PR #1563, opened 2015-05-20
Added support for site extension package that not created from standa…

shrimpy:
Added support for site extension package that not created from standard nuget tool.

Differences: Package that create from nuget.exe won`t have folder as separate metadata, but other zip tools might.


suwatch:
:shipit:

PR #1575, opened 2015-06-04
Turn on libgit2sharp by default, and update it to latest

davidebbo:
We'd had them side by side for ages, and have been running tests in both mode. We should now be able to use it by default.


suwatch:
:shipit:   any hidden goal (eg. try to work around git issue etc.).

davidebbo:
A few goals:

1. I'm hoping that this will reduce the odd of getting that ugly "VirtualAlloc pointer is null" error.
2. Our msysgit is ancient, and it's hard to upgrade it. Would like to move away from it
3. Generally cleaner to call a library than to spawn exe's left and right

PR #1576, opened 2015-06-05
Added default option to not init async lock in SiteExtensionInstallat…

shrimpy:
To avoid folder create->delete->create, added default option to not init async lock in SiteExtensionInstallationLock. 
since "InitializeAsyncLocks" will always try to ensure folder exist, while for site extension installation. if it is a brand new installation, folder would not be there at the first place.


davidebbo:
:shipit:

PR #1577, opened 2015-06-08
Remove dependency on System.Web.Mvc 3.0.0.0, and zip based test

davidebbo:
Long overdue cleanup. We had several tests that relied on having System.Web.Mvc 3.0.0.0 in the GAC, which is no longer supported. This forced us to hack our dev machines to make things work. Now our tests can run with 3.0.0.1, which is what the world is not using.

Also, I yanked the last vestige of tests the were committed to the repo as zip files.


shrimpy:
:shipit:

PR #1578, opened 2015-06-08
Adding schedule for triggered jobs.

amitapl:
This is the ability for a triggered job to use a cron expression to describe a schedule that will run it.
Did some refactoring to the file watcher and job logger code to reuse with the new triggered job scheduler.

Also added a trigger description for each triggered job run.

Currently missing tests.

### DIFF
--- a/Kudu.Core/Jobs/JobsFileWatcher.cs
+++ b/Kudu.Core/Jobs/JobsFileWatcher.cs
@@ -1,0 +1,234 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using Kudu.Contracts.Jobs;
+using Kudu.Contracts.Settings;
+using Kudu.Core.Infrastructure;
+using Kudu.Core.Tracing;
+
+namespace Kudu.Core.Jobs
+{
+    /// <summary>
+    /// Responsible for detecting when a file is changed in a job
+    /// </summary>
+    public class JobsFileWatcher : IDisposable
+    {
+        private readonly ITraceFactory _traceFactory;
+        private readonly IAnalytics _analytics;
+        private const int TimeoutUntilMakingChanges = 5*1000;
+        private const int CheckForWatcherTimeout = 30*1000;
+
+        private HashSet<string> _updatedJobs = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
+        private readonly object _lockObject = new object();
+
+        private Timer _makeChangesTimer;
+        private Timer _startFileWatcherTimer;
+        private FileSystemWatcher _fileSystemWatcher;
+
+        private bool _makingChanges;
+        private readonly string _watchedDirectoryPath;
+        private readonly Action<string> _onJobChanged;
+        private readonly string _filter;
+        private readonly Func<IEnumerable<string>> _listJobNames;
+
+        public JobsFileWatcher(
+            string watchedDirectoryPath,
+            Action<string> onJobChanged,
+            string filter,
+            Func<IEnumerable<string>> listJobNames,
+            ITraceFactory traceFactory,
+            IAnalytics analytics)
+        {
+            _traceFactory = traceFactory;
+            _analytics = analytics;
+            _watchedDirectoryPath = watchedDirectoryPath;
+            _onJobChanged = onJobChanged;
+            _filter = filter;
+            _listJobNames = listJobNames;
+
+            _makeChangesTimer = new Timer(OnMakeChanges);
+            _startFileWatcherTimer = new Timer(StartWatcher);
+            _startFileWatcherTimer.Change(0, Timeout.Infinite);
+
+            FirstTimeMakingChanges = true;
+        }
+
+        public bool FirstTimeMakingChanges { get; private set; }
+
+        private void OnMakeChanges(object state)
+        {
+            HashSet<string> updatedJobs;
+
+            lock (_lockObject)
+            {
+                if (_makingChanges)
+                {
+                    _makeChangesTimer.Change(TimeoutUntilMakingChanges, Timeout.Infinite);
+                    return;
+                }
+
+                _makingChanges = true;
+
+                updatedJobs = _updatedJobs;
+                _updatedJobs = new HashSet<string>();
+            }
+
+            foreach (string updatedJobName in updatedJobs)
+            {
+                try
+                {
+                    _onJobChanged(updatedJobName);
+                }
+                catch (Exception ex)
+                {
+                    _traceFactory.GetTracer().TraceError(ex);
+                }
+            }
+
+            _makingChanges = false;
+            FirstTimeMakingChanges = false;
+        }
+
+        private void StartWatcher(object state)
+        {
+            try
+            {
+                lock (_lockObject)
+                {
+                    // Check if there is a directory we can listen on
+                    if (!FileSystemHelpers.DirectoryExists(_watchedDirectoryPath))
+                    {
+                        // If not check again in 30 seconds
+                        _startFileWatcherTimer.Change(CheckForWatcherTimeout, Timeout.Infinite);
+                        return;
+                    }
+
+                    // Start file system watcher
+                    _fileSystemWatcher = _filter != null ? new FileSystemWatcher(_watchedDirectoryPath, _filter) : new FileSystemWatcher(_watchedDirectoryPath);
+                    _fileSystemWatcher.Created += OnChanged;
+                    _fileSystemWatcher.Changed += OnChanged;
+                    _fileSystemWatcher.Deleted += OnChanged;
+                    _fileSystemWatcher.Renamed += OnChanged;
+                    _fileSystemWatcher.Error += OnError;
+                    _fileSystemWatcher.NotifyFilter = NotifyFilters.CreationTime | NotifyFilters.DirectoryName |
+                                                      NotifyFilters.FileName |
+                                                      NotifyFilters.LastWrite;
+                    _fileSystemWatcher.IncludeSubdirectories = true;
+                    _fileSystemWatcher.EnableRaisingEvents = true;
+
+                    // Refresh all jobs
+                    IEnumerable<string> jobNames = _listJobNames();
+                    foreach (string jobName in jobNames)
+                    {
+                        MarkJobUpdated(jobName);
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                _analytics.UnexpectedException(ex);
+
+                // Retry in 30 seconds.
+                _startFileWatcherTimer.Change(CheckForWatcherTimeout, Timeout.Infinite);
+            }
+        }
+
+        private void OnError(object sender, ErrorEventArgs e)
+        {
+            Exception ex = e.GetException();
+            _traceFactory.GetTracer().TraceError(ex.ToString());
+            ResetWatcher();
+        }
+
+        private void ResetWatcher()
+        {
+            DisposeWatcher();
+
+            _startFileWatcherTimer.Change(CheckForWatcherTimeout, Timeout.Infinite);
+        }
+
+        private void DisposeWatcher()
+        {
+            if (_fileSystemWatcher != null)
+            {
+                _fileSystemWatcher.EnableRaisingEvents = false;
+                _fileSystemWatcher.Dispose();
+                _fileSystemWatcher = null;
+            }
+        }
+
+        public void Stop()
+        {
+            lock (_lockObject)
+            {
+                DisposeWatcher();
+                _startFileWatcherTimer.Change(Timeout.Infinite, Timeout.Infinite);
+            }
+        }
+
+        private void OnChanged(object sender, FileSystemEventArgs e)
+        {
+            string path = e.FullPath;
+            if (path != null && path.Length > _watchedDirectoryPath.Length)
+            {
+                path = path.Substring(_watchedDirectoryPath.Length).TrimStart(Path.DirectorySeparatorChar);
+                int firstSeparator = path.IndexOf(Path.DirectorySeparatorChar);
+                string jobName;
+                if (firstSeparator >= 0)
+                {
+                    jobName = path.Substring(0, firstSeparator);
+                }
+                else
+                {
+                    if (e.ChangeType == WatcherChangeTypes.Changed)
+                    {
+                        return;
+                    }
+
+                    jobName = path;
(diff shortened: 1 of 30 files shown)


Kudu.Core/Jobs/Schedule.cs
```diff
@@ -0,0 +1,65 @@
...
+                return crontabSchedule != null ? new Schedule(crontabSchedule, logger) : null;
+            }
+            catch (Exception ex)
+            {
+                logger.LogError("Failed to parse schedule \"{0}\" with error - {1}".FormatCurrentCulture(cronExpression, ex.Message));
+                return null;
+            }
+        }
+
+        public TimeSpan GetNextInterval(DateTime lastSchedule, bool ignoreMissed = false)
```
amitapl: Not sure why, but all comments are gone.
I did address them, like here.

amitapl:
Found the commit with the comments here: https://github.com/projectkudu/kudu/commit/8b44a28

amitapl:
Added tests

Kudu.Core/Jobs/JobsFileWatcher.cs
```diff
@@ -0,0 +1,231 @@
...
+using System.Text;
+using System.Threading;
+using Kudu.Contracts.Jobs;
+using Kudu.Contracts.Settings;
+using Kudu.Core.Infrastructure;
+using Kudu.Core.Tracing;
+
+namespace Kudu.Core.Jobs
+{
+    public class JobsFileWatcher : IDisposable
```
mathewc: It would be really great if for each of these classes you'd add code comments detailing what the class responsibilities are.

Kudu.Core/Jobs/Schedule.cs
```diff
@@ -0,0 +1,65 @@
...
+        public static Schedule BuildSchedule(string cronExpression, TriggeredJobSchedulerLogger logger)
+        {
+            try
+            {
+                var crontabSchedule = CrontabSchedule.Parse(cronExpression, new CrontabSchedule.ParseOptions() {IncludingSeconds = true});
+                return crontabSchedule != null ? new Schedule(crontabSchedule, logger) : null;
+            }
+            catch (Exception ex)
+            {
+                logger.LogError("Failed to parse schedule \"{0}\" with error - {1}".FormatCurrentCulture(cronExpression, ex.Message));
```
mathewc: Nit: "Failed to parse schedule "{0}". Error: {1}"

Kudu.Core/Jobs/Schedule.cs
```diff
@@ -0,0 +1,65 @@
...
+            DateTime nextOccurrence = _crontabSchedule.GetNextOccurrence(lastSchedule);
+
+            // If next occurence is in the future use it
+            if (nextOccurrence >= now)
+            {
+                return nextOccurrence - now;
+            }
+
+            // Otherwise if next occurence is up to 10 minutes in the past or ignore missed is true use now
+            if (ignoreMissed || nextOccurrence >= now - TimeSpan.FromMinutes(10))
```
mathewc: You might consider making this configurable - it could be a setting in the job file, next to the schedule expression. We could default it to some value, but people can override it.

Kudu.Core/Jobs/Schedule.cs
```diff
@@ -0,0 +1,65 @@
...
+            {
+                return TimeSpan.Zero;
+            }
+
+            _logger.LogWarning("Missed at least one schedule at " + nextOccurrence);
```
mathewc: Nit: "Missed at least one schedule occurrence at {0}"

Note that the CrontabSchedule allows you to compute the number of missed schedule occurrences (CrontabSchedule.GetNextOccurrences). You could log that, rather than "at least one" if you wanted to.

Kudu.Core/Jobs/TriggeredJobSchedule.cs
```diff
@@ -0,0 +1,83 @@
+using System;
+using System.Linq.Expressions;
+using System.Threading;
+using Kudu.Contracts.Jobs;
+using Kudu.Core.Tracing;
+
+namespace Kudu.Core.Jobs
+{
+    public class TriggeredJobSchedule : IDisposable
```
mathewc: Add class comments :)

Kudu.Core/Jobs/TriggeredJobSchedule.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        public TriggeredJobSchedulerLogger Logger { get; private set; }
+
+        public void Reschedule(DateTime lastRun, Schedule schedule = null)
+        {
+            Schedule = schedule ?? Schedule;
+
+            var nextInterval = Schedule.GetNextInterval(lastRun);
+
+            // Limit next interval to 40 days so it won't conflict with the maximum time allowed to set a timer.
```
mathewc: We need to handle this, not silently ignore it (which looks like what you're doing?) The way I'm planning on handling this in the SDK Extensions timer (I'm using System.Timers.Timer, so the max interval there is 30 days) is to set an additional timer for 29 days, and when that fires, I'll simply reschedule my job interval, starting from the new now. I think you should do the same.

Kudu.Core/Jobs/TriggeredJobSchedule.cs
```diff
@@ -0,0 +1,83 @@
...
+            nextInterval = nextInterval < MaximumTimeSpanInterval ? nextInterval : MaximumTimeSpanInterval;
+
+            if (schedule != null)
+            {
+                Logger.LogInformation("Scheduling WebJob with " + schedule + " next schedule expected in " + nextInterval);
+            }
+            else
+            {
+                Logger.LogInformation("Next schedule expected in " + nextInterval);
```
mathewc: Next schedule occurrence expected in

mathewc: Also, for both of these log messages, rather than output the interval, it's probably more useful to the user for you to ouput the time of the next occurrence?

Kudu.Core/Jobs/Schedule.cs
amitapl: Agree - added an issue https://github.com/projectkudu/kudu/issues/1580

Kudu.Core/Jobs/Schedule.cs
amitapl: Nice, I'll use that

Kudu.Core/Jobs/TriggeredJobSchedule.cs
amitapl: This is handled here, after 40 days the timer will fire but since the schedule is not up yet it will reschedule again.

Kudu.Core/Jobs/TriggeredJobSchedule.cs
amitapl: Since we log the current time I think logging the interval is better

Kudu.Core/Jobs/TriggeredJobsScheduler.cs
```diff
@@ -0,0 +1,153 @@
...
+using System.Linq;
+using System.Threading;
+using Kudu.Contracts.Jobs;
+using Kudu.Core.Hooks;
+using Kudu.Core.Infrastructure;
+using Kudu.Core.Tracing;
+
+namespace Kudu.Core.Jobs
+{
+    public class TriggeredJobsScheduler : IDisposable
```
mathewc: Perhaps this class should be renamed to ScheduledJobManager? Then rename TriggeredJobSchedule to ScheduledJob?

Then you have types Schedule, ScheduledJob, and ScheduledJobManager, which is conceptually clearer I think than: Schedule, TriggeredJobSchedule, TriggeredJobsScheduler

Thoughts?

amitapl: A schedule is an added property on top of a triggered job so I think it makes sense as it is.

Kudu.Core/Jobs/TriggeredJobsScheduler.cs
```diff
@@ -0,0 +1,153 @@
...
+                }
+                else
+                {
+                    triggeredJobSchedule.Reschedule(lastRun);
+                    return;
+                }
+            }
+            catch (ConflictException)
+            {
+                // Ignore as this is expected when running multiple instances
```
mathewc: Is there a way for you to ensure that the timer is only firing on one instance? Otherwise, if you have 5 instances and a 5 second timer, every five seconds all instances are racing to perform file operations and are conflicting with each other. It's better if the timer is only running in one place.

In the SDK extensions, I'm going to do this using the new Singleton feature I'm working on. Each instance will attempt to take a singleton log (blob lease under the covers) and only one will win. Only the winner is running the timer.

Kudu.FunctionalTests/Jobs/WebJobsTests.cs
```diff
@@ -285,6 +288,95 @@ public void TriggeredJobTriggers()
         }
 
         [Fact]
+        public void ScheduledTriggeredJobTriggers()
```
mathewc: This test name is unclear to me

Kudu.Core/Jobs/TriggeredJobsScheduler.cs
amitapl: yes, the triggered job can only be invoked on a single instance, the lock is already in place.

Kudu.FunctionalTests/Jobs/WebJobsTests.cs
```diff
@@ -285,6 +288,95 @@ public void TriggeredJobTriggers()
...
+            RunScenario("ScheduledTriggeredJobTriggers", appManager =>
+            {
+                const string jobName = "job1";
+
+                TestTracer.Trace("Copying the script to the triggered job directory");
+
+                appManager.JobsManager.CreateTriggeredJobAsync(jobName, "run.cmd", JobScript).Wait();
+
+                // Wait for file watcher to pick up
+                Thread.Sleep(30 * 1000);
```
mathewc: This test really waits for 30 seconds? I don't think we should do that.

Kudu.FunctionalTests/Jobs/WebJobsTests.cs
amitapl: cleared it hopefully

Kudu.FunctionalTests/Jobs/WebJobsTests.cs
amitapl: I think its fine, it is better to have a test than not have a test even at the cost of time

Kudu.FunctionalTests/Jobs/WebJobsTests.cs
```diff
@@ -12,12 +12,15 @@
 using Kudu.Client.Infrastructure;
```
mathewc: Need lots more tests. Perhaps some/most of these should be unit tests. Tests for:

- TriggeredJobSchedule class and all the logic in Reschedule
- TriggeredJobsScheduler class (lots going on in there)

Your end to end functional shows a happy path, but I want to make sure we have test coverage for all the rest.

amitapl: ok, I'll open an issue to improve test overage

Kudu.Services/Jobs/JobsController.cs
```diff
@@ -179,7 +179,7 @@ public HttpResponseMessage InvokeTriggeredJob(string jobName, string arguments =
         {
             try
             {
-                _triggeredJobsManager.InvokeTriggeredJob(jobName, arguments);
+                _triggeredJobsManager.InvokeTriggeredJob(jobName, arguments, "External - " + Request.Headers.UserAgent);
```
mathewc: What does the trigger value string for your scheduled jobs?

amitapl: Schedule - {the schedule}

Kudu.Core/Jobs/TriggeredJobSchedule.cs
mathewc: I see - there is interaction between this code, and the code in TriggeredJobsScheduler.OnSchedule. Definitely not obvious, at least to me. Also, it looks like that logic might be faulty. Currently you're checking the schedule again (as of now), but what if due to delay, it returns a negative interval (you're testing against TimeSpan.Zero). Wouldn't that mean you miss an interval? 

Please add unit tests for this logic.

Kudu.Core/Jobs/TriggeredJobSchedule.cs
mathewc: Sure I guess. The user will have to add - probably just personal preference.

Kudu.Core/Jobs/TriggeredJobsScheduler.cs
mathewc: This code looks like it runs on each schedule or timer fire. Doesn't that mean the timer is running  then on all instances?

amitapl: yes, the timer does but only one of them will invoke the triggered job

Kudu.FunctionalTests/Jobs/WebJobsTests.cs
mathewc: I wasn't suggesting not to have a test of course. I was asking whether you could use a better wait strategy. Initially I thought this test was running against local file system, which would mean file events would be very fast.

Kudu.Core/Jobs/TriggeredJobSchedule.cs
amitapl: The schedule cannot return a negative timespan.
logic is fine.

Kudu.FunctionalTests/Jobs/WebJobsTests.cs
mathewc: I think you should get unit test coverage complete for all the code you're about to check in, before you check in, so we don't have test debt right out of the gate.

Kudu.FunctionalTests/Jobs/WebJobsTests.cs
amitapl: It is running against local webserver or remote azure website.

Kudu.FunctionalTests/Jobs/WebJobsTests.cs
amitapl: noted

amitapl:
Added tests for schedule.

PR #1592, opened 2015-06-17
Adjust kudu to allow OneDrive continuous deployment work with Azure E2E

shrimpy:


suwatch:
Use key word Azure (not Antares) in commit and PR.

Kudu.Services/FetchHelpers/OneDriveHelper.cs
```diff
@@ -356,7 +359,14 @@ private void HandlingDeletion(OneDriveModel.OneDriveChange change, string wwwroo
             if (!change.IsDeleted)
                 return;
 
-            string fullPath = Path.Combine(wwwroot, change.Path);
+            string changePath = RemoveRootFromPath(change.Path);
```
suwatch: I would introduce new api GetDestinationPath(string wwwroot, string changePath);

shrimpy: sounds good too. let me update it.

PR #1593, opened 2015-06-18
Update SiteBuilderFactory.cs

regisbsb:
Check for ASP.NET 5 before WebSiteProject

### DIFF
--- a/Kudu.Core/Deployment/Generator/SiteBuilderFactory.cs
+++ b/Kudu.Core/Deployment/Generator/SiteBuilderFactory.cs
@@ -194,6 +194,17 @@ namespace Kudu.Core.Deployment.Generator
             {
                 return DetermineProject(repositoryRoot, projects[0], perDeploymentSettings, fileFinder);
             }
+            
+            // Check for ASP.NET 5 project without VS solution or project
+            string projectJson;
+            if (AspNet5Helper.TryAspNet5Project(targetPath, out projectJson))
+            {
+                return new AspNet5Builder(_environment,
+                                           perDeploymentSettings,
+                                           _propertyProvider,
+                                           repositoryRoot,
+                                           projectJson);
+            }
 
             if (tryWebSiteProject)
             {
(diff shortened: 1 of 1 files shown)


Kudu.Core/Deployment/Generator/SiteBuilderFactory.cs
```diff
@@ -194,6 +194,17 @@ private ISiteBuilder ResolveProject(string repositoryRoot, string targetPath, ID
...
+            // Check for ASP.NET 5 project without VS solution or project
+            string projectJson;
+            if (AspNet5Helper.TryAspNet5Project(targetPath, out projectJson))
+            {
+                return new AspNet5Builder(_environment,
+                                           perDeploymentSettings,
+                                           _propertyProvider,
+                                           repositoryRoot,
+                                           projectJson);
+            }
```
suwatch: @ahmelsayed do you see any regression by doing TryAspNet5Project before WebSiteProject?   In case, they are not mutually exclusive.

regisbsb: @suwatch @ahmelsayed  I think if don't do that the SLN is found first and then the Project is deemed as Website, even thought it is a ASP.NET 5 when you use the "Project" setting in the Azure portal.

ahmelsayed: @suwatch @regisbsb I do think it makes sense to check ASP.NET 5 before WebsiteProject since an ASP.NET 5 project could include another `csproj` while a normal Website would never include a `project.json`

suwatch: OK, then.   @ahmelsayed, could you take care merge this change in?

ahmelsayed:
Merged in 8b88341389b05e40675789865f6343fac1583778. Thanks @regisbsb  for your contribution :)

regisbsb:
No worries.

PR #1594, opened 2015-06-19
expand command and directory before executing

suwatch:
I made a surgical change targeted only this scenario.


shrimpy:
:shipit:

PR #1605, opened 2015-06-30
ASP.NET 5 beta5

### LINKED ISSUE #1604: Default custom deployment script doesn't work for aspnet 5
[Issue found by @Radek44]

Problem is that is sets `SCM_DNX_VERSION` as part of the AspNet5Builder logic. The minute you go custom, that logic no longer happens and it becomes broken.

Fix would be to bake that variable in the generated script.

ahmelsayed:
* Support ASP.NET 5 beta5
* Fix #1604 where ASP.NET5 deployment script doesn't work if it was a custom deployment script. Goes with https://github.com/projectkudu/KuduScript/pull/26


davidebbo:
:shipit:

PR #1612, opened 2015-07-07
Fix to force IE to not use compatibility mode

theadriangreen:
By default IE displays intranet sites in compatibility mode. The SCM site appears broken when it is displayed in compatibility mode. Many OnPrem deployments would be run on the local intranet, hence by default, the SCM sites for those deployments would appear to be broken. This PR fixes that behavior.
TFS work item ID: 4285419

### DIFF
--- a/Kudu.Services.Web/Web.config
+++ b/Kudu.Services.Web/Web.config
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <!--
   For more information on how to configure your ASP.NET application, please visit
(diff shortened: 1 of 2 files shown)


davidebbo:
My concern here is that this ends up adding this header for all requests (API calls). Instead, it may be better to limit ourselves to UI pages using meta tag. e.g. try adding this to `_layout.cshtml`:

    <meta http-equiv="X-UA-Compatible" content="IE=edge" />

It should have the same effect, and not affect anything non-HTML.

theadriangreen:
Thanks for the input David, I have updated the PR to change _layout.cshtml

davidebbo:
Ok, it's in, thanks!

PR #1615, opened 2015-07-09
Enable ListDeployment to work on Read-Only file system

shrimpy:


Kudu.Core/Infrastructure/FileSystemHelpers.cs
```diff
@@ -290,6 +289,27 @@ public static void DeleteDirectoryContentsSafe(string path, bool ignoreErrors =
...
+        public static bool IsFileSystemReadOnly(IEnvironment environment)
+        {
+            try
+            {
+                // Temp folder will be clean up everytime site starts up
+                // So it is fine to have leftover folder if deletion failed
+                string folderName = Path.Combine(environment.DataPath, "Temp", Guid.NewGuid().ToString("N"));
+                Instance.Directory.CreateDirectory(folderName);
+                Instance.Directory.Delete(folderName);
+                return false;
```
davidebbo: Assuming site must restart for readonly/writable state to change, we can cache this state so we avoid too many new i/o operations, which could have side effects. We probably take the lock quite often.

shrimpy: good point! I will wait for Kranthi to confirm, so that can decide the expiration time of the cache. if it require a restart then there is no need for any expiration.

shrimpy: we will still need a expiration time if FS is in RW mode, so that we do not introduce too many I/O.
working on the improvement.

davidebbo:
:shipit:

Kudu.Core/Infrastructure/LockFile.cs
```diff
@@ -247,6 +273,42 @@ private void OnLockReleasedInternal(object sender, FileSystemEventArgs e)
             }
         }
 
+        private static bool IsFileSystemReadOnly()
```
shrimpy: adding test case for this new logic

PR #1626, opened 2015-07-21
Update gitignore file to ignore .vs folder from VS2015

shrimpy:


ahmelsayed:
:+1: Can we use C# 6 features then? :)

shrimpy:
i think so, sooner or later everyone will use VS2015.

davidebbo:
Well, we can't use C# 6 features until everyone involved in Kudu development moves up. We'll get there, but let's give it a bit of time, unless there are life changing new features we must have now :)

PR #1628, opened 2015-07-21
Added ProfileManager.cs and some cleanup. plus added few tests, the r…

safihamid:
…est of the tests will be added in Kudu private


Kudu.Services/Diagnostics/ProcessController.cs
```diff
@@ -22,19 +22,12 @@
 using System.Threading.Tasks;
 using System.Text;
 using System.Threading;
+using Kudu.Services.Diagnostics;
```
ahmelsayed: Remove unused usings 

``` csharp
using System.Text;
using System.Threading;
using Kudu.Services.Diagnostics;
```

Kudu.Services/Diagnostics/ProfileManager.cs
```diff
@@ -0,0 +1,314 @@
...
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using System.Net;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Web.Http;
+
+namespace Kudu.Services.Performance
```
ahmelsayed: I think this should be in `Kudu.Core.Diagnostics`. @davidebbo ?

Kudu.Services/Diagnostics/ProfileManager.cs
```diff
@@ -0,0 +1,314 @@
...
+        }
+
+        private static int GetNextProfilingSessionId()
+        {
+            // TODO: This is not a good way to track active profiling sessions, but as of VS2015 RTM, the profiling service does not provide any API to track the current active sessions.  
+            // This is planned to be fixed in VS2015 Update 1.
+            var r = new Random();
+
+            int sessionId = r.Next(1, 255);
+            List<ProfileInfo> list = _profilingList.Values.ToList<ProfileInfo>();
```
ahmelsayed: I actually see why @shrimpy said use a HashSet now :)
``` csharp
var lookUp = new HashSet<int>(_profilingList.Values.Select(v => v.SessionId));

while (lookUp.Contains(sessionId))
{
    sessionId = r.Next(1, 255);
}
```

Kudu.Services/Diagnostics/ProfileManager.cs
```diff
@@ -0,0 +1,314 @@
...
+        private static ConcurrentDictionary<int, ProfileInfo> _profilingList = new ConcurrentDictionary<int, ProfileInfo>();
+        private static object _lockObject = new object();
+
+        // The profiling session timeout, this is temp fix before VS2015 Update 1.
+        private static readonly TimeSpan _profilingTimeout = TimeSpan.FromMinutes(15);
+        private static Timer _profilingIdleTimer;
+
+        private static string _processName = System.Environment.ExpandEnvironmentVariables("%SystemDrive%\\msvsmon\\profiler\\VSStandardCollector.Dev14.exe");
+
+        internal static async Task<HttpResponseInfo> StartProfileAsync(int processId, ITracer tracer = null)
```
ahmelsayed: I don't think this class should return `HttpResponseInfo`, The class doesn't need to take any dependency on HTTP. It should return a value in success and throwing exceptions on errors.

Kudu.Services/Diagnostics/ProfileManager.cs
```diff
@@ -0,0 +1,314 @@
...
+                    // TODO: VS 2015 Update 1 should have a better way to handle this.
+                    foreach (var item in _profilingList)
+                    {
+                        if (DateTime.UtcNow - item.Value.StartTime > _profilingTimeout)
+                        {
+                            Task.Run(() => StopProfileInternalAsync(item.Key, item.Value.SessionId, true));
+                        }
+                    }
+
+                    DisposeTimerIfNecessary();
```
ahmelsayed: since this is awaited `Task.Run(() => StopProfileInternalAsync(item.Key, item.Value.SessionId, true));` There is a race condition between that and this line.

Kudu.Services/Diagnostics/ProfileManager.cs
```diff
@@ -0,0 +1,314 @@
...
+            {
+                func();
+            }
+            catch (Exception)
+            {
+                // no-op
+            }
+        }
+
+        internal class HttpResponseInfo
```
ahmelsayed: I don't think we should need to define a custom return type to control flow. We should use exceptions for error cases and values for success cases.

safihamid: As we discussed I don't like to throw exceptions to control flow. even for error cases because these are application errors.

Kudu.Services/Diagnostics/ProfileManager.cs
safihamid: per my other comment we shouldn't throw exceptions to control flow. here is an example:

                // check if the profiling is running for the given process. If it doesn't return 404.
                if (!_profilingList.ContainsKey(processId))
                {
                    return new HttpResponseInfo(HttpStatusCode.NotFound, string.Format("Profiling for process '{0}' is not running.", processId));
                }

why would we throw exception here?

safihamid: This class is only making a dependency on a an HTTP enum for status codes.

Kudu.Services/Diagnostics/ProfileManager.cs
```diff
@@ -0,0 +1,314 @@
...
+            }
+        }
+
+        internal static string GetProfilePath(int processId)
+        {
+            string profileFileName = string.Format("profile_{0}_{1}_{2}.diagsession", InstanceIdUtility.GetShortInstanceId(), System.Diagnostics.Process.GetProcessById(processId).ProcessName, processId);
+            return System.Environment.ExpandEnvironmentVariables("%LOCAL_EXPANDED%\\Temp\\" + profileFileName);
+        }
+
+        internal static bool IsProfileRunnig(int processId)
```
mathewc: Method name is mispelled

Kudu.Services/Diagnostics/ProfileManager.cs
mathewc: I agree that this class shouldn't be returning anything HttpResponse related. It should return generic success/failure info, and callers decide how to interpret.

Kudu.Services/Diagnostics/ProfileManager.cs
```diff
@@ -0,0 +1,314 @@
...
+using System.Linq;
+using System.Net;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Web.Http;
+
+namespace Kudu.Services.Performance
+{
+    internal static class ProfileManager
```
mathewc: I don't think this should be a static class. I think this should be an instance service following the pattern of all our other services - see App_Start\NinjectServices.cs, where we new them all up and add to the DI container. Then, JobsController just takes it as a ctor param like all the other services.

Kudu.Services/Diagnostics/ProfileManager.cs
```diff
@@ -0,0 +1,314 @@
...
+        // The profiling session timeout, this is temp fix before VS2015 Update 1.
+        private static readonly TimeSpan _profilingTimeout = TimeSpan.FromMinutes(15);
+        private static Timer _profilingIdleTimer;
+
+        private static string _processName = System.Environment.ExpandEnvironmentVariables("%SystemDrive%\\msvsmon\\profiler\\VSStandardCollector.Dev14.exe");
+
+        internal static async Task<HttpResponseInfo> StartProfileAsync(int processId, ITracer tracer = null)
+        {
+            tracer = tracer ?? NullTracer.Instance;
+            using (tracer.Step("ProcessController.StartProfileAsync"))
```
mathewc: This class shouldn't have any "ProcessController" references in it

Kudu.Services/Diagnostics/ProfileManager.cs
mathewc: I agree with Ahmed that we shouldn't need a new wrapper type for this. Particularly in light of the other feedback that ProcessManager shouldn't be returning http status info anyways. It would likely instead have a TryStartProfilingAsync method

Kudu.Services/Diagnostics/ProcessController.cs
```diff
@@ -241,90 +233,41 @@ public HttpResponseMessage MiniDump(int id, int dumpType = 0, string format = nu
...
-                {
-                    _tracer.TraceError("Too many profiling sessions are currently running.");
-                    return Request.CreateErrorResponse(HttpStatusCode.ServiceUnavailable, "Too many profiling sessions are currently running. Please try again later.");
-                }
-
-                int profilingSessionId = GetNextProfilingSessionId();
-
-                string processName = System.Environment.ExpandEnvironmentVariables("%SystemDrive%\\msvsmon\\profiler\\VSStandardCollector.Dev14.exe");
-                string arguments = System.Environment.ExpandEnvironmentVariables(string.Format("start {0} /attach:{1} /loadAgent:4EA90761-2248-496C-B854-3C0399A591A4;DiagnosticsHub.CpuAgent.dll  /scratchLocation:%LOCAL_EXPANDED%\\Temp", profilingSessionId, id));
+                var result = await ProfileManager.StartProfileAsync(process.Id, _tracer);
```
mathewc: TryStartProfileAsync returning a bool?

Kudu.Services/Diagnostics/ProcessController.cs
```diff
@@ -241,90 +233,41 @@ public HttpResponseMessage MiniDump(int id, int dumpType = 0, string format = nu
...
-                    {
-                        profilingSessionId = _profilingList[process.Id].SessionId;
-                    }
-                }
-
-                string processName = System.Environment.ExpandEnvironmentVariables("%SystemDrive%\\msvsmon\\profiler\\VSStandardCollector.Dev14.exe");
-                string profileFileName = string.Format("profile_{0}_{1}_{2}.diagsession", InstanceIdUtility.GetShortInstanceId(), process.ProcessName, process.Id);
-                string profileFileFullPath = System.Environment.ExpandEnvironmentVariables("%LOCAL_EXPANDED%\\Temp\\" + profileFileName);
-                string arguments = string.Format("stop {0} /output:{1}", profilingSessionId, profileFileFullPath);
+                var result = await ProfileManager.StopProfileAsync(process.Id, _tracer);
```
mathewc: TryStopProfileAsync?

Kudu.Services/Diagnostics/ProfileManager.cs
```diff
@@ -0,0 +1,314 @@
+﻿using Kudu.Contracts.Tracing;
```
mathewc: Please add a unit test class for ProfileManager that exercises and tests all its methods and logic

Kudu.Services/Diagnostics/ProfileManager.cs
safihamid: @mathewc not sure what you mean here. ProcessController is added in NinjectServices and ProfileManager is just static methods being called from ProcessController which require no instance objects.

Kudu.Services/Diagnostics/ProfileManager.cs
safihamid: @mathewc this class has dependencies on some executables that need to be running on server. I was gonna add more test scenarios in Antares hosting for these, in there we could create sites, start/stop profiling etc.

Kudu.Services/Diagnostics/ProfileManager.cs
davidebbo: *mispelled* is misspelled ;)

Kudu.Services/Diagnostics/ProfileManager.cs
davidebbo: I just don't understand the use of random numbers here. Why not simply increment a number to get sequential IDs?

Kudu.Services/Diagnostics/ProfileManager.cs
mathewc: There may be some things you can't actually test, like starting the actual profile, but all the other logic in that class should be made testable in some form, so we can confidently make changes and have tests that allow us to step through the code, catch regressions, etc.

Kudu.Services/Diagnostics/ProfileManager.cs
safihamid: The profiling service is not in the sandbox and you could possibly have more than one kudu sites running on the same VM and both trying to do profiling. you can also start profiling sessions from VS. creating random numbers makes it less likely to have multiple profiling sessions from different sources to come up with same session numbers. though not 100% guaranteed and you may still get errors and need to retry starting a profiling session.

This issue has been communicated and understood by profiler team and in VS 2015 Update 1 we will pass a GUID for session and all this logic will go away.

davidebbo: SO with this code, if you pick a random number that is not being used in the _profilingList, there is still a chance that it is in fact used in another sandbox? If so, what's the point of even keeping track of numbers used in this one sandbox?

Also, what is the valid range of IDs? Is 255 the highest valid value?

safihamid: Yes but we are just significantly reducing the chances per recommendation.
and yes 255 is the highest valid value.

Kudu.Services/Diagnostics/ProfileManager.cs
mathewc: My point is that we have an established pattern for services, and this ProfileManager is another service peer to ITriggeredJobManager, IContinuousJobManager, etc. So I'd make it follow the same model - it's an instance type with instance methods, and a singleton instance is added to the DI container on startup. With an IProfileManager interface (like all our other services), the controller becomes much more easily testable.

Kudu.Services/Diagnostics/ProfileManager.cs
davidebbo: I guess we can live with this as long as there is a plan to clean up later. Do change to HashSet which is O(1) lookup.

Kudu.Services/Diagnostics/ProcessController.cs
safihamid: I like the idea of Try... pattern and we avoid throwing exceptions. will change the code.

Kudu.Services/Diagnostics/ProfileManager.cs
safihamid: Changing the return type name to ProfileResultInfo per our discussion.

Kudu.Services/Diagnostics/ProcessController.cs
safihamid: leaving it as is per our conversation. async methods can't return out arguments.

safihamid:
ping ping, I am not aware of any outstanding issues. I have added some basic unit tests and Matt was going to see what he could do to make these paths more unit testable when he gets a chance. In the meantime I am adding more E2E tests in Hosting platform. 
There is currently a CRI that will be addressed by this pull request: Customer not able to get a profile if CPU is constantly at 100% on a small VM(One core).

mathewc:
I have no comments other than those I've already left in the PR. There is a bunch of code in this PR that doesn't have tests in the Kudu reop (e.g. idle timer stopping of profiling, threading/locking in ProfileManager, exit code processing/error handling, max # of profile logic, etc.). We really need to ensure we have unit test coverage for everything we check in, otherwise we can't maintain it / make fixes with confidence.

Kudu.Services/Diagnostics/ProfileManager.cs
```diff
@@ -0,0 +1,316 @@
...
+using System.Net;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Kudu.Services.Performance
+{
+    internal static class ProfileManager
+    {
+        private const int ProcessExitTimeoutInSeconds = 300;
```
ahmelsayed: Can the profiler process stay idle for 5 minutes? Note that the Azure front end will kill an idle request after 3 minutes.

ahmelsayed:
I'm still not entirely convinced that a custom [`ProfileResultInfo`](https://github.com/projectkudu/kudu/pull/1628/files#diff-ea1eb1f565f23ba94d36f89ed90d4e88R290) object that wraps an http status codes and response message is needed here. I think `ProfileManager` should be completely HTTP agnostic and shouldn't require a custom `Result` object, that's said I'll leave it for someone else to make the call.

Kudu.Services/Diagnostics/ProfileManager.cs
safihamid: It's not really idle, if the CPU is at 100% on a small VM(single core) it will take a long time for the process to do its job and exit. I have been able to repro this and was taking around 120+ seconds. so I chose an upper limit. if azure has a 3 minute timeout I can change the code to 180 seconds then,

ahmelsayed: I see, I was just curious as to why the timeout jumped from 20 to 300 seconds. It looks fine.

davidebbo:
Merged.

Issue #1553, opened 2015-05-08
Making sure Kudu behave reasonably in Read-only env


suwatch:
There is a situation in Azure where the file system may be put to R/O mode.   We just need to make sure that Kudu functions reasonably in that env.   To name a few,

  - web job continue to function.   may be missing some log
  - browse to SCM site + its CO. (Debug Console/ProcessExplorer..) should be functioning.
  - REST GET apis should be functioning.   We can review case by case.

shrimpy:
[Fixing List Deployment operation](https://github.com/projectkudu/kudu/pull/1615)

shrimpy:
With fix https://github.com/projectkudu/kudu/commit/5d306cca27a25d1a1dd18c57da545f0d5fb0a9f8, for read-only action that require file lock, they should working as expect. 

next will work on webjob!

shrimpy:
we are good with continuous job, if the app crash, we already [had code to handle re-run it](https://github.com/projectkudu/kudu/blob/master/Kudu.Core/Jobs/ContinuousJobRunner.cs#L79).

shrimpy:
scheduled trigger job will also recover. we can close this issue now.

PR #1630, opened 2015-07-22
Return 503 for invoking trigger if file system is RO.

shrimpy:


shrimpy:
improvement for https://github.com/projectkudu/kudu/issues/1553

Kudu.Services/Jobs/JobsController.cs
```diff
@@ -179,6 +179,12 @@ public HttpResponseMessage InvokeTriggeredJob(string jobName, string arguments =
         {
             try
             {
+                if (FileSystemHelpers.IsFileSystemReadOnly())
```
davidebbo: Could this be moved to a `catch` block instead, so it doesn't affect the happy path?

Kudu.Core/Infrastructure/FileSystemHelpers.cs
```diff
@@ -290,6 +293,27 @@ public static void DeleteDirectoryContentsSafe(string path, bool ignoreErrors =
...
+                return false;
+            }
+
+            try
+            {
+                string folder = Path.Combine(TmpFolder, Guid.NewGuid().ToString());
+                CreateDirectory(folder);
+                DeleteDirectorySafe(folder, ignoreErrors: false);
+                return false;
+            }
```
mathewc: Can this value be cached? How often might this change, or can we safely assume that after we check once, for the lifetime of the app domain it won't change?

Kudu.Services.Test/Jobs/InvokeTriggeredJobReadonlyFileSystemTests.cs
```diff
@@ -0,0 +1,39 @@
...
+using System;
+using System.IO.Abstractions;
+using System.Net;
+using System.Net.Http;
+using Xunit;
+
+namespace Kudu.Services.Test.Jobs
+{
+    [KuduXunitTestClass]
+    public class InvokeTriggeredJobReadonlyFileSystemTests
```
mathewc: Since this test is testing a controller method of JobsController, I think it should be moved to an existing JobController  test class (hopefully one exists), otherwise create a new test class for JobsController. Then, the test method should be named something like InvokeTriggeredJob_ReturnsServiceUnavailable_WhenFileSystemIsReadOnly

Kudu.Core/Infrastructure/LockFile.cs
```diff
@@ -287,27 +285,6 @@ private void OnLockReleasedInternal(object sender, FileSystemEventArgs e)
             }
         }
 
-        private static bool IsFileSystemReadOnly()
```
mathewc: Is there a test class for FileSystemHelpers? If so, please add a new unit test for your new method. If a test class doesn't exist, please create one :)

Kudu.Services.Test/Jobs/InvokeTriggeredJobReadonlyFileSystemTests.cs
```diff
@@ -0,0 +1,39 @@
...
+                Mock.Of<ITriggeredJobsManager>(),
+                Mock.Of<IContinuousJobsManager>(),
+                Mock.Of<ITracer>());
+
+            var fileSystem = new Mock<IFileSystem>();
+            var dirBase = new Mock<DirectoryBase>();
+            fileSystem.Setup(f => f.Directory).Returns(dirBase.Object);
+            dirBase.Setup(d => d.CreateDirectory(It.IsAny<string>())).Throws<UnauthorizedAccessException>();
+            FileSystemHelpers.Instance = fileSystem.Object;
+            FileSystemHelpers.TmpFolder = @"D:\";
```
mathewc: This test will only run on machines that have a D drive? Do we make such assumptions in other tests?

Kudu.Services.Test/Jobs/InvokeTriggeredJobReadonlyFileSystemTests.cs
```diff
@@ -0,0 +1,39 @@
...
+
+            controller.Request = new HttpRequestMessage();
+
+            HttpResponseMessage resMsg = controller.InvokeTriggeredJob("foo");
+            Assert.Equal(HttpStatusCode.ServiceUnavailable, resMsg.StatusCode);
```
mathewc: Perhaps add an additional assertion to this test, where you change the mock to be not read only, and make the request again, expecting 202 Accepted? That way we're testing both paths.

Kudu.Services/Jobs/JobsController.cs
shrimpy: yep... updated

Kudu.Core/Infrastructure/FileSystemHelpers.cs
shrimpy: will be good not to cached it.  since there will be a cause that file system will change to different mod silently. and we only expect file system will stay in RO mod in a very short period. 

and this method shouldn`t be invoke often, or only use when there is exception.

mathewc: Got it, thanks.

Kudu.Services.Test/Jobs/InvokeTriggeredJobReadonlyFileSystemTests.cs
shrimpy: right now tests are run in parallel per class, i think it is prefer to create more class than putting multiple test within same class.

JobsController is under "Kudu.Services" that is why i create similar folder structure. 
i can name this new class start with JobsController, to make it more easier to related to JobsController.cs

Kudu.Core/Infrastructure/LockFile.cs
shrimpy: we have a test file for FileSystemHelpers.

i was thinking other test case already cover this code path, so do not want to duplicate the effort. but sure that i can add one test case to just test this function.

Kudu.Services.Test/Jobs/InvokeTriggeredJobReadonlyFileSystemTests.cs
shrimpy: the value for the tmp folder doesn`t really matter, as long as it is not default value. i will add comment

Kudu.Services.Test/Jobs/InvokeTriggeredJobReadonlyFileSystemTests.cs
shrimpy: sure, will add a test case for happy path

Kudu.Services.Test/Jobs/InvokeTriggeredJobReadonlyFileSystemTests.cs
shrimpy: discuss with Mathew offline, i will rename the file to be "JobsControllerTests"

Kudu.Services.Test/Jobs/JobsControllerTests.cs
```diff
@@ -0,0 +1,63 @@
...
+            fileSystem.Setup(f => f.Directory).Returns(dirBase.Object);
+            dirBase.Setup(d => d.CreateDirectory(It.IsAny<string>())).Throws<UnauthorizedAccessException>();
+            FileSystemHelpers.Instance = fileSystem.Object;
+            FileSystemHelpers.TmpFolder = @"D:\";   // value doesn`t really matter, just need to have something other than default value
+
+            controller.Request = new HttpRequestMessage();
+
+            HttpResponseMessage resMsg = controller.InvokeTriggeredJob("foo");
+            Assert.Equal(HttpStatusCode.ServiceUnavailable, resMsg.StatusCode);
+
```
mathewc: nit: extra spaces

Kudu.Services.Test/Jobs/JobsControllerTests.cs
```diff
@@ -0,0 +1,63 @@
...
+            var controller = new JobsController(
+                Mock.Of<ITriggeredJobsManager>(),
+                Mock.Of<IContinuousJobsManager>(),
+                Mock.Of<ITracer>());
+
+            controller.Request = new HttpRequestMessage();
+
+            HttpResponseMessage resMsg = controller.InvokeTriggeredJob("foo");
+            Assert.Equal(HttpStatusCode.Accepted, resMsg.StatusCode);
+
```
mathewc: nit: extra spaces

Kudu.Services.Test/Jobs/JobsControllerTests.cs
```diff
@@ -0,0 +1,63 @@
...
+            FileSystemHelpers.Instance = fileSystem.Object;
+            FileSystemHelpers.TmpFolder = @"D:\";   // value doesn`t really matter, just need to have something other than default value
+
+            controller.Request = new HttpRequestMessage();
+
+            HttpResponseMessage resMsg = controller.InvokeTriggeredJob("foo");
+            Assert.Equal(HttpStatusCode.ServiceUnavailable, resMsg.StatusCode);
+
+        }
+
```
mathewc: nit: extra spaces

Kudu.Services.Test/Jobs/JobsControllerTests.cs
```diff
@@ -0,0 +1,63 @@
...
+            controller.Request = new HttpRequestMessage();
+
+            HttpResponseMessage resMsg = controller.InvokeTriggeredJob("foo");
+            Assert.Equal(HttpStatusCode.Accepted, resMsg.StatusCode);
+
+        }
+
+        [Fact]
+        public void InvokeTriggeredJobShouldReturn503()
```
mathewc: InvokeTriggeredJob_ReadOnlyFileSystem_ShouldReturn503

Kudu.Services.Test/Jobs/JobsControllerTests.cs
shrimpy: ah.. thanks! fixed.

Kudu.Services.Test/Jobs/JobsControllerTests.cs
shrimpy: fixed

mathewc:
Couple minor comments, otherwise looks good to me :)

Kudu.Services.Test/Jobs/JobsControllerTests.cs
shrimpy: fixed

Kudu.Services.Test/Jobs/JobsControllerTests.cs
shrimpy: sure. updated

davidebbo:
:shipit:

PR #1637, opened 2015-07-26
Fixed site extensions don't overwrite files correctly

shibayan:
If the file size after the update is small, it is not performed correctly file overwriting.

### DIFF
--- a/Kudu.Core/SiteExtensions/FeedExtensions.cs
+++ b/Kudu.Core/SiteExtensions/FeedExtensions.cs
@@ -266,10 +266,13 @@ namespace Kudu.Core.SiteExtensions
             FileSystemHelpers.EnsureDirectory(packageFolderPath);
             using (Stream writeStream = FileSystemHelpers.OpenWrite(filePath))
             {
-                // reset length of file stream
-                writeStream.SetLength(0);
+                OperationManager.Attempt(() =>
+                {
+                    // reset length of file stream
+                    writeStream.SetLength(0);
 
-                OperationManager.Attempt(() => stream.CopyTo(writeStream));
+                    stream.CopyTo(writeStream);
+                });
             }
         }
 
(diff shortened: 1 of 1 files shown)


Kudu.Core/SiteExtensions/FeedExtensions.cs
```diff
@@ -103,6 +103,9 @@ public static class FeedExtensions
                         FileSystemHelpers.EnsureDirectory(Path.GetDirectoryName(fullPath));
                         using (Stream writeStream = FileSystemHelpers.OpenWrite(fullPath))
                         {
+                            // reset length of file stream
+                            writeStream.SetLength(0);
```
shrimpy: why [line 107](https://github.com/projectkudu/kudu/pull/1637/files#diff-aeb25288ddaac13782bf8b797f7a288cR107)  and [line 192](https://github.com/projectkudu/kudu/pull/1637/files#diff-aeb25288ddaac13782bf8b797f7a288cR192) are necessary?

Kudu.Core/SiteExtensions/FeedExtensions.cs
```diff
@@ -260,6 +266,9 @@ private static void WriteStreamToFile(Stream stream, string filePath)
             FileSystemHelpers.EnsureDirectory(packageFolderPath);
             using (Stream writeStream = FileSystemHelpers.OpenWrite(filePath))
             {
+                // reset length of file stream
+                writeStream.SetLength(0);
```
shrimpy: i think line 270 should be moved to right before "stream.CopyTo" where retry actually happen

shibayan:
Writing a small data to stream opened by OpenWrite, because that corresponds to the problem the original data remains.

sample code
```csharp
var reader = new MemoryStream(Encoding.UTF8.GetBytes("aaa"));
var writer = new MemoryStream(Encoding.UTF8.GetBytes("foobarbaz"));

//writer.SetLength(0);
reader.CopyToAsync(writer).Wait();

// comment out the SetLength(0) : aaabarbaz / don't comment out SetLength(0) : aaa
Console.WriteLine(Encoding.UTF8.GetString(writer.ToArray()));
```

The same thing is happening in the installation of Site Extension

shrimpy:
@shibayan right, thanks for the fix. can you update your code regarding to my other comment. just in-case retry will write half of the content.

shibayan:
@shrimpy Thank you for feedback. I committed modify the processing of retry.

PR #1639, opened 2015-07-29
Fix Ace editor focus after load

snobu:
Ace editor now comes into focus after loading: `editor.focus()`. Thanks @davidebbo 
Also, the statusbar padding was 1px off.


davidebbo:
Looks good, thanks!

Issue #1566, opened 2015-05-24
Update error message regarding low CPU activity to mention WEBJOBS_IDLE_TIMEOUT instead of SCM_COMMAND_IDLE_TIMEOUT 


miparnisari:
We have a **triggered webjob** (disk intensive) set up in one website. The job sometimes failed due to a "timeout" error:

`Command ... aborted due to no output and CPU activity for 121 seconds.  You may increase SCM_COMMAND_IDLE_TIMEOUT setting to solve the issue.`

We increased the value of `SCM_COMMAND_IDLE_TIMEOUT` from 60 to 9999. 

Did we set the correct parameter, or did we have to set WEBJOBS_IDLE_TIMEOUT?

amitapl:
`WEBJOBS_IDLE_TIMEOUT` is the correct one to update, we'll need to fix the error message.

But updating `SCM_COMMAND_IDLE_TIMEOUT` shouldn't have caused any issues, can you share what random errors you see?

miparnisari:
Oh, never mind, the error was totally unrelated :)

HansOlavS:
I've created a PR with a fix for this in #1640.

PR #1640, opened 2015-07-30
Fix for issue #1566

HansOlavS:
Changed method CreateIdleTimeoutException to suggest correct setting 'WEBJOBS_IDLE_TIMEOUT' for increasing the idle timeout.

### DIFF
--- a/Kudu.Core/Infrastructure/IdleManager.cs
+++ b/Kudu.Core/Infrastructure/IdleManager.cs
@@ -135,7 +135,7 @@ namespace Kudu.Core.Infrastructure
             string message = String.Format(Resources.Error_ProcessAborted,
                                 process.Name + " " + arguments, 
                                 totalWaitDuration.TotalSeconds, 
-                                SettingsKeys.CommandIdleTimeout);
+                                SettingsKeys.WebJobsIdleTimeoutInSeconds);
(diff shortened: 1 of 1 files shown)


davidebbo:
It's not that simple. This is general code that also gets called outside of WebJobs, and in most cases `SCM_COMMAND_IDLE_TIMEOUT` is correct. e.g. during a long git deployment waiting for some operation.

Making the message correct would probably involve passing in the flag name, or something like that. We'd need to look at all the call trees leading to it to see what makes sense.

HansOlavS:
@davidebbo, what about a new method spesifically used from webjobs? That way I won't break anything dependent on the existing method used...

mathewc:
I don't think we should introduce a new method. How about simply updating the message so that it reads like "You may increase SCM_COMMAND_IDLE_TIMEOUT (or WEBJOBS_IDLE_TIMEOUT if this is a web job) to solve the issue". This way we don't need to propagate a flag. I'm not opposed to the flag, if there is a clean way to do it.

davidebbo:
Definitely we don't want to duplicate the entire logic here. I think @mathewc suggestion is the cheapest and safest, even if it is not the 'purest'. :) Let's go with that.

HansOlavS:
Excellent! I thought about that path too, but felt it was a bit too simple (not pure).

PR #1641, opened 2015-07-31
kudu favicon ftw!

kenegozi:


davidebbo:
Good catch! Had to add it to csproj as well.

kenegozi:
lol half a year away from .NET and I totally forgot about csproj and not automatically including everything in the folder structure.

davidebbo:
I haven't been away at all, and I still messed it up! https://github.com/projectkudu/kudu/commit/394df9b423cba6c8c7cc659baa827ea33834feb7

davidebbo:
One thing I just noticed is that the favicon ends up also applying to the WebJobs dashboard (or any site extension), since it's hosted in the scm site. A bit quirky but no big deal. Is it even possible to override the favicon for a sub path?

PR #1643, opened 2015-08-03
Replacing XmlLogger with StructuredTextLogger for deployment logs to improve performance

### LINKED ISSUE #1622: Deployment logging slows the build tremendously
My project contains thousands of files to output (guess which project it is). The task output all these files in the console, and Kudu creates a new entry tag in its deployment log file for every new line.

```xml
        <entry time="2015-07-17T20:24:11.304537Z" id="1a9c5937-9e92-4135-a3bd-94dcaf2d959d" type="0">
            <message>  Copying file from "D:\home\site\repository\build\Compile\System.Web.WebPages.Razor.dll" to "D:\home\site\repository\build\Compile\_PublishedWebsites\Lucene\bin\System.Web.WebPages.Razor.dll".</message>
        </entry>
        <entry time="2015-07-17T20:24:11.5470539Z" id="7aabecc1-5270-4945-be6c-459f6bb85107" type="0">
            <message>  Copying file from "D:\home\site\repository\build\Compile\System.Web.Razor.dll" to "D:\home\site\repository\build\Compile\_PublishedWebsites\Lucene\bin\System.Web.Razor.dll".</message>
        </entry>
        <entry time="2015-07-17T20:24:11.8595585Z" id="9f135636-24b9-44e7-814a-af5872fd03a6" type="0">
            <message>  Copying file from "D:\home\site\repository\build\Compile\Owin.dll" to "D:\home\site\repository\build\Compile\_PublishedWebsites\Lucene\bin\Owin.dll".</message>
        </entry>
```

This makes the build process much slower than usual, in my case around 1h30, because I assume the file is flushed on the disk every seconds or so, and when the file reaches 1MB then most of the time is spent writing on disk.

I would suggest having an option to use a simple append only text file at least as the XML structure makes it very slow in this case.

ahmelsayed:
Fixes #1622
With the following `deploy.cmd`

``` batch
@ECHO OFF
REM this prints 5,000 lines
FOR /L %%i in (1,1,5000) do echo "Iteration %%i"
```

_| XmlLogger | StructuredTextLogger
--- | --- | ---
Time | 00:**10:26** | 00:00:**55**
File Name | `log.xml` | `log.log`
File Size | `768 KB` | `572 KB`
Parsing Time | ~`25 msec` for `XDocument.Load()` |~~`240 msec`~~ ~`120 msec`

Log format looks like

```
message="Updating branch 'master'." time="8/3/2015 9:43:09 PM" id="0884ef1a-d1f9-40a1-80af-43d27dbbb988" type="0"
message="Updating submodules." time="8/3/2015 9:43:09 PM" id="72566408-30a1-4e13-a15d-0734ab5c7e6e" type="0"
message="Preparing deployment for commit id '7fefd88f08'." time="8/3/2015 9:43:09 PM" id="c71756ba-729d-49f5-90cf-cf958a71ba24" type="0"
message="Running custom deployment command..." time="8/3/2015 9:43:10 PM" id="c4c96196-32eb-4b90-a82a-fecc161f8be5" type="0"
message="Running deployment command..." time="8/3/2015 9:43:10 PM" id="fbd0ebf5-2587-47e9-afd5-70db8242d163" type="0"
	message="Command: deploy.cmd" time="8/3/2015 9:43:10 PM" id="b0aff723-2a12-4df2-99b7-c0fdff5d360f" type="0"
	message="&quot;Iteration 1&quot;" time="8/3/2015 9:43:11 PM" id="1cb3dd5e-37b7-4652-b849-0d86bce1c83c" type="0"
	message="&quot;Iteration 2&quot;" time="8/3/2015 9:43:11 PM" id="577e0fdb-9070-4bc0-9093-c4532fb2c131" type="0"
	message="&quot;Iteration 3&quot;" time="8/3/2015 9:43:11 PM" id="4c7c460f-c111-4669-a73a-a73a4ca4b1d7" type="0"
	message="&quot;Iteration 4&quot;" time="8/3/2015 9:43:11 PM" id="aa17724c-e1df-459c-8e0d-b93a02bef249" type="0"
...
message="Deployment successful." time="8/3/2015 9:43:53 PM" id="234e947a-4d78-4c76-a53d-2b3d6e619c8a" type="0"
```
**Unlike** the `XmlLogger`, this logger doesn't support out-of-order logging.

**Update** with 0084c729ed3be8ec6428c58c4b29020128256c54 the current format is 
note the removal of logId `if depth > 0`

```
2015-08-04T22:08:27.2522415Z,Updating branch 'master'.,5da51d61-0b80-410e-b7a0-97e9b2f961a3,0
2015-08-04T22:08:27.5491152Z,Updating submodules.,4db90cb9-0ba1-4829-986c-81b5182a03c3,0
2015-08-04T22:08:27.6116162Z,Preparing deployment for commit id '1fd0c85025'.,360561a0-ff2c-4619-ae3a-2c72f8aeef54,0
2015-08-04T22:08:27.9553831Z,Generating deployment script.,7f2d1293-3095-42a1-83c0-288145845b67,0
	2015-08-04T22:08:28.2991456Z,Using the following command to generate deployment script: 'azure site deploymentscript -y --no-dot-deployment -r "D:\home\site\repository" -o "D:\home\site\deployments\tools" --aspNet5 "D:\home\site\repository\src\HelloWeb\project.json" --aspNet5Version "1.0.0-beta5" --aspNet5Runtime "CLR" --aspNet5Architecture "x86"'.,,0
	2015-08-04T22:08:29.1585510Z,Project file path: .\src\HelloWeb\project.json,,0
	2015-08-04T22:08:29.1741795Z,Generating deployment script for ASP.NET 5 Application,,0
2015-08-04T22:08:29.2835620Z,Running deployment command...,26f2008f-98a5-49dc-b9b9-9497073436f9,0
	2015-08-04T22:08:29.2991817Z,Command: "D:\home\site\deployments\tools\deploy.cmd",,0
	2015-08-04T22:08:30.4547876Z,Handling ASP.NET 5 Web Application deployment.,,0
...
2015-08-04T22:09:06.5465687Z,Deployment successful.,8037065c-0e94-4c2c-898b-20cf8155fc49,0
```

### DIFF
--- a/Kudu.Core/Deployment/DeploymentManager.cs
+++ b/Kudu.Core/Deployment/DeploymentManager.cs
@@ -33,7 +33,8 @@ namespace Kudu.Core.Deployment
         private readonly IDeploymentStatusManager _status;
         private readonly IWebHooksManager _hooksManager;
 
-        private const string LogFile = "log.xml";
+        private const string XmlLogFile = "log.xml";
+        private const string TextLogFile = "log.log";
         private const string TemporaryDeploymentIdPrefix = "temp-";
         public const int MaxSuccessDeploymentResults = 10;
 
@@ -94,7 +95,7 @@ namespace Kudu.Core.Deployment
 
                 VerifyDeployment(id, IsDeploying);
 
-                var logger = new XmlLogger(path, _analytics);
+                var logger = GetLoggerForFile(path);
                 List<LogEntry> entries = logger.GetLogEntries().ToList();
 
                 // Determine if there's details to show at all
@@ -121,7 +122,7 @@ namespace Kudu.Core.Deployment
 
                 VerifyDeployment(id, IsDeploying);
 
-                var logger = new XmlLogger(path, _analytics);
+                var logger = GetLoggerForFile(path);
 
                 return logger.GetLogEntryDetails(entryId).ToList();
             }
@@ -792,8 +793,8 @@ namespace Kudu.Core.Deployment
         public ILogger GetLogger(string id)
         {
             var path = GetLogPath(id);
-            var xmlLogger = new XmlLogger(path, _analytics);
-            return new ProgressLogger(id, _status, new CascadeLogger(xmlLogger, _globalLogger));
+            var logger = GetLoggerForFile(path);
+            return new ProgressLogger(id, _status, new CascadeLogger(logger, _globalLogger));
         }
 
         /// <summary>
@@ -834,9 +835,21 @@ namespace Kudu.Core.Deployment
             return Path.Combine(GetRoot(id), Constants.ManifestFileName);
         }
 
+        /// <summary>
+        /// This function handles getting the path for the log file.
+        /// If 'log.xml' exists then use that which will use XmlLogger, this is to support existing log files
+        /// else use 'log.log' which will use TextLogger. Moving forward deployment will always use text logger.
+        /// The logic to get the right logger is in GetLoggerForFile()
+        /// </summary>
+        /// <param name="id">deploymentId which is part of the path for the log file</param>
+        /// <param name="ensureDirectory">Create the directory if it doesn't exist</param>
+        /// <returns>log file path</returns>
         private string GetLogPath(string id, bool ensureDirectory = true)
         {
-            return Path.Combine(GetRoot(id, ensureDirectory), LogFile);
+            var logPath = Path.Combine(GetRoot(id, ensureDirectory), XmlLogFile);
+            return FileSystemHelpers.FileExists(logPath)
+                ? logPath
+                : Path.Combine(GetRoot(id, ensureDirectory), TextLogFile);
         }
 
         private string GetRoot(string id, bool ensureDirectory = true)
@@ -854,6 +867,24 @@ namespace Kudu.Core.Deployment
         private bool IsActive(string id)
         {
             return id.Equals(_status.ActiveDeploymentId, StringComparison.OrdinalIgnoreCase);
+        }
+
+        /// <summary>
+        /// use XmlLogger if the file ends with .xml
+        /// Otherwise use StructuredTextLogger
+        /// </summary>
+        /// <param name="logPath"></param>
+        /// <returns>XmlLogger or StructuredTextLogger</returns>
+        private IDetailedLogger GetLoggerForFile(string logPath)
+        {
+            if (logPath.EndsWith(".xml", StringComparison.OrdinalIgnoreCase))
+            {
+                return new XmlLogger(logPath, _analytics);
+            }
+            else
+            {
+                return new StructuredTextLogger(logPath, _analytics);
+            }
         }
 
         private class DeploymentAnalytics : IDisposable
--- a/Kudu.Core/Deployment/StructuredTextDocument.cs
+++ b/Kudu.Core/Deployment/StructuredTextDocument.cs
@@ -1,0 +1,129 @@
+﻿using Kudu.Core.Infrastructure;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Kudu.Core.Deployment
+{
+    public class StructuredTextDocument<T>
+    {
+        private const string LogTemplate = "{0}{1}";
+        private readonly string _path;
+        private readonly Func<T, string> _serializer;
+        private readonly Func<string, T> _deserializer;
+
+        public StructuredTextDocument(string path, Func<T, string> serializer, Func<string, T> deserializer)
+        {
+            _path = path;
+            _serializer = serializer;
+            _deserializer = deserializer;
+        }
+
+        public IEnumerable<StructuredTextDocumentEntry<T>> GetDocument()
+        {
+            return ParseLines(FileSystemHelpers.ReadAllLines(_path), startIndex: 0, depth: 0);
+        }
+
+        /// <summary>
+        /// Write uses the serializer that was specified in the constructor to convert T into a string
+        /// then logs it with the correct depth.
+        /// The depth determines how many \t are there in front of the log message. Which implies the
+        /// Parent/Child relationship between the log entries.
+        /// Here is an example of how the log looks like with 1 level of depth
+        /// 0000-00-00T00:00:00.0000000Z,Message 1,00000000-0000-0000-0000-000000000000,0
+        /// 0000-00-00T00:00:00.0000000Z,Message 2,00000000-0000-0000-0000-000000000000,0
+        /// \t0000-00-00T00:00:00.0000000Z,Message 2.1,,0
+        /// \t0000-00-00T00:00:00.0000000Z,Message 2.2,,0
+        /// 0000-00-00T00:00:00.0000000Z,Message 3,00000000-0000-0000-0000-000000000000,0
+        /// </summary>
+        /// <param name="value">Object to be serialized and logged in the StructuredTextDocument</param>
+        /// <param name="depth">Determines the Parent/Child relationship between log entries. depth = 0 is parent for depth = 1</param>
+        public void Write(T value, int depth)
+        {
+            var stringValue = _serializer(value);
+            ValidateStringMessage(stringValue);
+
+            var stringDepth = string.Concat(Enumerable.Range(0, depth).Select(_ => StructuredTextDocument.LeadingCharacter));
+            var logEntry = string.Format(LogTemplate, stringDepth, stringValue);
+            Log(logEntry);
+        }
+
+        private IEnumerable<StructuredTextDocumentEntry<T>> ParseLines(string[] lines, int startIndex, int depth)
+        {
+            List<StructuredTextDocumentEntry<T>> collection = new List<StructuredTextDocumentEntry<T>>();
+            StructuredTextDocumentEntry<T> logEntry = null;
+            for (var i = startIndex; i < lines.Length; i++)
+            {
+                //skip empty lines
+                if (string.IsNullOrWhiteSpace(lines[i]))
+                    continue;
+
+                var entryDepth = StructuredTextDocument.GetEntryDepth(lines[i]);
+                if (entryDepth == depth)
+                {
+                    logEntry = new StructuredTextDocumentEntry<T>
+                    {
+                        LogEntry = _deserializer(lines[i].Substring(depth)),
+                        Children = Enumerable.Empty<StructuredTextDocumentEntry<T>>()
+                    };
+                    collection.Add(logEntry);
+                }
+                else if (entryDepth > depth)
+                {
+                    if (logEntry == null)
+                    {
+                        throw new FormatException(string.Format("log file '{0}' is in an invalid format", _path));
+                    }
+
+                    logEntry.Children = ParseLines(lines, i, entryDepth);
+                    i += (logEntry.Children.Count() - 1);
+                }
+                else if (entryDepth < depth)
+                {
+                    return collection;
+                }
+            }
+            return collection;
+        }
+
+        private void Log(string logEntry)
+        {
+            FileSystemHelpers.AppendAllTextToFile(_path, string.Concat(logEntry, System.Environment.NewLine));
+        }
+
+        private static void ValidateStringMessage(string message)
+        {
+            if (message.StartsWith(StructuredTextDocument.LeadingCharacter, StringComparison.OrdinalIgnoreCase))
+            {
+                throw new FormatException(
+                    string.Format("Serialized log of type '{0}' => '{1}' can't start with character '{2}'",
+                    typeof(T).Name,
+                    message,
+                    StructuredTextDocument.PrintableLeadingCharacter));
+            }
+
+            foreach (var pair in StructuredTextDocument.NotAllowedSequances)
+            {
+                if (message.IndexOf(pair.Key, StringComparison.OrdinalIgnoreCase) != -1)
+                {
+                    throw new FormatException(
+                        string.Format("Serialized log of type '{0}' => '{1}' contains an invalid sequance '{2}' for a logEntry in a StructuredTextDocument",
+                        typeof(T).Name,
+                        message,
+                        pair.Value));
+                }
+            }
+        }
+    }
+
+    public static class StructuredTextDocument
+    {
+        public const string LeadingCharacter = "\t";
+        public const string PrintableLeadingCharacter = "\\t";
+        public static readonly IDictionary<string, string> NotAllowedSequances = new Dictionary<string, string> {  { "\r\n", "\\r\\n" }, { "\r", "\\r" }, { "\n", "\\n" }};
+        public static int GetEntryDepth(string entry)
+        {
+            return entry.TakeWhile(c => LeadingCharacter.Equals(c.ToString())).Count();
+        }
+    }
+}
--- a/Kudu.Core/Deployment/StructuredTextLogger.cs
+++ b/Kudu.Core/Deployment/StructuredTextLogger.cs
@@ -1,0 +1,127 @@
+﻿using Kudu.Core.Tracing;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Kudu.Core.Deployment
+{
+    public class StructuredTextLogger : IDetailedLogger
+    {
+        private readonly static object DocumentLock = new object();
+        private readonly IAnalytics _analytics;
+        private int _depth;
+        private readonly string _path;
+        private const string LogEntrySeparator = ",";
+        private readonly StructuredTextDocument<LogEntry> _structuredTextDocument;
+
+        private static IEnumerable<KeyValuePair<string, string>> EscapeChars = new Dictionary<string, string>
+        {
+            { ",", "&comma;" }
+        }
+        .Union(StructuredTextDocument.NotAllowedSequances);
+
+        public StructuredTextLogger(string path, IAnalytics analytics)
+        {
+            _depth = 0;
+            _path = path;
+            _analytics = analytics;
+            _structuredTextDocument = new StructuredTextDocument<LogEntry>(path,
+                // DateTime.ToString("o") => "2015-08-04T00:08:38.5489308Z"
+                e => string.Join(LogEntrySeparator, e.LogTime.ToString("o"), e.Message, e.Id, (int)e.Type),
+                str =>
+                {
+                    var splitted = str.Split(new[] { LogEntrySeparator }, StringSplitOptions.None);
+                    if (splitted.Length == 4)
+                    {
+                        var time = DateTime.Parse(splitted[0]).ToUniversalTime();
+                        var message = UnsanitizeValue(splitted[1]);
+                        var id = splitted[2];
+                        var type = (LogEntryType)Int32.Parse(splitted[3]);
+                        return new LogEntry(time, id, message, type);
+                    }
+                    else
+                    {
+                        throw new FormatException(string.Format("the log line \"{0}\" is in an invalid format", str));
+                    }
+                });
+        }
+
+        public ILogger Log(string value, LogEntryType type)
+        {
+            try
+            {
+                value = SanitizeValue(value);
+                lock(DocumentLock)
+                {
+                    var logId = _depth == 0 ? Guid.NewGuid().ToString() : string.Empty;
+                    _structuredTextDocument.Write(new LogEntry(DateTime.UtcNow, logId, value, type), _depth);
+                }
+
+                return new StructuredTextLogger(_path, _analytics)
+                {
+                    _depth = _depth + 1
+                };
+            }
+            catch(Exception e)
+            {
+                _analytics.UnexpectedException(e);
+                throw;
+            }
+        }
+
+        public IEnumerable<LogEntry> GetLogEntries()
+        {
+            try
+            {
+                return _structuredTextDocument.GetDocument().Select(e => e.LogEntry);
+            }
+            catch(Exception e)
+            {
+                _analytics.UnexpectedException(e);
+                throw;
+            }
+        }
+
+        public IEnumerable<LogEntry> GetLogEntryDetails(string entryId)
+        {
+            try
+            {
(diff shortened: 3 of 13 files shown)


davidebbo:
Haven't looked at the code, but trying to run it I'm seeing issues. Using the deployments API, trying to get the details of "Running deployment command...", just returns:

```js
[
  {
    log_time: "2015-08-03T23:59:45",
    id: "f5572ee8-c41c-480f-bb36-326361f91283",
    message: "Running deployment command...",
    type: 0,
    details_url: null
  }
]
```

Instead of returning entries for each output line.

Other question: how did you end up getting a bigger file size for the new logger (768k vs 572k), given the more concise syntax?

Kudu.Core/Deployment/StructuredTextDocument.cs
```diff
@@ -0,0 +1,119 @@
...
+            _serializer = serializer;
+            _deserializer = deserializer;
+        }
+
+        public IEnumerable<StructuredTextDocumentEntry<T>> GetDocument()
+        {
+            return ParseLines(FileSystemHelpers.ReadAllLines(_path), startIndex: 0, depth: 0);
+        }
+
+        public void Write(T value, int depth)
```
shrimpy: is the "Write" function a public API method? what does "depth" mean?

will be good to have more comment and class document to describe how it work, and have some sample output in the class comment would be a great help when reading the code.

Kudu.Core/Deployment/StructuredTextLogger.cs
```diff
@@ -0,0 +1,117 @@
...
+{
+    public class StructuredTextLogger : IDetailedLogger
+    {
+        private readonly IAnalytics _analytics;
+        private readonly int _depth;
+        private readonly string _path;
+        private const string LogEntryTemplate = "message=\"{0}\" time=\"{1}\" id=\"{2}\" type=\"{3}\"";
+        private const string LogEntryPattern = "message=\"(.*?)\"\\s+time=\"(.*?)\"\\s+id=\"(.*?)\"\\s+type=\"(.*?)\"";
+        private readonly Regex LogEntryRegex = new Regex(LogEntryPattern, RegexOptions.IgnoreCase | RegexOptions.Singleline);
+        private readonly StructuredTextDocument<LogEntry> _structuredTextDocment;
```
davidebbo: 'docment' typo

Kudu.Core/Deployment/StructuredTextLogger.cs
```diff
@@ -0,0 +1,117 @@
...
+                        throw new FormatException(string.Format("the log line \"{0}\" is in an invalid format", str));
+                    }
+                });
+        }
+
+        public ILogger Log(string value, LogEntryType type)
+        {
+            try
+            {
+                value = SanatizeValue(value);
```
davidebbo: Correct spelling is 'sanitize'. :) Same in a few other tokens.

Kudu.Core/Deployment/StructuredTextLogger.cs
```diff
@@ -0,0 +1,117 @@
...
+                _analytics.UnexpectedException(e);
+                throw;
+            }
+        }
+
+        public IEnumerable<LogEntry> GetLogEntryDetails(string entryId)
+        {
+            try
+            {
+                return _structuredTextDocment.GetDocument().Where(s => s.LogEntry.Id.Equals(entryId, StringComparison.OrdinalIgnoreCase)).Select(e => e.LogEntry);
```
davidebbo: Probably need to get the `Children` here.

ahmelsayed:
@davidebbo  regarding not getting the details, 16853d3d744175148bb9b06e5f6c7378e5fb3e05 should fix it. I was returning the node instead of the list of children.

ahmelsayed:
The file size was my mistake, I just swapped them in the table. The xml file is larger of course

davidebbo:
As for the parsed time being slower, that's probably ok as long as we no longer need to parse the whole file each time we add an entry, unlike with xml. Or maybe these were reversed too?

Kudu.Core/Deployment/StructuredTextLogger.cs
ahmelsayed: thanks :) I kept saying I'll check the spelling for it, but I forgot. Fixed in 7ff9acfee33dfdce4c3b797c25456104197992f0

Kudu.Core/Deployment/StructuredTextDocument.cs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    public static class StructuredTextDocument
+    {
+        public const string LeadingCharacter = "\t";
+        public const string PrintableLeadingCharacter = "\\t";
+        public static readonly IDictionary<string, string> NotAllowedSequances = new Dictionary<string, string> {  { "\r\n", "\\r\\n" }, { "\r", "\\r" }, { "\n", "\\n" }};
+        public static int GetEntryDepth(string entry)
+        {
+            return entry.TakeWhile(c => LeadingCharacter.Equals(c.ToString(), StringComparison.OrdinalIgnoreCase)).Count();
```
shrimpy: "StringComparison.OrdinalIgnoreCase" seems unnecessary, i don`t think "\t" should equal to "\T"

ahmelsayed:
No, these weren't reversed. I'm not sure how much of the xml parsing actually happens on `XDocument.Load` vs lazily later, but for `StructuredTextLogger` this is the full parsing time for a 5,000 line file.
However, yes we only parse the text log file when the deployment APIs are hit. Otherwise we just append the current log entry to the file, which is very fast.

Kudu.Core/Deployment/StructuredTextLogger.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        public IEnumerable<LogEntry> GetLogEntryDetails(string entryId)
+        {
+            try
+            {
+                return _structuredTextDocment.GetDocument()
+                    .Where(s => s.LogEntry.Id.Equals(entryId, StringComparison.OrdinalIgnoreCase))
+                    .Select(e => e.Children.Select(c => c.LogEntry))
+                    .SelectMany(i => i);
```
davidebbo: Seems a bit more generic than needed, since you can never have a case where multiple top level entries share the same id. Technically, it's more correct to find the one and only entry, and then just return its children.

Kudu.Core/Deployment/StructuredTextDocument.cs
ahmelsayed: I think FxCop was throwing an error for IndexOf so I added that for all. I'll change it

Kudu.Core/Deployment/StructuredTextLogger.cs
```diff
@@ -0,0 +1,120 @@
...
+        private readonly StructuredTextDocument<LogEntry> _structuredTextDocment;
+
+        private static IEnumerable<KeyValuePair<string, string>> EscapeChars = new Dictionary<string, string>
+        {
+            { "\"", "&quot;" }
+        }
+        .Union(StructuredTextDocument.NotAllowedSequances);
+
+        public StructuredTextLogger(string path, IAnalytics analytics, int depth = 0)
+        {
```
shrimpy: parameter "depth" should be internal?

when using StructuredTextLogger, i believe it is always start with "depth = 0", no?

Kudu.Core/Deployment/StructuredTextLogger.cs
```diff
@@ -0,0 +1,120 @@
...
+    {
+        private readonly IAnalytics _analytics;
+        private readonly int _depth;
+        private readonly string _path;
+        private const string LogEntryTemplate = "message=\"{0}\" time=\"{1}\" id=\"{2}\" type=\"{3}\"";
+        private const string LogEntryPattern = "message=\"(.*?)\"\\s+time=\"(.*?)\"\\s+id=\"(.*?)\"\\s+type=\"(.*?)\"";
+        private readonly Regex LogEntryRegex = new Regex(LogEntryPattern, RegexOptions.IgnoreCase | RegexOptions.Singleline);
+        private readonly StructuredTextDocument<LogEntry> _structuredTextDocment;
+
+        private static IEnumerable<KeyValuePair<string, string>> EscapeChars = new Dictionary<string, string>
```
shrimpy: Shall we also define "EscapeChars" as a Dictionary? so that when we find char to escape/replace, we can do containsKey instead of loop thru the collection one by one?

Kudu.Core/Deployment/StructuredTextLogger.cs
```diff
@@ -0,0 +1,120 @@
...
+                    }
+                });
+        }
+
+        public ILogger Log(string value, LogEntryType type)
+        {
+            try
+            {
+                value = SanatizeValue(value);
+                _structuredTextDocment.Write(new LogEntry(DateTime.UtcNow, Guid.NewGuid().ToString(), value, type), _depth);
```
davidebbo: I'm seeing the datetime format being different from before, e.g.

- xml log: time="2015-08-04T00:08:38.5489308Z"
- new log: time="8/4/2015 12:55:26 AM"

Kudu.Core/Deployment/StructuredTextLogger.cs
shrimpy: ah.. after reading "SanatizeValue" and "UnsanatizeValue", i take my comment back. you will need to go thru each of them one by one.

Kudu.Core/Deployment/StructuredTextLogger.cs
ahmelsayed: Good point. Does c8e8a951792260d27d995d3b194c3ab25d74fc0d look right?

Kudu.Core/Deployment/StructuredTextDocument.cs
```diff
@@ -0,0 +1,119 @@
+﻿using Kudu.Core.Infrastructure;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Kudu.Core.Deployment
+{
+    public class StructuredTextDocument<T>
+    {
+        private readonly static object DocumentLock = new object();
```
shrimpy: "DocumentLock" should per log file? shouldn`t be static? 

right now all logs are share the same lock object

Kudu.Core/Deployment/StructuredTextDocument.cs
```diff
@@ -0,0 +1,119 @@
...
+                    {
+                        throw new FormatException(string.Format("log file '{0}' is in an invalid format", _path));
+                    }
+
+                    logEntry.Children = ParseLines(lines, i, entryDepth);
+                    i += (logEntry.Children.Count() - 1);
+                }
+                else if (entryDepth < depth)
+                {
+                    return collection;
```
shrimpy: maybe change to 

````
else 
{
    break;
}
````

so that we only have one return statement

Comment (davidebbo) [inline Kudu.Core/Deployment/StructuredTextDocument.cs]: Personally, that doesn't bother me :)

File: Kudu.Core/Deployment/StructuredTextLogger.cs (hunk 2 above)
Comment (ahmelsayed) [inline Kudu.Core/Deployment/StructuredTextLogger.cs]: Fixed.

**Comment by ahmelsayed** [comment]
using `string.Split()` instead of the regex matching cuts the parsing time in half. I changed it in 503c847d45496a581edd8f6989895f4b2b5fc996

File: Kudu.Core/Deployment/StructuredTextDocument.cs (hunk 10 above)
Comment (ahmelsayed) [inline Kudu.Core/Deployment/StructuredTextDocument.cs]: I think it should be a static lock in the logger and not the document.

File: Kudu.Core/Deployment/StructuredTextLogger.cs (hunk 9 above)
Comment (ahmelsayed) [inline Kudu.Core/Deployment/StructuredTextLogger.cs]: it seems that [XDocument pretty much re-implements how DateTime is serialized ](http://referencesource.microsoft.com/#System.Xml/System/Xml/Schema/XsdDateTime.cs,505) Though I think `.ToString("o")` has the same output we want.

**Comment by davidebbo** [comment]
So here is a thought: why not have each line be more of an ordered sequence without keys? e.g. csv or iis log style. e.g.

```
message="Running deployment command..." time="8/3/2015 9:43:10 PM" id="fbd0ebf5-2587-47e9-afd5-70db8242d163" type="0"
    message="Command: deploy.cmd" time="8/3/2015 9:43:10 PM" id="b0aff723-2a12-4df2-99b7-c0fdff5d360f" type="0"
```

becomes something like:

```
8/3/2015 9:43:10 PM,Running deployment command...,fbd0ebf5-2587-47e9-afd5-70db8242d163,0
    8/3/2015 9:43:10 PM,Command: deploy.cmd,b0aff723-2a12-4df2-99b7-c0fdff5d360f,0
```

File: Kudu.Core/Deployment/StructuredTextLogger.cs [hunk 12]
```diff
@@ -0,0 +1,131 @@
+﻿using Kudu.Core.Tracing;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Kudu.Core.Deployment
+{
+    public class StructuredTextLogger : IDetailedLogger
+    {
+        private readonly static object DocumentLock = new object();
```
shrimpy: so entire application life cycle we only have one instance of "StructuredTextLogger" ?

Kudu.Core/Deployment/StructuredTextLogger.cs
```diff
@@ -0,0 +1,131 @@
...
+                if (value.Contains(pair.Value))
+                {
+                    value = value.Replace(pair.Value, pair.Key);
+                }
+            }
+
+            return value;
+        }
+
+        private class InnerStructuredTextLogger : StructuredTextLogger
```
shrimpy: why we need an extra class here? why not just use StructuredTextLogger?

Kudu.Core/Deployment/StructuredTextLogger.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        public ILogger Log(string value, LogEntryType type)
+        {
+            try
+            {
+                value = SanitizeValue(value);
+                lock(DocumentLock)
+                {
+                    _structuredTextDocument.Write(new LogEntry(DateTime.UtcNow, Guid.NewGuid().ToString(), value, type), _depth);
```
shrimpy: do you think use a auto increase integer as LogEntry ID would be a bit easier when reading the log? also keep the log message short.

Kudu.Core/Deployment/StructuredTextLogger.cs
ahmelsayed: no, but multiple instances of `StructuredTextLogger` write to the same file.

Kudu.Core/Deployment/StructuredTextLogger.cs
davidebbo: Actually, I'm not sure that the ID provides any value at all, since we don't have any routes that gets an individual item. Though we need to be careful not to break existing clients (like portal) in case they look for this ID in the API.

Kudu.Core/Deployment/StructuredTextLogger.cs
shrimpy: i see, so we only have one deployment log file per application.
so shall we rename this logger to specific to deployment? since it might not be ideal if we use this logger in other scenario that have multiple log files.

ahmelsayed: Not quite. Each deployment ends up with a log file. Though during the deployment each time you call `.Log("Step 1")` you get back another instance of the same logger that you can then use to call `.Log("Step 1.1")` which will be a nested message under `"Step 1"`. This is the same way `XmlLogger` works today. Does that make sense?

Kudu.Core/Deployment/StructuredTextLogger.cs
ahmelsayed: Actually I think [`{logId}` here is this ID](https://github.com/projectkudu/kudu/blob/master/Kudu.Services.Web/App_Start/NinjectServices.cs#L413) though it's only for top level logs, not for logs with `depth > 0`

Kudu.Core/Deployment/StructuredTextLogger.cs
shrimpy: i see. and we only have one deployment happen at a time, so the lock object won`t be share amount multiple files.

still, it won`t be ideal if we use it in scenario where we want to write message into multiple files at the same time. 

and this lock object seems specific to deployment log scenario. i would suggest either we rename the file to be more deployment specific or have a way to do lock per file for general case (maybe use a static Dictionary<file path, lock object> to store lock objects, but need to make sure we have another routine to clean invalid data from the Dictionary where log file haven`t been touched anymore or deleted).  

what do you think?

ahmelsayed:
with 0084c729ed3be8ec6428c58c4b29020128256c54 the current format is 
note the removal of logId `if depth > 0`

```
2015-08-04T22:08:27.2522415Z,Updating branch 'master'.,5da51d61-0b80-410e-b7a0-97e9b2f961a3,0
2015-08-04T22:08:27.5491152Z,Updating submodules.,4db90cb9-0ba1-4829-986c-81b5182a03c3,0
2015-08-04T22:08:27.6116162Z,Preparing deployment for commit id '1fd0c85025'.,360561a0-ff2c-4619-ae3a-2c72f8aeef54,0
2015-08-04T22:08:27.9553831Z,Generating deployment script.,7f2d1293-3095-42a1-83c0-288145845b67,0
	2015-08-04T22:08:28.2991456Z,Using the following command to generate deployment script: 'azure site deploymentscript -y --no-dot-deployment -r "D:\home\site\repository" -o "D:\home\site\deployments\tools" --aspNet5 "D:\home\site\repository\src\HelloWeb\project.json" --aspNet5Version "1.0.0-beta5" --aspNet5Runtime "CLR" --aspNet5Architecture "x86"'.,,0
	2015-08-04T22:08:29.1585510Z,Project file path: .\src\HelloWeb\project.json,,0
	2015-08-04T22:08:29.1741795Z,Generating deployment script for ASP.NET 5 Application,,0
2015-08-04T22:08:29.2835620Z,Running deployment command...,26f2008f-98a5-49dc-b9b9-9497073436f9,0
	2015-08-04T22:08:29.2991817Z,Command: "D:\home\site\deployments\tools\deploy.cmd",,0
	2015-08-04T22:08:30.4547876Z,Handling ASP.NET 5 Web Application deployment.,,0
...
2015-08-04T22:09:06.5465687Z,Deployment successful.,8037065c-0e94-4c2c-898b-20cf8155fc49,0
```

Kudu.Core/Deployment/StructuredTextLogger.cs
ahmelsayed: by rename the file, you mean the `StructuredTextLogger.cs` right? I mean it's in the `Kudu.Core.Deployment` namespace :)
I can change the name, I just think that having to manage a static collection for the locks might be too much for this logger at this point

Kudu.Core/Deployment/StructuredTextLogger.cs
ahmelsayed: Fixed

Kudu.Core/Deployment/StructuredTextLogger.cs
shrimpy: agree that might be too much.. and didn`t notice the namespace.. maybe put some class comments will be good enough, to avoid using this Logger in the wrong way.

Kudu.Core/Deployment/StructuredTextDocument.cs
ahmelsayed: Fixed in 598ea2c3b51f24215798570bc32efcb650b3c1be

shrimpy:
:shipit:

Kudu.Services.Web/_Layout.cshtml
```diff
@@ -36,7 +36,7 @@
                     <span class="icon-bar"></span>
                     <span class="icon-bar"></span>
                 </button>
-                <a class="navbar-brand" href="~/"><img src="/Content/Images/kudu.svg" height="22" width="22">Kudu</a>
+                <a class="navbar-brand" href="~/"><img src="/Content/Images/kudu.svg" height="22" width="22" style="margin-right: -3px">udu</a>
```
davidebbo: Wrong PR?

ahmelsayed: Yes! sorry the commit was already in my repo when I created the branch. I'll make sure to not merge it.

Kudu.Core/Deployment/DeploymentManager.cs
```diff
@@ -836,7 +837,10 @@ private string GetDeploymentManifestPath(string id)
 
         private string GetLogPath(string id, bool ensureDirectory = true)
         {
-            return Path.Combine(GetRoot(id, ensureDirectory), LogFile);
+            var logPath = Path.Combine(GetRoot(id, ensureDirectory), XmlLogFile);
+            return FileSystemHelpers.FileExists(logPath)
+                ? logPath
+                : Path.Combine(GetRoot(id, ensureDirectory), TextLogFile);
```
suwatch: Mind adding comments on what scenario either Xml or Text logger will be used.  Looking at codes checking File exist is not obvious enough.

davidebbo: We'll always use Text going forward. Xml is to support existing sites that already have the old format. But yes, comment would clarify.

ahmelsayed: I addedd the following comment

``` csharp
/// <summary>
/// This function handles getting the path for the log file.
/// If 'log.xml' exists then use that which will use XmlLogger, this is to support existing log files
/// else use 'log.log' which will use the new TextLogger. Moving forward deployment will always use text logger.
/// The logic to get the right logger is in GetLoggerForFile()
/// </summary>
/// <param name="id">deploymentId which is part of the path for the log file</param>
/// <param name="ensureDirectory">Create the directory if it doesn't exist</param>
/// <returns>log file path</returns>
```

suwatch: loud and clear, thank!

PR #1644, opened 2015-08-05
Ace performance improvements and more syntax rules

snobu:
* Removed logic for multiple ace instances from `knockout-ace.js`, removed the file altogether and only added the bare minimum into `ace-init.js` (i.e. we don't need to track multiple ace instances so why drag that around).

* Added highlight rules for `.xdt`, `.asp`, `.aspx`, `.cshtml`.

* I'm now a good guy to jQuery and only set the status bar to red on first change instead of _every_ change.

```javascript
editor.on('change', function () {
    //
    if (contentHasChanged) {
        return;
    }
    //
}
```

* Switched `.css()` for `.addClass()`, just because it's so much faster.

* Overall there's a significat performance improvement now when editing large files.
Tested on a modest machine, i'm breezing throgh a 3MB .xml file with syntax turned on and there's very little cursor lag.

I wish i could tell you exactly _which_ change did this :)


davidebbo:
Seems your pull request got corrupted with unrelated commits from Ahmed. My guess is that you rebased things incorrectly.

snobu:
Bad merge. Ignore this PR, will open a new one now.

PR #1647, opened 2015-08-06
Allow any characters in URL to support complex vfs file names

davidebbo:
I don't think those restrictions have any security value in Kudu, because:
- the site is always authenticated
- if someone can get through authentication, they already own the site regardless of these restrictions


davidebbo:
I added test coverage to the vfs test.

suwatch:
:shipit:

PR #1650, opened 2015-08-07
Don't replace \\r and \\n when reading the log file

ahmelsayed:
Originally the text logger would replace all `["\r", "\n", "\r\n"] => ["\\r", "\\n", "\\r\\n"]` when writing the file, then replace them back when reading the file. This makes strings like `"C:\\Program Files\\npm\\"` become `"C:\\Program Files\npm\\"` which is of course wrong (note the new line inserted in the middle of the path to replace `"\\n"`

### DIFF
--- a/Kudu.Core/Deployment/StructuredTextLogger.cs
+++ b/Kudu.Core/Deployment/StructuredTextLogger.cs
@@ -17,8 +17,7 @@ namespace Kudu.Core.Deployment
         private static IEnumerable<KeyValuePair<string, string>> EscapeChars = new Dictionary<string, string>
         {
             { ",", "&comma;" }
-        }
-        .Union(StructuredTextDocument.NotAllowedSequances);
+        };
 
         public StructuredTextLogger(string path, IAnalytics analytics)
         {
@@ -100,7 +99,7 @@ namespace Kudu.Core.Deployment
 
         private static string SanitizeValue(string value)
         {
-            foreach (var pair in EscapeChars)
+            foreach (var pair in EscapeChars.Union(StructuredTextDocument.NotAllowedSequences))
             {
                 if (value.Contains(pair.Key))
                 {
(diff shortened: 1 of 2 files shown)


shrimpy:
do you have a sample output? the change looks small and good, but will be good to see a sample output for what is fixed.

ahmelsayed:
The log file won't change. When we read the log file and return the `LogEntry[]` before this we would have returned 

``` json
{
    "log_time": "2015-08-07T20:39:51.9453898Z",
    "id": "",
    "message": "command=\"C:\\Program Files (x86)\nodejs\\0.12.7\node.exe\"",
    "type": 0,
    "details_url": null
}
```

and now it would return 

``` json
{
    "log_time": "2015-08-07T20:39:51.9453898Z",
    "id": "",
    "message": "command=\"C:\\Program Files (x86)\\nodejs\\0.12.7\node.exe\"",
    "type": 0,
    "details_url": null
}
```

the change is small but basically `\\node` vs `\node` since `\node` is really `[NewLine]ode`

ahmelsayed:
Oh, also if there was indeed a new line in the message being logged, we will not return that as a new line when reading the log. It'll be `\\r\\n` or `\\n`

shrimpy:
shouldn`t the path be 

````
C:\\Program Files (x86)\\nodejs\\0.12.7\\node.exe
````

seems like missing one "\" before "node.exe"

File: Kudu.Core/Deployment/StructuredTextLogger.cs [hunk 1]
```diff
@@ -100,7 +99,7 @@ public IEnumerable<LogEntry> GetLogEntryDetails(string entryId)
 
         private static string SanitizeValue(string value)
         {
-            foreach (var pair in EscapeChars)
+            foreach (var pair in EscapeChars.Union(StructuredTextDocument.NotAllowedSequances))
```
davidebbo: Spelling: 'sequ**e**nce' :)

ahmelsayed: fixed the typo

PR #1651, opened 2015-08-07
Set parent LogEntryType to Max of children entry types

ahmelsayed:


Kudu.Core/Deployment/StructuredTextLogger.cs
```diff
@@ -98,6 +107,25 @@ public IEnumerable<LogEntry> GetLogEntryDetails(string entryId)
             }
         }
 
+        private LogEntryType ChildTypeFix(IEnumerable<StructuredTextDocumentEntry<LogEntry>> tree)
```
suwatch: Could use comments.  Other than that looks great.

davidebbo: Right, the fact that the enum is ordered by severity is no clear enough

ahmelsayed: added this commend 

``` csharp
/// <summary>
/// StructuredTextDocument does pure File Append for each new message. This is not sufficient to replicate the behavior of XmlLogger
/// XmlLogger would set the Type of the parent LogEntry to be the Max of all the Types of the sub entries.
/// Since this logger only does append, we don't have a change to update the parent while writing the file, so will do it on reading here.
/// LogEntryType is an Enum with Message = 0, Warning = 1, and Error = 2.
/// When reading the tree of StructuredTextDocumentEntry, we need to fix the parent to equal the max of the all the children's Type.
/// </summary>
/// <param name="tree">A StructuredTextDocumentEntry tree to fix all the nodes in</param>
/// <returns>Max LogEntryType of the children</returns>
private LogEntryType ChildTypeFix(IEnumerable<StructuredTextDocumentEntry<LogEntry>> tree)
```

Issue #1648, opened 2015-08-07
TriggeredJobFinished hook is not documented anywhere


chuchuva:
Would be nice to have some documentation. Also, `https://<mysite>.scm.azurewebsites.net/WebHooks` does not mention TriggeredJobFinished event:

![image](https://cloud.githubusercontent.com/assets/237722/9127261/5a200e80-3cfb-11e5-8cd0-04e2910f1793.png)

davidebbo:
I think closest thing is @amitapl's post: http://blog.amitapple.com/post/2014/03/webjobs-notification/. But yes, we need some doc cleanup here.

As for the WebHooks UI, support would need to be added for this hook type. Any volunteer? :)

sitereactor:
I gave it a shot in this PR: https://github.com/projectkudu/kudu/pull/1652 :smile:

davidebbo:
Thanks @sitereactor change works great! This will be on Azure in our next Kudu deployment.

PR #1652, opened 2015-08-08
Adds the TriggeredJobFinished event to the WebHooks form

sitereactor:
Based on issue #1648 this PR adds a dropdown to the WebHooks form, which has both the `PostDeployment` and `TriggeredJobFinished`events available to choose when creating new webhooks.

This is tested locally, and seems to work fine. But had some issues running the build, as some tests (unrelated to this PR) were failing.


davidebbo:
Nice and simple, thanks!

PR #1662, opened 2015-08-18
Use network service identity for app pools to avoid strange issue

davidebbo:


davidebbo:
Here is a really long path I see in procmon:

`C:\Windows\ServiceProfiles\NetworkService\AppData\Local\Temp\04be9d04-a2e8-4436-9cba-fffa5616810c\approot\packages\Microsoft.Framework.Configuration.EnvironmentVariables\1.0.0-beta5\Microsoft.Framework.Configuration.EnvironmentVariables.1.0.0-beta5.nupkg`.

That's 254 chars, and maybe it needs to unzip it afterwards.

We need to make this path shorter somehow if we're going to use network service.

davidebbo:
Solved by using `DateTime.UtcNow.Ticks.ToString("x")` instead of guid.

PR #1664, opened 2015-08-19
Support DNX based WebJobs

ahmelsayed:


ahmelsayed:
FYI - changes in `dnvm.ps1` match changes here https://github.com/aspnet/dnvm/pull/396 so if they accept that PR, we won't need a special version of `dnvm.ps1`

Kudu.Core/Jobs/DnxScriptHost.cs
```diff
@@ -0,0 +1,28 @@
...
+        private static readonly string[] Supported = { "project.json" };
+
+        public DnxScriptHost()
+            : base("cmd", "/c runDnxWebJob.cmd")
+        {
+        }
+
+        public override IEnumerable<string> SupportedExtensions
+        {
+            get { return Enumerable.Empty<string>(); }
```
davidebbo: I'd move that to `ScriptHostBase` instead of having the method be abstract.

Kudu.Core/Jobs/JobsManagerBase.cs
```diff
@@ -495,6 +496,16 @@ private static string FindCommandToRun(FileInfoBase[] files, out IScriptHost scr
                         }
                     }
                 }
+
+                foreach (var supportedFileName in scriptHost.SupportedFileNames)
```
mathewc: Should the SupportedFileNames check be performed before supported extensions? For example, what happens if there are .js files in a Dnx directory root - would the Node ScriptHost be selected?

Kudu.Core/Jobs/DnxScriptHost.cs
```diff
@@ -0,0 +1,23 @@
...
+    class DnxScriptHost : ScriptHostBase
+    {
+        private static readonly string[] Supported = { "project.json" };
+
+        public DnxScriptHost()
+            : base("cmd", "/c runDnxWebJob.cmd")
+        {
+        }
+
+        public override IEnumerable<string> SupportedFileNames
```
mathewc: Now we have 2 paths - SupportedExtensions and SupportedFileNames. I wonder if a cleaner design would be to instead have an IScriptHost.GetEntryScript(FileInfoBase[] files) method that allows each host to control the scan logic itself, and return the entry script if found?

Kudu.Core/Jobs/JobsManagerBase.cs
davidebbo: That was an issue before, but with latest commit, SupportedFileNames would still end up winning over supported extensions. Yes, the logic is a bit tricky.

Kudu.Core/Jobs/DnxScriptHost.cs
davidebbo: The primary/secondary logic makes that a bit tricky.

Kudu.Core/Jobs/JobsManagerBase.cs
mathewc: I don't see any unit tests for any of this tricky logic. That would be very useful - tests that verify this.

ahmelsayed: There is a functional test [`JobsScriptsShouldBeChosenByOrder()`](https://github.com/projectkudu/kudu/blob/master/Kudu.FunctionalTests/Jobs/WebJobsTests.cs#L697-L747) for this logic. I'm adding coverage for the `DnxScriptHost`

mathewc: Cool, thanks. Generally, I look for unit tests by name, or by finding all references to the type, which is why I wasn't finding any. Functional tests are good, but it's ideal to have unit tests as well for everything :) Much easier to run/debug.

davidebbo: We can have the bigger functional/unit test discussion. As it stands, we know we don't have good coverage, and this is pre-existing code.

mathewc: Yep, understand.

mathewc:
Rest of code looks good to me - no further comments :)

Kudu.Core/Jobs/JobsManagerBase.cs
ahmelsayed: Added a test case in bb94ee0c9d41eade914b747fa56aa73094c3a239. Also updated [`JobsScriptsShouldBeChosenByOrder()`](https://github.com/projectkudu/kudu/blob/bb94ee0c9d41eade914b747fa56aa73094c3a239/Kudu.FunctionalTests/Jobs/WebJobsTests.cs#L715) to look for `project.json`

mathewc:
Nice :)

PR #1667, opened 2015-08-24
Added script to select latest version.

shrimpy:


davidebbo:
I don't really understand the context of this change. When is this script to get latest version being used? Probably more comments needed.

davidebbo:
Is this related to Go support?

Kudu.Core/Deployment/WellKnownEnvironmentVariables.cs
```diff
@@ -36,5 +36,6 @@ internal static class WellKnownEnvironmentVariables
         public const string SiteBitness = "SITE_BITNESS";
 
         public const string GoWebConfigTemplate = "GO_WEB_CONFIG_TEMPLATE";
+        public const string SelectLatestVersionCommandKey = "SELECT_LATEST_VERSION_CMD";
```
mathewc: Yeah, as David pointed out, this is probably about selecting the latest version of Go, in which case the script and code members should be named appropriately.

davidebbo: I think the reasoning is that the SelectLatest command can in theory be used for any such scenario, so it's not Go specific, even if as it stands that's the use case. So I'm ok with the naming, but a few more comments in the code about what this command is would be helpful.

Kudu.Core/Scripts/selectLatestVersion.ps1
```diff
@@ -0,0 +1,17 @@
+﻿PARAM(
```
mathewc: Can't you get the go version more simply by executing "go version"?

davidebbo: That would get you the version of whatever go you're running. The goal is to find the latest go version among the various versions we have on the box. So it comes down to searching through folders.

Kudu.Core/Deployment/WellKnownEnvironmentVariables.cs
shrimpy: right, it is not Go specific. i will add some more command to describe it. basically it only look at the folders from a given path (NOT recursive), and find the largest version.

shrimpy:
merged in https://github.com/projectkudu/kudu/commit/31788cd15034074b321eab546637de47dd72a132

PR #1673, opened 2015-08-27
Request passthrough webjobs

ahmelsayed:
The webjob has to have the following in `settings.job` to get a port number from kudu

```json
{ "is_webhook": true }
```

### DIFF
--- a/Kudu.Contracts/Jobs/IContinuousJobsManager.cs
+++ b/Kudu.Contracts/Jobs/IContinuousJobsManager.cs
@@ -1,9 +1,14 @@
-﻿namespace Kudu.Contracts.Jobs
+﻿using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace Kudu.Contracts.Jobs
 {
     public interface IContinuousJobsManager : IJobsManager<ContinuousJob>
     {
         void DisableJob(string jobName);
 
         void EnableJob(string jobName);
+
+        Task<HttpResponseMessage> HandleRequest(string jobName, string path, HttpRequestMessage request);
     }
 }
--- a/Kudu.Contracts/Jobs/JobSettings.cs
+++ b/Kudu.Contracts/Jobs/JobSettings.cs
@@ -31,7 +31,6 @@ namespace Kudu.Contracts.Jobs
                 return GetSetting(JobSettingsKeys.IsSingleton, false);
             }
         }
-
         public TimeSpan GetStoppingWaitTime(long defaultTime)
         {
             return TimeSpan.FromSeconds(GetSetting(JobSettingsKeys.StoppingWaitTime, defaultTime));
@@ -46,5 +45,6 @@ namespace Kudu.Contracts.Jobs
         {
             return GetSetting<string>(JobSettingsKeys.Schedule);
         }
+
     }
 }
--- a/Kudu.Core/Deployment/Generator/ExternalCommandFactory.cs
+++ b/Kudu.Core/Deployment/Generator/ExternalCommandFactory.cs
@@ -87,7 +87,8 @@ namespace Kudu.Core.Deployment.Generator
                 Path.GetDirectoryName(PathUtility.ResolveMSBuildPath()),
                 Path.GetDirectoryName(PathUtility.ResolveGitPath()),
                 Path.GetDirectoryName(PathUtility.ResolveVsTestPath()),
-                Path.GetDirectoryName(PathUtility.ResolveSQLCmdPath())
+                Path.GetDirectoryName(PathUtility.ResolveSQLCmdPath()),
+                Path.GetDirectoryName(PathUtility.ResolveFSharpCPath())
             };
 
             toolsPaths.AddRange(PathUtility.ResolveNodeNpmPaths());
--- a/Kudu.Core/Deployment/WellKnownEnvironmentVariables.cs
+++ b/Kudu.Core/Deployment/WellKnownEnvironmentVariables.cs
@@ -29,6 +29,7 @@
         public const string WebJobsRunId = "WEBJOBS_RUN_ID";
         public const string WebJobsShutdownNotificationFile = "WEBJOBS_SHUTDOWN_FILE";
         public const string WebJobsCommandArguments = "WEBJOBS_COMMAND_ARGUMENTS";
+        public const string WebJobsPort = "WEBJOBS_PORT";
 
         public const string CommitId = "SCM_COMMIT_ID";
         public const string DnvmPath = "SCM_DNVM_PS_PATH";
--- a/Kudu.Core/Jobs/ContinuousJobRunner.cs
+++ b/Kudu.Core/Jobs/ContinuousJobRunner.cs
@@ -6,6 +6,8 @@ using Kudu.Contracts.Jobs;
 using Kudu.Contracts.Settings;
 using Kudu.Core.Infrastructure;
 using Kudu.Core.Tracing;
+using System.Net.Sockets;
+using System.Net;
 
 namespace Kudu.Core.Jobs
 {
@@ -54,6 +56,8 @@ namespace Kudu.Core.Jobs
             get { return TimeSpan.MaxValue; }
         }
 
+        internal int WebJobPort { get; private set; }
+
         private void StartJob(ContinuousJob continuousJob)
         {
             // Do not go further if already started or job is disabled
@@ -93,7 +97,8 @@ namespace Kudu.Core.Jobs
                         using (new Timer(LogStillRunning, null, TimeSpan.FromHours(1), TimeSpan.FromHours(12)))
                         {
                             InitializeJobInstance(continuousJob, _continuousJobLogger);
-                            RunJobInstance(continuousJob, _continuousJobLogger, String.Empty, String.Empty);
+                            WebJobPort = GetAvailableJobPort();
+                            RunJobInstance(continuousJob, _continuousJobLogger, String.Empty, String.Empty, WebJobPort);
                         }
 
                         if (_started == 1 && !IsDisabled)
@@ -295,6 +300,20 @@ namespace Kudu.Core.Jobs
                     _continuousJobLogger.Dispose();
                     _continuousJobLogger = null;
                 }
+            }
+        }
+
+        private int GetAvailableJobPort()
+        {
+            var listener = new TcpListener(IPAddress.Loopback, 0);
+            try
+            {
+                listener.Start();
+                return ((IPEndPoint)listener.LocalEndpoint).Port;
+            }
+            finally
+            {
+                listener.Stop();
             }
         }
     }
--- a/Kudu.Core/Jobs/ContinuousJobsManager.cs
+++ b/Kudu.Core/Jobs/ContinuousJobsManager.cs
@@ -2,17 +2,22 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Net.Http;
 using System.Text;
 using Kudu.Contracts.Jobs;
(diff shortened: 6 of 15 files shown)


Kudu.Contracts/Jobs/JobSettings.cs
```diff
@@ -46,5 +46,10 @@ public string GetSchedule()
         {
             return GetSetting<string>(JobSettingsKeys.Schedule);
         }
+
+        public bool GetIsWebHook()
```
davidebbo: 'GetIs' sounds funny. Just `IsWebHook`?

Kudu.Core/Deployment/Generator/ExternalCommandFactory.cs
```diff
@@ -87,7 +87,8 @@ public Executable BuildCommandExecutable(string commandPath, string workingDirec
                 Path.GetDirectoryName(PathUtility.ResolveMSBuildPath()),
                 Path.GetDirectoryName(PathUtility.ResolveGitPath()),
                 Path.GetDirectoryName(PathUtility.ResolveVsTestPath()),
-                Path.GetDirectoryName(PathUtility.ResolveSQLCmdPath())
+                Path.GetDirectoryName(PathUtility.ResolveSQLCmdPath()),
+                Path.GetDirectoryName(PathUtility.ResolveFSharpCPath())
```
davidebbo: Not related to this PR, right?

Kudu.Core/Jobs/ContinuousJobRunner.cs
```diff
@@ -297,5 +302,21 @@ protected virtual void Dispose(bool disposing)
                 }
             }
         }
+
+        private int GetAvailablePortForWebJob()
+        {
+            if (JobSettings.GetIsWebHook())
+            {
+                TcpListener listener = new TcpListener(IPAddress.Loopback, 0);
```
davidebbo: var :)

Kudu.Core/Jobs/ContinuousJobRunner.cs
```diff
@@ -297,5 +302,21 @@ protected virtual void Dispose(bool disposing)
...
+                listener.Start();
+                int port = ((IPEndPoint)listener.LocalEndpoint).Port;
+                listener.Stop();
```
davidebbo: Maybe stop in a finally, to be really sure we don't leak anything? Plus this way you don't need the `port` variable at all...

Kudu.Contracts/Jobs/JobSettings.cs
ahmelsayed: It was to be consistent with `GetIsInPlace` but yes I'll fix it.

Kudu.Core/Deployment/Generator/ExternalCommandFactory.cs
ahmelsayed: I also added support for F# scripts `*.fsx`  as webjobs, which I was using for testing, and that requires the path to `fsi.exe` on the path. I can remove it, but I figured it why not have it.

davidebbo: You should leave it.

amitapl:
why "is_webhook"? seems more generic than that.

ahmelsayed:
Yes I guess the main scenario that we have in mind is WebHook triggers for the SDK but you're right it should be called `is_request_pass_through`? thoughts?

amitapl:
maybe `enable_request_pass_through`

Kudu.Contracts/Jobs/IContinuousJobsManager.cs
```diff
@@ -1,9 +1,14 @@
...
+
+namespace Kudu.Contracts.Jobs
 {
     public interface IContinuousJobsManager : IJobsManager<ContinuousJob>
     {
         void DisableJob(string jobName);
 
         void EnableJob(string jobName);
+
+        Task<HttpResponseMessage> RequestPassthrough(string jobName, string path, HttpRequestMessage request);
```
mathewc: Can you make this method name a verb - perhaps HandleRequest, ProcessRequest, etc.?

Kudu.Core/Deployment/WellKnownEnvironmentVariables.cs
```diff
@@ -29,6 +29,7 @@ internal static class WellKnownEnvironmentVariables
         public const string WebJobsRunId = "WEBJOBS_RUN_ID";
         public const string WebJobsShutdownNotificationFile = "WEBJOBS_SHUTDOWN_FILE";
         public const string WebJobsCommandArguments = "WEBJOBS_COMMAND_ARGUMENTS";
+        public const string WebJobPort = "WEBJOB_PORT";
```
mathewc: All the other ver names are plural - why not this one? (i.e.. WEBJOBS_PORT)

Kudu.Contracts/Jobs/JobSettings.cs
```diff
@@ -31,6 +31,13 @@ public bool IsSingleton
                 return GetSetting(JobSettingsKeys.IsSingleton, false);
             }
         }
+        public bool IsWebHook
```
mathewc: Agree that the naming of this should be generalized. Who sets the setting, and how is it used?

Kudu.Core/Jobs/ContinuousJobRunner.cs
```diff
@@ -297,5 +302,26 @@ protected virtual void Dispose(bool disposing)
                 }
             }
         }
+
+        private int GetAvailablePortForWebJob()
```
mathewc: nit: GetAvailablePort?

Kudu.Contracts/Jobs/JobSettingsKeys.cs
```diff
@@ -6,5 +6,6 @@ public static class JobSettingsKeys
         public const string StoppingWaitTime = "stopping_wait_time";
         public const string IsInPlace = "is_in_place";
         public const string Schedule = "schedule";
+        public const string IsWebHook = "is_webhook";
```
davidebbo: I don't really like using the term 'passthrough' for this setting, because the WebJobs author doesn't see it as such.

Let's step back: how expensive is it to find an available port? If it's cheap enough, we should just do away with this setting and always do it. Are there any blatant drawbacks?

Kudu.Core/Jobs/ContinuousJobRunner.cs
```diff
@@ -297,5 +302,26 @@ protected virtual void Dispose(bool disposing)
...
+            {
+                var listener = new TcpListener(IPAddress.Loopback, 0);
+                try
+                {
+                    listener.Start();
+                    return ((IPEndPoint)listener.LocalEndpoint).Port;
+                }
+                finally
+                {
+                    listener.Stop();
```
mathewc: You start/stop a listener here to determine an available port at a moment in time. Will that port remain available/the same?

Kudu.Core/Jobs/ContinuousJobsManager.cs
```diff
@@ -57,6 +62,39 @@ public void EnableJob(string jobName)
...
+        public async Task<HttpResponseMessage> RequestPassthrough(string jobName, string path, HttpRequestMessage request)
+        {
+            using (var client = new HttpClient())
+            {
+                var forwardRequest = GetWebJobForwardRequest(jobName, path, request);
+                return await client.SendAsync(forwardRequest);
+            }
+        }
+
+        private HttpRequestMessage GetWebJobForwardRequest(string jobName, string path, HttpRequestMessage original)
```
mathewc: nit: GetForwardRequest? (basically I'm questining the "WebJob" name prefix you're using everywhere - seems unecessary)

Kudu.Core/Jobs/ContinuousJobsManager.cs
```diff
@@ -57,6 +62,39 @@ public void EnableJob(string jobName)
...
+            var webJobBaseUri = new Uri(string.Concat("http://", Localhost, ":", jobRunner.WebJobPort));
+            var fullUriPath = new Uri(webJobBaseUri, path);
+            var clone = new HttpRequestMessage(original.Method, fullUriPath);
+
+            if (original.Method != HttpMethod.Get)
+            {
+                clone.Content = original.Content;
+            }
+
+            original.Headers.Aggregate(clone.Headers, (a, b) =>
```
mathewc: Please add unit tests for this request clone logic to verify it clones various request types properly, headers, etc.

Kudu.Core/Jobs/ContinuousJobsManager.cs
```diff
@@ -57,6 +62,39 @@ public void EnableJob(string jobName)
             continuousJobRunner.EnableJob();
         }
 
+        public async Task<HttpResponseMessage> RequestPassthrough(string jobName, string path, HttpRequestMessage request)
```
mathewc: Should this belong on the ContinuousJobRunner class? Also, please add a unit test for this method. I see there is a ContinuousJobRunnerFacts unit test class already. Otherwise, we may need a ContinuousJobManagerFacts class.

Kudu.Core/Jobs/ContinuousJobRunner.cs
davidebbo: The idea is that WebJob will start listening as soon as it starts, grabbing the port. That of course leaves a small race condition window. Whether it's something that can realistically happen depends on how the ports are allocated. If each call, results in a mostly random available port, then the odds of a race are low enough that we don't care. But if the port always come sequentially, then it becomes plausible for two WebJobs (possibly in different sites) to end up with the same port.

ahmelsayed: It's pretty sequential how the OS picks a port number, but I think it's sequential OS wide so I don't think 2 Kudus running on the same machine could get the same port number. 
Also a job in a different site won't be able to steal the request because the sandbox doesn't allow communicating with local ports outside the sandbox so there is no security issue. The only real 100% guaranteed alternative is for kudu to wait for the webjob to pick a random port and then communicate that port back to kudu.

Kudu.Core/Jobs/ContinuousJobsManager.cs
ahmelsayed: As far as I can tell, the ContinuousJobRunner is the one that manages all the runtime needs for the continuous webjobs.

Kudu.Core/Jobs/ContinuousJobRunner.cs
davidebbo: If it's sequential such that 2 calls won't get the same even if neither actually uses the port, then we're fine.

In any case, we have a fallback if some extremely unlikely scenario causes a conflict: the job will fail to start when it tries to listen on the port, and we'll restart it.

So I think we're fine.

Kudu.Contracts/Jobs/JobSettingsKeys.cs
ahmelsayed: I think it's pretty cheap. I think the thinking was that the wide majority of webjobs won't need this, so there is no point in doing this for every continuous webjob but I do agree that not having a setting is more convenient, I'll remove it.

Kudu.Contracts/Jobs/JobSettings.cs
ahmelsayed: Will remove the setting all together.

Kudu.Core/Jobs/ContinuousJobsManager.cs
ahmelsayed: Added a unit test to ContinuousJobsManagerFacts

PR #1678, opened 2015-09-02
Force set webpages:Enabled=true

davidebbo:
This is to make sure Kudu isn't taken down by a site App Setting that
sets it to false.

https://social.msdn.microsoft.com/Forums/en-US/986ff298-0ad9-4580-9e91-b1525df4e27c/no-route-registered-for-of-scm-site?forum=windowsazurewebsitespreview


suwatch:
:shipit:

davidebbo:
This fix is not good as it would also turn on the flag for all site extensions in scm, possibly breaking them.

PR #1686, opened 2015-09-09
reduce git traces in log

suwatch:


davidebbo:
:shipit:

Kudu.TestHarness/Git.cs
```diff
@@ -217,8 +217,9 @@ public static string CloneToLocal(string cloneUri, string path = null)
             string repositoryPath = path ?? Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
             Directory.CreateDirectory(repositoryPath);
             var exe = new GitExecutable(repositoryPath, idleTimeout: TimeSpan.FromSeconds(3600));
-            exe.SetTraceLevel(2);
-            exe.SetHttpVerbose(true);
+            // reduce noises in log
+            //exe.SetTraceLevel(2);
+            //exe.SetHttpVerbose(true);
```
shrimpy: why not remove them if we do not need them?

suwatch: Just taking a safe approach first - see how it goes.

PR #1690, opened 2015-09-13
Allow Kudu PowerShell console to execute scripts

davidebbo:
This is the same flag we already pass when executing PowerShell WebJobs.


ahmelsayed:
:shipit:

PR #1699, opened 2015-09-20
Change the WebJobs invalidation logic to look at latest time

davidebbo:
Previous logic was broken when a file was changed or added in the temp folder. In particular, the .exe.config file which we replace was almost guaranteed to be newer than the source, incorrectly causing the cache to be treated as invalid.

### DIFF
--- a/Kudu.Core/Infrastructure/FileSystemHelpers.cs
+++ b/Kudu.Core/Infrastructure/FileSystemHelpers.cs
@@ -130,6 +130,15 @@ namespace Kudu.Core.Infrastructure
             return Instance.File.GetLastWriteTimeUtc(path);
         }
 
+        // Get the latest LastWriteTime of all the files in a folder
+        public static DateTime GetFolderLatestWriteTimeUtc(string path)
+        {
+            DirectoryInfoBase jobBinariesDirectory = DirectoryInfoFromDirectoryName(path);
+            FileInfoBase[] files = jobBinariesDirectory.GetFiles("*.*", SearchOption.AllDirectories);
+
+            return files.Select(f => f.LastWriteTimeUtc).OrderByDescending(t => t).FirstOrDefault();
+        }
+
         public static void WriteAllBytes(string path, byte[] contents)
         {
             Instance.File.WriteAllBytes(path, contents);
--- a/Kudu.Core/Jobs/BaseJobRunner.cs
+++ b/Kudu.Core/Jobs/BaseJobRunner.cs
@@ -72,19 +72,6 @@ namespace Kudu.Core.Jobs
 
         protected JobSettings JobSettings { get; set; }
 
-        private static int CalculateHashForJob(string jobBinariesPath)
-        {
-            var updateDatesString = new StringBuilder();
-            DirectoryInfoBase jobBinariesDirectory = FileSystemHelpers.DirectoryInfoFromDirectoryName(jobBinariesPath);
-            FileInfoBase[] files = jobBinariesDirectory.GetFiles("*.*", SearchOption.AllDirectories);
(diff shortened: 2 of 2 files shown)


Kudu.Core/Infrastructure/FileSystemHelpers.cs
```diff
@@ -130,6 +130,15 @@ public static DateTime GetLastWriteTimeUtc(string path)
             return Instance.File.GetLastWriteTimeUtc(path);
         }
 
+        // Get the latest LastWriteTime of all the files in a folder
+        public static DateTime GetFolderLatestWriteTimeUtc(string path)
+        {
+            DirectoryInfoBase jobBinariesDirectory = DirectoryInfoFromDirectoryName(path);
+            FileInfoBase[] files = jobBinariesDirectory.GetFiles("*.*", SearchOption.AllDirectories);
+
+            return files.Select(f => f.LastWriteTimeUtc).OrderByDescending(t => t).FirstOrDefault();
```
amitapl: is it possible for a file with a LastWriteTime that is not the latest one in the folder to be updated (and still have its LastWriteTime as not the latest)?
Also LastWriteTime could be sometime in the future (for whatever reason) which can be a problem.

Kudu.Core/Jobs/BaseJobRunner.cs
```diff
@@ -102,10 +89,14 @@ private void CacheJobBinaries(JobBase job, IJobLogger logger)
             {
                 try
                 {
-                    int currentHash = CalculateHashForJob(JobBinariesPath);
```
mathewc: There is a "Failed to calculate hash for WebJob" error message below that needs to be updated if we're no longer computing hashes

Kudu.Core/Infrastructure/FileSystemHelpers.cs
davidebbo: @amitapl you are correct, and this leaves some scenarios where we'll fail to update. @mathewc is looking at a somewhat more elaborate fix that you handle those cases. Thanks!

Kudu.Core/Jobs/BaseJobRunner.cs
```diff
@@ -102,10 +89,14 @@ private void CacheJobBinaries(JobBase job, IJobLogger logger)
...
-
-                    if (lastHash == currentHash)
+                    // If the newest file in the WebJobs folder is newer than the newest file in the cache,
+                    // treat the cache as invalid.
+                    // Note that we don't want to invalidate the cache if files are added in the cache. Also, we
+                    // update the .exe.config file, and we don't want that to invalidate the cache.
+                    DateTime lastWebJobsWriteTime = FileSystemHelpers.GetFolderLatestWriteTimeUtc(JobBinariesPath);
+                    DateTime lastWorkingDirectoryWriteTime = FileSystemHelpers.GetFolderLatestWriteTimeUtc(WorkingDirectory);
+
+                    if (lastWebJobsWriteTime <= lastWorkingDirectoryWriteTime)
```
suwatch01: Out of curiosity, when the source's time is newer than destination's time, do we do a smart copy (only copy those newer, remove those being deleted etc)?  Given it is more expensive to read the content than to read attribute (last modified time etc).

davidebbo: There is no smart copy. It kills the whole tmp folder and creates a new one. I'd say smart copy is conceivable (e.g. with KuduSync, manifest, ...), but I'm not sure it is worth the complexity. As long as we fixed the logic so it correctly uses the cache when it should, there shouldn't be too much file copying going on.

davidebbo:
@mathewc is working on more elaborate fix.

PR #1701, opened 2015-09-21
Fixing bug in WebJobs change detection logic

mathewc:


Kudu.Core/Jobs/BaseJobRunner.cs
```diff
@@ -140,6 +185,10 @@ private void CacheJobBinaries(JobBase job, IJobLogger logger)
                     UpdateAppConfigs(tempJobInstancePath);
 
                     _workingDirectory = tempJobInstancePath;
+
+                    // cache the file map snapshot for next time (to aid in detecting
+                    // file deletions)
+                    _cachedSourceDirectoryFileMap = sourceDirectoryFileMap;
```
amitapl: shouldn't you write this to the file system, otherwise it seems meaningless (after restart)

Kudu.Core/Jobs/BaseJobRunner.cs
```diff
@@ -72,17 +71,61 @@ protected string WorkingDirectory
...
-            var updateDatesString = new StringBuilder();
-            DirectoryInfoBase jobBinariesDirectory = FileSystemHelpers.DirectoryInfoFromDirectoryName(jobBinariesPath);
+            // enumerate all source directory files, and compare against the files
+            // in the working directory (i.e. the cached directory)
+            FileInfoBase foundEntry = null;
+            foreach (var entry in sourceDirectoryFileMap)
+            {  
+                if (workingDirectoryFileMap.TryGetValue(entry.Key, out foundEntry))
+                {
+                    if (entry.Value.LastWriteTimeUtc > foundEntry.LastWriteTimeUtc)
```
amitapl: I think it should be != as it is possible to get an "older" file in the source.

Kudu.Core/Jobs/BaseJobRunner.cs
```diff
@@ -72,17 +71,61 @@ protected string WorkingDirectory
...
+            Dictionary<string, FileInfoBase> cachedSourceDirectoryFileMap)
         {
-            var updateDatesString = new StringBuilder();
-            DirectoryInfoBase jobBinariesDirectory = FileSystemHelpers.DirectoryInfoFromDirectoryName(jobBinariesPath);
+            // enumerate all source directory files, and compare against the files
+            // in the working directory (i.e. the cached directory)
+            FileInfoBase foundEntry = null;
+            foreach (var entry in sourceDirectoryFileMap)
+            {  
+                if (workingDirectoryFileMap.TryGetValue(entry.Key, out foundEntry))
```
amitapl: shouldn't `workingDirectoryFileMap` actually be `cachedSourceDirectoryFileMap`? this way changes in the working directory won't have affect

Kudu.Core/Jobs/BaseJobRunner.cs
mathewc: I think I can avoid that by also performing this same initialization on the before the "no changes detected" quick return above. At that point, we know source files = working files.

amitapl: not sure I understand do you mean line 154? but it relies on the value in `_cachedSourceDirectoryFileMap`.

Kudu.Core/Jobs/BaseJobRunner.cs
mathewc: I don't think so. I'm trying to compare source v.s. working - on an initial pass, cachedSourceDirectoryFileMap will be null. cachedSourceDirectoryFileMap is only used to detect deletions.

Kudu.Core/Jobs/BaseJobRunner.cs
mathewc: See latest revision. I'm now initializing this in memory cache on both codepaths - ensuring that on job startup, it is always set to the correct value, and we have a cache snapshot taken at a point in time where we know the source/working directories are the same.

Kudu.Core/Jobs/BaseJobRunner.cs
mathewc: Done. Note that in doing this, I'm changing the behavior from the previous revisions. I'm now treating any modifications/deletions to source files in the working directory as changes that will cause a new copy. File additions in the working directory won't trigger a copy (to allow for the case where a job might be writing temp files). I think this makes sense when viewed as a "repair" operation to modified/missing source files. Note this is also how the existing prod code behaves (since it's computing hash codes over the entire directories).

Kudu.Core/Jobs/BaseJobRunner.cs
```diff
@@ -72,17 +71,65 @@ protected string WorkingDirectory
...
+                {
+                    if (entry.Value.LastWriteTimeUtc != foundEntry.LastWriteTimeUtc)
```
mathewc: I'm not sure this handles the exe.config file modifications we do now (for app settings). Previously when I had was using > for the comparison, the modifications made to exe.config wouldn't trigger a dirty directory. Now it does I believe. Might need to special case the exe.config file. Thoughts?

davidebbo: Indeed, as it stands that won't be handled correctly. I think we can't avoid special casing the extension here. We have a list of extensions in the code that you could use to avoid hard coding it here.

davidebbo: And of course that'll leave the edge case where if someone just updates the .exe.config (and no other file), we may not do the right thing. But I can live with that.

Kudu.Core/Jobs/BaseJobRunner.cs
```diff
@@ -72,17 +71,65 @@ protected string WorkingDirectory
...
+                        // file has been modified in the working directory.
+                        return true;
+                    }
+                }
+                else
+                {
+                    // A file exists in source that doesn't exist in our working
+                    // directory. This is either because a file was actually added
+                    // to source, or a file that previously existed in source has been
+                    // deleted from the working directory.
```
davidebbo: 'deleted from the working directory' is hopefully not something that would happen in real world use!

Kudu.Core/Jobs/BaseJobRunner.cs
```diff
@@ -72,17 +71,65 @@ protected string WorkingDirectory
...
+                    }
+                }
+            }
+
+            return false;
+        }
+
+        internal static Dictionary<string, FileInfoBase> GetJobDirectoryFileMap(string sourceDirectory)
+        {
+            Dictionary<string, FileInfoBase> fileMap = new Dictionary<string, FileInfoBase>();
```
davidebbo: Needs to be case insensitive. Someone could update a file and change its case, but it's still the same file.

Kudu.Core/Jobs/BaseJobRunner.cs
```diff
@@ -72,17 +71,65 @@ protected string WorkingDirectory
...
+
+            DirectoryInfoBase jobBinariesDirectory = FileSystemHelpers.DirectoryInfoFromDirectoryName(sourceDirectory);
             FileInfoBase[] files = jobBinariesDirectory.GetFiles("*.*", SearchOption.AllDirectories);
             foreach (FileInfoBase file in files)
```
davidebbo: Whole loop in 1-liner if you use files.ToDictionary(). Also eliminates the separate dictionary creation.

Kudu.Core/Jobs/BaseJobRunner.cs
mathewc: Yes - however for completeness I think it makes sense to have any modifications or deletions of the initial set of job files trigger a restorative recopy/cache. We ignore file additions in the working/temp directory.

Kudu.Core/Jobs/BaseJobRunner.cs
mathewc: I'll special case the config file then. I'll do some full e2e/integration testing on that to verify the behavior.

Kudu.Core/Jobs/BaseJobRunner.cs
```diff
@@ -72,17 +71,65 @@ protected string WorkingDirectory
 
         protected JobSettings JobSettings { get; set; }
 
-        private static int CalculateHashForJob(string jobBinariesPath)
+        internal static bool JobDirectoryHasChanged(
```
davidebbo: One it has changed, might be nice to log the file that caused it. Would have saved me a lot of time if we had had that when I first debugged the .exe.config condition :)

Kudu.Core.Test/Jobs/BaseJobRunnerFacts.cs
```diff
@@ -0,0 +1,172 @@
...
+
+            // add some files in root
+            File.WriteAllText(Path.Combine(_testJobSourceDir, "test1.txt"), "test");
+            File.WriteAllText(Path.Combine(_testJobSourceDir, "test2.txt"), "test");
+            File.WriteAllText(Path.Combine(_testJobSourceDir, "test3.txt"), "test");
+
+            // add some files in a sub directory
+            string testSubDir = Path.Combine(_testJobSourceDir, "subdir");
+            Directory.CreateDirectory(testSubDir);
+            File.WriteAllText(Path.Combine(testSubDir, "test1.txt"), "test");
```
davidebbo: Vary the casing on some files

Kudu.Core/Jobs/BaseJobRunner.cs
amitapl: maybe I'm not seeing this right but you are initializing `_cachedSourceDirectoryFileMap` only after the first call to check the cache in which you use this value but since it's null at first call it is basically useless.

Kudu.Core/Jobs/BaseJobRunner.cs
amitapl: maybe its easier to set the `LastWriteTime` of the app.config so it seems like it wasn't updated.
with that I'm not sure what is the difference between this and the previous code.

davidebbo: Yes, I like that idea much better than special casing the extension.

Kudu.Core/Jobs/BaseJobRunner.cs
mathewc: On first call of CacheJobBinaries, WorkingDirectory will be null, so we don't do the JobDirectoryHasChanged check at all. We copy/cache the job files and set WorkingDirectory at that point, **along with _cachedSourceDirectoryFileMap**. So on subsequent runs that will be set and the map contents will be the initial unmodified set of files we cached.

Is my understanding correct here? I'm going to do some e2e testing now to confirm.

amitapl: got it, but `WorkingDirectory` will not be `null` - https://github.com/projectkudu/kudu/blob/mathewc-work/Kudu.Core/Jobs/BaseJobRunner.cs#L149-L159

Kudu.Core/Jobs/BaseJobRunner.cs
mathewc: Yes, making case insensitive.

Kudu.Core/Jobs/BaseJobRunner.cs
mathewc: Yes, will do.

Kudu.Core/Jobs/BaseJobRunner.cs
mathewc: Adding descriptive log entries (along with tests verifying the logs)

Kudu.Core.Test/Jobs/BaseJobRunnerFacts.cs
mathewc: Added a test verifying case insensitivity

Kudu.Core/Jobs/BaseJobRunner.cs
mathewc: _workingDirectory is only set after all the files are copied to the cache dir, right? Before that it is null, and WorkingDirectory is null. So from what I see in the code, when a job is initialized, WorkingDirectory is null until the initial copy is complete. Can you please come over to my desk so we can discuss? :)

Kudu.Core/Jobs/BaseJobRunner.cs
mathewc: Agreed. Will do.

Kudu.Core/Jobs/BaseJobRunner.cs
amitapl: omw :+1:

Kudu.Core/Jobs/BaseJobRunner.cs
mathewc: Done. Also added a test verifying that the LastUpdateTime is maintained.

Kudu.Core/Jobs/BaseJobRunner.cs
mathewc: To close on this, we only do the directory diff when WorkingDirectory is not null, and whenever WorkingDirectory is not null, we have set _cachedSourceDirectoryFileMap to the initial set of source files we copied/cached. So we're good here. We're doing actual e2e tests to confirm this.

PR #1703, opened 2015-09-22
improve onedrive concurrency

suwatch:
The key fix is the concurrency.  Instead of base on the Processor Count, it is set to 6 concurrent downloading threads (same approach as dropbox but slight different value).   

The remaining fix is making the status more diagnosable.


davidebbo:
:shipit:

PR #1704, opened 2015-09-23
fix changes count reporting for OneDrive

suwatch:
This is a fix to report changes count more realistic for OneDrive.   I did a simple fix (prefer file changes count) which is most common scenario.
  - initial sync
  - add new file
  - update file

A more complicated fix would address directory remove - but risk is higher and may not be worth it.


davidebbo:
:shipit:

PR #1706, opened 2015-09-24
Fix search Find Handle dialog width

snobu:
Added width property to `Find Handle` dialog in __/ProcessExplorer__

#### Before ####
before

#### After ####
after


suwatch01:
This looks trivial enough.  It is somewhat improve user experience than the default size.  Not sure how to reason about 800 or 600 or 700 width size, though.   We could start with that and see if any feed back.

snobu:
I looked at making the width dynamic but quickly realized I'm waaay over engineering this. I mean if it's a file under a nodejs module path.. There aren't enough pixels in the world to fit that in.

PR #1712, opened 2015-09-29
fix casing issue with VSO RepoUrl

suwatch:


davidebbo:
:shipit:

Kudu.Services/ServiceHookHandlers/VSOHandler.cs
```diff
@@ -45,7 +45,7 @@ protected virtual GitDeploymentInfo GetDeploymentInfo(HttpRequestBase request, J
 
             // even it is empty password we need to explicitly say so (with colon).
             // without colon, LibGit2Sharp is not working
-            Uri remoteUrl = new Uri(_settings.GetValue("repoUrl"));
+            Uri remoteUrl = new Uri(_settings.GetValue("RepoUrl"));
```
suwatch: Long story short, there is existing constant in Azure and we should simply reuse that.

PR #1720, opened 2015-10-01
Fixes #1711, clean up unused code.

### LINKED ISSUE #1711: Debug build is failing code analysis
```
build.cmd Debug
```

it's failing because [`file` is declared but not used](https://github.com/projectkudu/kudu/blob/ee55a13a5b3cdba8623d74fad40cf731f81c38bc/Kudu.Core/SiteExtensions/SiteExtensionStatus.cs#L145-L152). I'm not sure what the loop is for

shrimpy:
Removed unused for-loop. for loop was implement when working on the old design, where status setting file sit under each siteextension`s folder, once installation finished, status setting file will be removed. 

Later we move the status setting file to a separate folder and keep it there. So that we can keep track of the status and errors if any, but miss the clean up (the for loop doesn`t do any harm other than waste some CPU cycle)

### DIFF
--- a/Kudu.Core/SiteExtensions/SiteExtensionStatus.cs
+++ b/Kudu.Core/SiteExtensions/SiteExtensionStatus.cs
@@ -124,8 +124,9 @@ namespace Kudu.Core.SiteExtensions
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


davidebbo:
:shipit:

PR #1724, opened 2015-10-05
[Ace] Added Help and Save without closing (CTRL-S)

snobu:
#### New feature - Save file without closing ####

The gif can't tell the full story, the status bar color turns green on a smooth half-second bezier curve. Really polished :)

ctrls

#### Better error handling for both CTRL+S and Save button ####
If you can't save (for whatever reason), your changes are not lost, you are dropped back into the editor.

alert

#### Ace has some really great features. I've picked a few _seriously_ useful ones. Get Help! ####
acehelpbutton

Still on Ace 1.1.9.
1.2.0 is out for a month or so but i need to test it heavily before upgrading.

Not sure why the commit looks like i've completely re-written `Default.cshtml`, i've only added this modal here (line 115):

https://github.com/projectkudu/kudu/compare/acedev?expand=1#diff-b82317deb2356eb11b04b8758c5a1853R115

### DIFF
--- a/Kudu.Services.Web/Content/Scripts/FileBrowser.js
+++ b/Kudu.Services.Web/Content/Scripts/FileBrowser.js
@@ -9,18 +9,21 @@ var statusbar = {
             catch(e) {
                 filename = 'Can not get filename. See console for details.';
                 if (typeof console == 'object') {
-                    console.log('Can not get filename: %s', e);
+                    console.error('Can not get filename: %s', e);
                 }
             }
             finally {
                 $('#statusbar').text(filename);
-                $('#statusbar').removeClass('statusbar-red');
             }
         },
     reset:
         function () {
             $('#statusbar').text('');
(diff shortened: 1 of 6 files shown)


suwatch:
@snobu  looks good.  note: I simulated the error by trying to edit read-only file.

snobu:
Thanks! I was using `tail -f filename` in another window to keep the file handle open.

ahmelsayed:
This is awesome!! I can't tell you how many times I hit ctrl+s and had the chrome save window pop to me. Thank you for adding this

snobu:
This still happens if the editor is not in focus when hitting CTRL+S. I still have to come up with a way to steal the default CTRL+S behavior from the browser. I can do it in some browsers but not across the board. Needs more research.

davidebbo:
Nice! I'm always hitting Ctrl-S on that thing and it's always a source of frustration, so a welcome change.

PR #1725, opened 2015-10-06
[Ace] Fix statusbar max width

snobu:
statusbar_max_width

Fix for super long filenames breaking statusbar layout.


davidebbo:
FYI. it's live in Azure, along with all previous changes!

PR #1728, opened 2015-10-08
Proper cleanup site contents before reusing the site

suwatch:
This is up for debate.  

Every now and then (2 in the last 8 runs), one or two tests failed due to files being locked.   This was due to reusing dirty sites.  The fix is to cleanup more reliably by stopping app pool, cleanup site contents and restarting it.  

Pros is hopefully tests run more reliable.  Cons is it adds 5-10s more for each test (stop/start app pool and cold start Kudu).  Given roughly 200 functional tests, with 4 test in parallel, my calculation is we are looking at ~5-10 mins extra time for CI to complete.   Currently CI took 15-20mins.

Thought?

### DIFF
--- a/Kudu.Core/Infrastructure/FileSystemHelpers.cs
+++ b/Kudu.Core/Infrastructure/FileSystemHelpers.cs
@@ -360,9 +360,9 @@ namespace Kudu.Core.Infrastructure
             {
                 fileSystemInfo.Attributes = FileAttributes.Normal;
             }
-            catch
+            catch (Exception ex)
             {
-                if (!ignoreErrors) throw;
+                if (!ignoreErrors) throw new InvalidOperationException(String.Format("Set {0} attribute failed with {1}", fileSystemInfo.FullName, ex.Message), ex);
             }
 
             var directoryInfo = fileSystemInfo as DirectoryInfoBase;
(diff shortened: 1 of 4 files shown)


Kudu.Core/Infrastructure/FileSystemHelpers.cs
```diff
@@ -360,9 +360,9 @@ private static void DeleteFileSystemInfo(FileSystemInfoBase fileSystemInfo, bool
             {
                 fileSystemInfo.Attributes = FileAttributes.Normal;
             }
-            catch
+            catch (Exception ex)
             {
-                if (!ignoreErrors) throw;
+                if (!ignoreErrors) throw new InvalidOperationException(String.Format("Set {0} attribute failed with {1}", fileSystemInfo.FullName, ex.Message));
```
davidebbo: Should you pass the inner exception to preserve the stack?

Kudu.SiteManagement/SiteManager.cs
```diff
@@ -347,6 +345,57 @@ public bool RemoveSiteBinding(string applicationName, string siteBinding, SiteTy
...
         }
 
+        public async Task StartAppPool(string applicationName)
+        {
+            using (ServerManager iis = GetServerManager())
+            {
+                string appPoolName = GetAppPool(applicationName);
+                var kuduAppPool = iis.ApplicationPools[appPoolName];
+
+                if (kuduAppPool.State != ObjectState.Started)
```
shrimpy: will there be a case that we try to start the app pool while it is in "Starting" status? if so, will it work?

Kudu.TestHarness/SitePool.cs
```diff
@@ -163,6 +163,23 @@ private static void WriteIndexHtml(ApplicationManager appManager)
                 throw;
             }
         }
+
+        private static void CleapUpContents(string path)
```
davidebbo: Typo 'Cleap' -> 'Clean'

Kudu.Core/Infrastructure/FileSystemHelpers.cs
suwatch: will do.

Kudu.TestHarness/SitePool.cs
suwatch: fix, thank.

Kudu.SiteManagement/SiteManager.cs
suwatch: if the appPool is in Starting state, it's already in a failure situation.   Whether this appPool start succeeds or fails will not worsen the situation.

snobu:
Ah, now i know why my builds were failing due to locks, randomly, for no real reason. Cleaning the solution from VS and the re-running `build.cmd` worked every time. Thanks for fixing this!

PR #1729, opened 2015-10-08
added modal for uploading files

badescuga:
When you start drag-and-dropping files, you see the percent near the loader as till now. 
screenshot_12

Once you push the percentage button, you get a modal showing your recent uploaded files, and the ones in progress. 

screenshot_13

**Performance impact** (rated in according with last week's pull request).

When the modal is closed, the impact is 0. Extra, i added an auto-cleaner for the cache once it got too big.

When the modal is opened, the only extra impact is the draw of the knockout collection object.

=================

**Personal observations**
The modal performed quite well even with hundred of files. Didn't see any lag or drawbacks. Did also clean the previous implementation a bit.

### DIFF
--- a/Kudu.Services.Web/Content/Scripts/FileBrowser.js
+++ b/Kudu.Services.Web/Content/Scripts/FileBrowser.js
@@ -41,7 +41,65 @@ var statusbar = {
         }
 }
 
+var copyProgressObjects = {};
+var copyObjectsManager = {
+    init: function() {
+        copyProgressObjects = {};
+    },
+    addCopyStats: function (uri, loadedData, totalData
+   )  { 
+        if (copyProgressObjects[uri]) {
+            if (loadedData === totalData) { 
+                copyProgressObjects[uri].endDate = $.now();
+            }
+        } else {
+            copyProgressObjects[uri] = {};
+            copyProgressObjects[uri].startDate = $.now();
+            //this is used for when copying multiple files in the same time so that i may stii have a coherent percentage
+            copyProgressObjects[uri].transactionPackFinished = false; 
+        }
+
+        copyProgressObjects[uri].loadedData = loadedData;
+        copyProgressObjects[uri].totalData = totalData;
+    },
+    getCopyStats: function () {
+        return copyProgressObjects;
+    },
+    getCurrentPercentCompletion: function () {
+        var currentTransfered = 0;
+        var finalTransfered = 0;
+        var foundItem = false;
+
+        for (var key in copyProgressObjects) {
+            var co = copyProgressObjects[key];
+            if(co.transactionPackFinished === false) {
+                foundItem = true;
+                currentTransfered += co.loadedData;
+                finalTransfered += co.totalData;
+            }
+        }
+
+        var perc = 0;
+        if (foundItem) {
+            perc = parseInt((currentTransfered / finalTransfered) * 100);
+        } else { // to avoid 0/0
+            perc = 100;
+        }
+
+        if (perc === 100 && foundItem) { // if all transactions have finished & have some unmarked transaction pack, cancel it out
+            for (var key in copyProgressObjects) {
+                copyProgressObjects[key].transactionPackFinished = true;
+            }
+        }
+
+        return perc;
+    }
+}
+
 var statusbarObj = Object.create(statusbar);
+var copyObjectsManagerObj = Object.create(copyObjectsManager);
+copyObjectsManagerObj.init();
+
 
 $.connection.hub.url = appRoot + "api/filesystemhub";
 var fileSystemHub = $.connection.fileSystemHub;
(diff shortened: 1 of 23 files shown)


shrimpy:
image

Love this feature, it is good when uploading large files and see the progress.
Notice that the percentage is for each file. if i drap and drop multiple files, i will see multiple "100%". will be good to show an overall progress, e.g drap and drop 10 files, once a file is uploaded, progress 10%.

Kudu.Services.Web/Content/Scripts/FileBrowser.js
```diff
@@ -80,39 +80,47 @@ function showAceHelpModal() {
...
+        this._infoMessage = message;
     },
-    addCopyStats: function (uri, loadedData, totalData
-   )  { 
-        if (copyProgressObjects[uri]) {
-            if (loadedData === totalData) { 
-                copyProgressObjects[uri].endDate = $.now();
+    addCopyStats: function (uri, loadedData, totalData) {
+
+        uri = uri.substring(uri.indexOf('/vfs')+5, uri.length); // slice uri to be prettier
```
shrimpy: for "uri", will be good to add a comment to have a sample uri. make reading the code easier.

Kudu.Services.Web/Content/Scripts/FileBrowser.js
```diff
@@ -80,39 +80,47 @@ function showAceHelpModal() {
...
+                this._copyProgressObjects[uri].copyPackEnded = false;
             }
         } else {
-            copyProgressObjects[uri] = {};
-            copyProgressObjects[uri].startDate = $.now();
-            //this is used for when copying multiple files in the same time so that i may stii have a coherent percentage
-            copyProgressObjects[uri].transactionPackFinished = false; 
+            this._copyProgressObjects[uri] = {};
+            this._copyProgressObjects[uri].startDate = $.now();
+            this._copyProgressObjects[uri].copyPackEnded = false; //this is used for when copying multiple files in the same time so that i may stii have a coherent percentage
```
shrimpy: typo "still" ?

Kudu.Services.Web/Content/Scripts/FileBrowser.js
```diff
@@ -503,9 +535,43 @@ $.connection.hub.start().done(function () {
...
+        
+            if (perc == 100 || perc == 0) {
+                viewModel.isTransferInProgress(false);
+            } else {
+                viewModel.isTransferInProgress(true);
+            }
+
+            //handler for clearing out cache once it gets too large 
+            var currentObjCount = Object.keys(copyObjs).length;
+           // console.log('CURRENT OBJ COUNT: %s', currentObjCount);
```
shrimpy: remove unused debug code

badescuga:
Great to hear you like it! The percentage is done for all files that are
being copied. You see the percentage restarting because i start counting
for files that start uploading. Not all files start as once(some have to
finish uploading before others start).

You can see multifile progress by copying 2 larger files.

shrimpy:
RE: You see the percentage restarting because i start counting for files that start uploading. Not all files start as once

I might be picky about the experience, but seeing the progress restarting seems a bit confusing to me.
I would expect the overall progress bar to start from 0 and progress as file uploaded, only reach 100% when all files are done and only show 100% once as it is overall progress.

Another issue for the current implementation is, for the case I try to upload below 8 files. I saw three “100%” and then the overall percentage goes away while file is still uploading in background.
For the overall progress bar, I would suggest to grab the total file count at the beginning, and as a file is uploaded, we progress base on the file count.



[cid:image002.jpg@01D101B7.584031A0]


[cid:image006.jpg@01D101B7.584031A0]

Thanks,
Xiaomin Wu

badescuga:
I agree with you and have thought about implementing at the start, but
tried to keep it as simple as i can. I can redo this.

badescuga:
I'll take a look at this in a couple of hours to see if there is a quick
fix.

badescuga:
Hi guys,

Found another way to do the processing so that I process all the files at start (at least their size). Will cancel the pull request tomorrow (in like 10-12 hours) and will work on a better solution.

Thanks!
From:  Xiaomin Wu
Reply-To:  projectkudu/kudu
Date:  Thursday 8 October 2015 at 20:52
To:  projectkudu/kudu
Cc:  Alex Badescu
Subject:  Re: [kudu] added modal for uploading files (#1729)

RE: You see the percentage restarting because i start counting for files that start uploading. Not all files start as once 

I might be picky about the experience, but seeing the progress restarting seems a bit confusing to me. 
I would expect the overall progress bar to start from 0 and progress as file uploaded, only reach 100% when all files are done and only show 100% once as it is overall progress. 

Another issue for the current implementation is, for the case I try to upload below 8 files. I saw three “100%” and then the overall percentage goes away while file is still uploading in background. 
For the overall progress bar, I would suggest to grab the total file count at the beginning, and as a file is uploaded, we progress base on the file count. 



[cid:image002.jpg@01D101B7.584031A0] 


[cid:image006.jpg@01D101B7.584031A0] 

Thanks, 
Xiaomin Wu

PR #1731, opened 2015-10-09
create modal for monitoring file copying & fixes for previous solution

badescuga:
Hi guys,

Added monitoring file-copying modal. Fixed previous mentioned issues (#1729). Small changes were needed.

Just as a reminder:

Here is the progress when you start copying.

screenshot_12

And the modal:

screenshot_13

### DIFF
--- a/Kudu.Core.Test/Deployment/AspNet5HelperFacts.cs
+++ b/Kudu.Core.Test/Deployment/AspNet5HelperFacts.cs
@@ -1,4 +1,5 @@
 ﻿using Kudu.Core.Infrastructure;
+using Kudu.Core.SourceControl;
 using Newtonsoft.Json;
 using System;
 using System.Collections.Generic;
@@ -6,6 +7,7 @@ using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
 using Xunit;
+using System.IO;
 
 namespace Kudu.Core.Test.Deployment
 {
@@ -21,8 +23,9 @@ namespace Kudu.Core.Test.Deployment
             };
             FileSystemHelpers.Instance = MockFileSystem.GetMockFileSystem(@"x:\repo\global.json", () => JsonConvert.SerializeObject(sdkObject));
             var arch = AspNet5Helper.GetDefaultAspNet5RuntimeArchitecture();
+            var mockFileFinder = new MockFileFinder();
             //Act
-            var sdk = AspNet5Helper.GetAspNet5Sdk(@"x:\repo");
+            var sdk = AspNet5Helper.GetAspNet5Sdk(@"x:\repo", mockFileFinder);
 
             //Assert
             Assert.NotNull(sdk);
@@ -47,8 +50,10 @@ namespace Kudu.Core.Test.Deployment
             };
             FileSystemHelpers.Instance = MockFileSystem.GetMockFileSystem(@"x:\repo\global.json", () => JsonConvert.SerializeObject(sdkObject));
             var arch = AspNet5Helper.GetDefaultAspNet5RuntimeArchitecture();
+            var mockFileFinder = new MockFileFinder();
+
             //Act
-            var sdk = AspNet5Helper.GetAspNet5Sdk(@"x:\repo");
+            var sdk = AspNet5Helper.GetAspNet5Sdk(@"x:\repo", mockFileFinder);
 
             //Assert
             Assert.NotNull(sdk);
@@ -57,5 +62,13 @@ namespace Kudu.Core.Test.Deployment
             Assert.Equal(sdkObject.sdk.architecture, sdk.Architecture);
         }
 
+    }
+
+    internal class MockFileFinder : IFileFinder
+    {
+        public IEnumerable<string> ListFiles(string path, SearchOption searchOption, params string[] lookupList)
+        {
+            return lookupList.Select(i => FileSystemHelpers.GetFiles(path, i, searchOption)).SelectMany(i => i);
+        }
     }
 }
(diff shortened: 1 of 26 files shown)


shrimpy:
@badescuga thanks for taking the feedback and create this new pull request. 

I notice three issues:

1) when upload a large file (60MB file), progress indicate 100%, but actual uploading is still ongoing (size still increasing and spinner is still spinning)

image

2) when upload multiple file (8 files 10MB to 2MB each), progress bar reach 100% around 5 seconds during my testing, but spinner will still spin for another 5 seconds or so and then file show up on list. 
(notice that sometimes progress bar is correct, i think issue is kind of similar to issue one, when the last file is a larger file, progress will disappear before actual upload is done)

image

3) when select multiple file and drop to "Drop here upload and unzip", we will see multiple 100% overall progress bar again (same issue as in the old pull request)

badescuga:
Hi,

Let me mention every issue by point

I have not modified the spinner of the app or any code that handled it. This issue is present from before I actually added code. The percentage I am showing is the one provided from the ajax request which i’m sure is the correct value (which is a live updated value). I suspect that, for some reason, the view(file browser & spinner) is updated with a delay. 
2. See point 1. 

3.  i’ve seen that there are some 500s been thrown in the browser’s console(chrome) (issue also there before I made any mods). You can even see a reproducible example: drag a non-zip file to the drag-and-drop zone. There are some bugs regarding that unzip-zone, which were present before I got into the code. I even get 500s sometimes when I try to copy multiple .zip files at once. Can you give me some more info regarding the files actually being copied? (zips? How many? Size?)



Regarding the issues already present, i’ll take a look in the coming week, but as I mentioned, they were already present.  

From:  Xiaomin Wu
Reply-To:  projectkudu/kudu
Date:  Monday 12 October 2015 at 20:26
To:  projectkudu/kudu
Cc:  Alex Badescu
Subject:  Re: [kudu] create modal for monitoring file copying & fixes for previous solution (#1731)


@badescuga thanks for taking the feedback and create this new pull request. 

I notice three issues:

1) when upload a large file (60MB file), progress indicate 100%, but actual uploading is still ongoing (size still increasing and spinner is still spinning)

2) when upload multiple file (8 files 10MB to 2MB each), progress bar reach 100% around 5 seconds during my testing, but spinner will still spin for another 5 seconds or so and then file show up on list. 
(notice that sometimes progress bar is correct, i think issue is kind of similar to issue one, when the last file is a larger file, progress will disappear before actual upload is done)

3) when select multiple file and drop to "Drop here upload and unzip", we will see multiple 100% overall progress bar again (same issue as in the old pull request)

shrimpy:
@badescuga 

for issue 1, 2:
Before the progress bar change, spinner and the file list is in sync.
Haven`t look into the code, I guess the 100% on the over all progress bar might only indicate the percentage of data that send to server, not the percentage of data that server accepted. In between, there will be latency. And the progress bar should stay till server confirm actual upload is done just like the spinner.

for issue 3, I use the same sets of file as from issue testing (8 files, 10MB to 2MB each). I can help and look into the 500 later (please open an issue for this). but without looking at the backend for the 500 error, it will be good to confirm from frontend javascript code, if that is related to the symptom that cause overall progress bar disappear and show up multiple time.

badescuga:
For 1&2 you might be right regarding server accepted data, but i don't see any clean way to implement this. The overhead might be too big for a feature that is this small. In my opinion, the current implementation is not as big of a drawback. Feedback is still present, after a few seconds (if the case at all) the loading dissapears.


For 3, i will debug tomorrow more in depth. I did not see this in my tests.


Sent by Outlook for Android



From: Xiaomin Wu

Sent: Monday, October 12, 21:58

Subject: Re: [kudu] create modal for monitoring file copying & fixes for previous solution (#1731)

To: projectkudu/kudu

Cc: Alex Badescu



@badescuga 


for issue 1, 2:

Before the progress bar change, spinner and the file list is in sync.

Haven`t look into the code, I guess the 100% on the over all progress bar might only indicate the percentage of data that send to server, not the percentage of data that server accepted. In between, there will be latency. And the progress bar should stay till server confirm actual upload is done just like the spinner.


for issue 3, I use the same sets of file as from issue testing (8 files, 10MB to 2MB each). I can help and look into the 500 later (please open an issue for this). but without looking at the backend for the 500 error, it will be good to confirm from frontend javascript code, if that is related to the symptom that cause overall progress bar disappear and show up multiple time.

shrimpy:
@badescuga

RE: For 1&2 you might be right regarding server accepted data, but i don't see any clean way to implement this. The overhead might be too big for a feature that is this small. In my opinion, the current implementation is not as big of a drawback. Feedback is still present, after a few seconds (if the case at all) the loading dissapears.

i might confuse you on this. My suggestion is that keep what you have right now, but make the overall progress bar stay till spinner is gone. So that UI are all in sync to present a nicely user experience.

badescuga:
Ohhh got it! My bad! This is an easy fix.

Will also test for 3.

badescuga:
Hi Xiaomin,

Have tested for issue 3 ("Drop here upload and unzip") but was unable to
reproduce what you've said. Made some samples for 8 and 16 files, all with
sizes in the range 2 - 10 mb. Was unable to see the percentage restarting
as you described.

 Have you tried resetting you browser cache? Can you give me more info
about your testing environment?
Can you try and test it in another browser also?

I've got a Windows Server 2012 R2 and have tested in both IE (11) and
Chrome (45) with no restarting of the loading-percentage.

shrimpy:
@badescuga click below link to see my repro 

http://1drv.ms/1R9I12L

i think for issue 3 is not that critical. Let`s focus on the other issue that to have the overall progress bar in sync with spinner

badescuga:
Just pushed #1740 with the fix for issue 1 & 2.Tested also on another
machine. Will look around for 3, although i'm still not able to reproduce
the issue.

PR #1745, opened 2015-10-15
Use NPM 3.3.6 if the default node is 4.1.2 to workaround path issues …

ahmelsayed:
…with aspnet 5

### DIFF
--- a/Kudu.Core/Infrastructure/PathUtility.cs
+++ b/Kudu.Core/Infrastructure/PathUtility.cs
@@ -136,10 +136,23 @@ namespace Kudu.Core.Infrastructure
         private static string ResolveNpmVersion(string nodeVersion)
         {
             string programFiles = SystemEnvironment.GetFolderPath(SystemEnvironment.SpecialFolder.ProgramFilesX86);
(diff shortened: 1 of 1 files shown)


Kudu.Core/Infrastructure/PathUtility.cs
```diff
@@ -136,10 +136,23 @@ private static string ResolveNpmVersion()
         private static string ResolveNpmVersion(string nodeVersion)
         {
             string programFiles = SystemEnvironment.GetFolderPath(SystemEnvironment.SpecialFolder.ProgramFilesX86);
+            string appSettingNpmVersion = SystemEnvironment.GetEnvironmentVariable("APPSETTING_WEBSITE_NPM_DEFAULT_VERSION");
```
davidebbo: Just `WEBSITE_NPM_DEFAULT_VERSION` should work too and is cleaner for the non-Azure case.

davidebbo:
Added to doc: https://github.com/projectkudu/kudu/wiki/Configurable-settings#change-the-npm-version

davidebbo:
Looks good other than minor comment.

ahmelsayed:
Updated and merged. Thanks!

PR #1754, opened 2015-10-20
Remove Kudu from the automatic list of site extensions

davidebbo:
This was an indirect way to get a good link from the Portal, but now that
there is direct Kudu UI, it is no longer useful and just pollutes the list.


suwatch01:
:shipit:

davidebbo:
6de824186fcfe687aa17b3febb4b796b031518df

PR #1756, opened 2015-10-20
VSOHandler fallback to use raw remote url as repo url when session token is not available.

shrimpy:


davidebbo:
:shipit:

suwatch01:
looks good.

PR #1759, opened 2015-10-21
fix #1632

### LINKED ISSUE #1632: Update the logic for browser download popup when stop profiling or creating a dump file
There are two feature requests here:

1) When user clicks on "Stop Profiling" button, it may take few seconds or even minutes for the browser download popup to show up. During this time we want the text of the button to change to "Stopping Profiler..."

2) If "Stop Profiling" returns an error, we want to display that in a pop-up similar to what we do for "Start Profiling", currently nothing shows up unless user does F12 or uses fiddler and look at the response error.

badescuga:
Hi guys,

So, the original issue was:

1) When user clicks on "Stop Profiling" button, it may take few seconds or even minutes for the browser download popup to show up. During this time we want the text of the button to change to "Stopping Profiler..."

2) If "Stop Profiling" returns an error, we want to display that in a pop-up similar to what we do for "Start Profiling", currently nothing shows up unless user does F12 or uses fiddler and look at the response error.

========

1) First of all, the "Stop Profiling" turns red when the user starts profiling. When the user presses "Stop Profiling" he will be prompted with an alert notifying him that it may take a few minutes to generate the report.

2)So, the current implementation uses iframe to download the report. Looking into ProcessController.cs, i see that on generate error, there is a HttpResponseMessage (with an error) sent back to the iframe. The best way i found to treat this was to show an alert with the content of iframe message.

for example: if you run this in the browser console:

```javascript
Utilities.downloadURL("https//kududev3.scm.azurewebsites.net/api/processes/thisisafakeprocessid/profile/stop", true)
```
you will be notified in an alert with the message: 
screenshot_1

I'm not very fond of the alert pop-up, but it does the job.

I suggest we create a general popup in the future that can be called from anywhere.

Let me know if this is ok and(if) changes are needed.


ahmelsayed:
Can you rebase your changes on kudu's master? I'm having trouble rebasing it myself and not getting a clean diff on my machine for the changes. I think it's must have been a while since you pulled from kudu and the 2 repos have diverged quite a bit. At least that's what I'm seeing

Kudu.Services.Web/DebugConsole/Default.cshtml
```diff
@@ -1,186 +1,186 @@
-﻿@{
```
shrimpy: can you fix this file as well. i don`t think you touch every line of the file.  it is hard to see what is your changes.

ahmelsayed: I don't think the PR actually changes any thing in that file other than whitespace. It's hard to see on github, which is why I was trying to use a better diffing tool to look at the PR :)

snobu: This always seem to happen when i make changes to this file and the next commit is @badescuga 's. I'll sort this out tomorrow with him. Admittedly i use Vim to edit stuff, so i might be a hipster, the jury is still out on that.

badescuga:
@ahmelsayed 
I think that we are seeing a lot of commits coming from me because my previous commits were previously “squashed” by the person who merged my last pull request (while on my fork they weren’t).

Will look into this tomorrow. Thanks!

davidebbo:
Yes, we always rebase and squash before merging into master. After your change gets merged, it's best to reset your head to the new merged master.

PR #1761, opened 2015-10-22
 fix #1632; take 2

### LINKED ISSUE #1632: Update the logic for browser download popup when stop profiling or creating a dump file

badescuga:
Hi guys,

Had to create a new pull request; The last one was a mess.

So, the original issue was:

1) When user clicks on "Stop Profiling" button, it may take few seconds or even minutes for the browser download popup to show up. During this time we want the text of the button to change to "Stopping Profiler..."

2) If "Stop Profiling" returns an error, we want to display that in a pop-up similar to what we do for "Start Profiling", currently nothing shows up unless user does F12 or uses fiddler and look at the response error.

1) First of all, the "Stop Profiling" turns red when the user starts profiling. When the user presses "Stop Profiling" he will be prompted with an alert notifying him that it may take a few minutes to generate the report.

2)So, the current implementation uses iframe to download the report. Looking into ProcessController.cs, i see that on generate error, there is a HttpResponseMessage (with an error) sent back to the iframe. The best way i found to treat this was to show an alert with the content of iframe message.

**UPDATED FROM LAST PULL REQUEST**

Created general modal that's clean to integrate into views. Replaced the alerts i used. I plan to add "Ok" and "Cancel" buttons in a future edit & maybe make it a bit more pretty. This helps us as it's not blocking as alerts are.

I wrote a short description in GeneralModal.js for it so you can check it out.


shrimpy:
@badescuga can we use modal from [jquery](http://jqueryui.com/dialog/#default) instead of the new GeneralModal?

Kudu.Services.Web/Content/Scripts/GeneralModal.js
```diff
@@ -0,0 +1,58 @@
...
+
+
+
+var generalModal = {
+    _createIfNotExistant: function () {
+        var popup = $("#generalModal");
+       // console.dir(popup);
```
shrimpy: might want to do some cleanup, see a few commented console.log/console.dir/alert

shrimpy:
IMO, i think it is better to stick with one pull request, so that we do not lost other reviewers and discussion.

snobu:
Rounded corners with a radius of 15px? What is this, 2008?

badescuga:
I agree. I never worked with pull requests till now, but i got it now.

badescuga:
@snobu yes, 10 years after we stopped using vim.

shrimpy:
@snobu 

````
    RE: Rounded corners with a radius of 15px? What is this, 2008?
````

it is not that bad, and since we already have these library. I would suggest to just use what we have.

https://jsfiddle.net/shrimpy/a9oyLscr/2/

image

**Comment by badescuga** [comment]
@shrimpy ok, i'll go with the jquery modal. I can't remember why i didn't go with jquery but i'm sure it's not a big deal. 

But, besides that, is a general modal that you could call just from JS a good idea from your perspective? I'm not fond of writing the same thing twice if we do end up needing a modal in another view, and since this is more of a general component it makes sense (to me) to have it available in a general way.

**Comment by shrimpy** [comment]
@badescuga  thanks for taking the feedback. 

The generalModal seems to provide two good thing:
    1) build-in template
    2) centralize interface and look

the downside are:
    1) lost the flexibility of customize your own template when using it
    2) JQuery interface seems already simple enough, and we are hiding lots of good feature that JQuery expose, e.g assign custom css, size etc ...
    3) We do not have a  lots of dialog right now, and i don`t think we have a lot more in the foreseeable future
    4) to make all dialog look the same, we can just define a global css and make jquery-ui dialog title background to be white and border to 0


so I would suggest right now try to have a global css to override jquery-ui if possible. 
If not, we just keep things simple for now. if later we found certain thing we really want to abstract and share, we then create centralize helper function for it.

**Comment by badescuga** [comment]
@shrimpy got it. I was asking more to make it clear in the future which way i should go if i do encounter such a case. I come from a mobile background and i'm used to prefab-ing the hell out of everything.

**Comment by badescuga** [comment]
@shrimpy I made the modal mods. You can check'em out!

File: Kudu.Services.Web/Content/Scripts/ProcessExplorer.js [hunk 2]
```diff
@@ -158,9 +158,17 @@ var Utilities = (function () {
             iframe.style.display = "none";
             document.body.appendChild(iframe);
         }
-        
+        iframe.onload = function (e) {
+            if (showResponseMessage) {
+                var iframeDocument = iframe.contentDocument || iframe.contentWindow.document; // for both IE and other
+                var iFrameBody = iframeDocument.getElementsByTagName('body')[0];
+                if (iFrameBody.innerText && iFrameBody.innerText.length > 0) {
```
shrimpy: some of the browser might not support innertext
http://stackoverflow.com/questions/1359469/innertext-works-in-ie-but-not-in-firefox

will be better if you just use JQuery library to get the text

badescuga:
@shrimpy replaced innerText.

shrimpy:
@badescuga thanks for updating that. One more issue is, will be good to not auto focus on the "X" when we launch the dialog, it doesn`t look good. Can you adjust it?  Apart from that looks good.

image

badescuga:
@shrimpy I was not seeing that, after clearing the cache i did. Fixed.

shrimpy:
@badescuga looks good. i am going to merge your change in. Thanks for picking up this work item :+1:

shrimpy:
Merged in https://github.com/projectkudu/kudu/commit/8a7739ec049fa46121976d392636731ace2466c0

PR #1764, opened 2015-10-23
Changing CultureInfo to be en-US specific to match inline data

sitereactor:
Running the `ScheduleProvidesNextIntervalCorrectly` test on a non en-US (culture) made this test fail for the cases with `12/31/2014` in the InlineData attribute. Changed CultureInfo to use `en-US` specifically as that is the only culture where that date is valid :wink:


suwatch01:
Wouldn't InvariantCulture be more suitable?

sitereactor:
PR updated to use Invariant instead of `en-US`

suwatch01:
:shipit:

suwatch01:
We will merge the change.

PR #1775, opened 2015-10-28
add decoding punycode for displayname

shibayan:
for multibyte-text languages.


ahmelsayed:
Looks good to me, but do you actually see the punycode right now in the displayName? I'm wondering if the frontEnd is flowing it correctly to the worker

shibayan:
If the MS account name is a multi-byte character, seems to have passed is Punycode. The image below is the case of the Japanese.
punycode

ahmelsayed:
merged in fcdcd81126c7d44bc47ad5524d6b3e9d133386be
Thanks @shibayan :)

Issue #1076, opened 2014-03-24
node.js deployment: Error MSB3428: Could not load the Visual C++ component "VCBuild.exe"


bpasero:
Issue:
npm deployment produces the following error

```
remote: Building the projects in this solution one at a time. To enable parallel build, please add the "/m" switch.
remote: MSBUILD : error MSB3428: Could not load the Visual C++ component "VCBuild.exe". To fix this, 1) install the .NET Framework 2.0 SDK, 2) install Microsoft Visual Studio 2005 or 3) add the location of the component to the system path if it is installed elsewhere.  [D:\home\site\wwwroot\node_modules\tsd\node_modules\bufferstream\node_modules\buffertools\build\binding.sln]
```

To reproduce:
- enable git for a azure web site
- clone locally
- add a package.json (see contents below)
- git push to trigger deployment

package.json
```
{
	"name": "s",
	"description": "s",
	"version": "0.6.0",
	"author": {
    	"name": "Microsoft Corporation"
 	},
	"private": true,
	"engines": { "node": ">= 0.10.0 < 0.11.0" },
	
	"dependencies": {
		"tsd": "*"
	}
}
```

xdansmith:
Are there any plans to remove this limitation? MSBuild is present, so I don't see why the VC++ build tools couldn't be as well.

We're trying to move over to Azure Websites but it seems running node.js is a little trickier than we originally thought because of limitations like this.

phairoh:
Every time that I want to love Azure I run in to issues like this and end up struggling for hours trying to figure out work arounds.

This seems like a very silly limitation as all it would take to fix it is installing the appropriate build tools.  Is there any plan to fix this or are we left to figure out workarounds indefinitely or switch away from Azure to a competing cloud computing provider?

adambuczynski:
+1

DavidRouyer:
This is a blocker issue for a lot of people. It is almost impossible to build a decent workflow or a decent application without one or more native modules.

Nobody wants to put hundred on megabytes of node_modules dependencies inside their source control repository.
And how can we build a continuous deployment script if we can't put the "npm install" command inside our scripts ? (obviously npm install is failing due to node_gyp modules)

Please add VC++ build tools ! We don't want workarounds, we want a first class user experience when using Node.js on Azure !

francesconero:
Yes, being able to build native modules would remove so many workarounds needed to deploy a complex node.js project!

DavidRouyer:
[...]

PR #1777, opened 2015-10-29
Add Environment variables required by MSBuild and node-gyp for ...

ahmelsayed:
…building native projects. Needed for #1076

### DIFF
--- a/Kudu.Core/Deployment/WellKnownEnvironmentVariables.cs
+++ b/Kudu.Core/Deployment/WellKnownEnvironmentVariables.cs
@@ -38,5 +38,9 @@
 
         public const string GoWebConfigTemplate = "GO_WEB_CONFIG_TEMPLATE";
         public const string SelectLatestVersionCommandKey = "SELECT_LATEST_VERSION_CMD";
+
+        public const string GypMsvsVersion = "GYP_MSVS_VERSION";
+        public const string VCTargetsPath = "VCTargetsPath";
+        public const string VCInstallDir140 = "VCInstallDir_140";
     }
 }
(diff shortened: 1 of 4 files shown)


ahmelsayed:
I should add a test case as well. Will do that now

davidebbo:
I usually use (abuse?) the command executor test for things like that to avoid adding new test. https://github.com/projectkudu/kudu/blob/master/Kudu.FunctionalTests/CommandExecutorTests.cs#L31-L45

davidebbo:
:shipit:

ahmelsayed:
I see I'll add it there too, but I actually meant adding a nodejs deployment test that restores the 4 packages I tested with.

davidebbo:
Oh yes, we definitely should have that.

Kudu.Core/Deployment/WellKnownEnvironmentVariables.cs
```diff
@@ -38,5 +38,9 @@ internal static class WellKnownEnvironmentVariables
 
         public const string GoWebConfigTemplate = "GO_WEB_CONFIG_TEMPLATE";
         public const string SelectLatestVersionCommandKey = "SELECT_LATEST_VERSION_CMD";
+
+        public const string GypMsvsVersion = "GYP_MSVS_VERSION";
+        public const string VCTargetsPath = "VCTargetsPath";
```
mathewc: nit: should the variables be prefixed as VC_TargetsPath, VC_InstallDir140, to match the pattern of other vars? (e.g. GO_)

ahmelsayed: These are actually consumed by MSBuild.exe so they have to match what MSBuild is expecting.

davidebbo: We don't pick those. They are existing env variables that the compiler expects.

mathewc: Got it.

PR #1778, opened 2015-10-29
support add deployment history api

suwatch:
Per [REST API spec](https://github.com/projectkudu/kudu/wiki/REST-API#deploymen).

### DIFF
--- a/Kudu.Services/Deployment/DeploymentController.cs
+++ b/Kudu.Services/Deployment/DeploymentController.cs
@@ -25,6 +25,8 @@ namespace Kudu.Services.Deployment
 {
     public class DeploymentController : ApiController
     {
+        private readonly IEnvironment _environment;
+        private readonly IAnalytics _analytics;
         private readonly IDeploymentManager _deploymentManager;
         private readonly IDeploymentStatusManager _status;
         private readonly ITracer _tracer;
@@ -33,6 +35,8 @@ namespace Kudu.Services.Deployment
         private readonly IAutoSwapHandler _autoSwapHandler;
 
         public DeploymentController(ITracer tracer,
+                                    IEnvironment environment,
+                                    IAnalytics analytics,
                                     IDeploymentManager deploymentManager,
                                     IDeploymentStatusManager status,
                                     IOperationLock deploymentLock,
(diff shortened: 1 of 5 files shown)


Kudu.Services/Deployment/DeploymentController.cs
```diff
@@ -79,17 +85,33 @@ public void Delete(string id)
...
+
+                                deployResult.Url = Request.RequestUri;
+                                deployResult.LogUrl = UriHelper.MakeRelative(Request.RequestUri, "log");
+
+                                response = Request.CreateResponse(HttpStatusCode.OK, ArmUtils.AddEnvelopeOnArmRequest(deployResult, Request));
+                                return;
+                            }
+                        }
+
                         if (_autoSwapHandler.IsAutoSwapOngoing())
```
shrimpy: why not do this check before above new code?

suwatch: good catch.  fixed to check for AutoSwap first.

shrimpy: Have another question, might be out of scope of this PR.

seems like during deployment, an AutoSwap could happen, do we need to make sure that during a deployment, no auto swap can be made?

suwatch: that is what `_deploymentLock.LockHttpOperationAsync` is for.   Pending deployment will have this lock.

shrimpy: yes, that will make sure there will not be two deployment happen at the same time. but [AutoSwapHandler](https://github.com/projectkudu/kudu/blob/master/Kudu.Core/Deployment/AutoSwapHandler.cs#L54) doesn`t seem to check deployment lock

suwatch: Umm .. I am not keen with the logic flow in autoswap.   Perhaps, open a new issue to track it.    Unless this fix will make it worse, then we should discuss here.

shrimpy: right, it is out of scope of this PR. since the logic has been working and we do not have issue with it. i think we are good. just curios about it.

:shipit:

Issue #1772, opened 2015-10-27
WEBSITE_NPM_DEFAULT_VERSION is ignored when node is no 4.1.2


Philo:
pull request #1745 allows NPM 3.3.6 to be used when explicitly settings WEBSITE_NODE_DEFAULT_VERSION to 4.1.2 but the WEBSITE_NPM_DEFAULT_VERSION is otherwise ignored.  Surely it would be better to allow both WEBSITE_NPM_DEFAULT_VERSION and WEBSITE_NODE_DEFAULT_VERSION to be set and honoured individually for a deployment.

For example I have the requirement to use node 0.12.6 but due to node_modules nesting I wish to use npm 3+:

WEBSITE_NODE_DEFAULT_VERSION=0.12.6
WEBSITE_NPM_DEFAULT_VERSION=3.3.6

At the moment my only option is to directly alter my deploy.cmd to fix the path to NPM_JS_PATH

davidebbo:
Setting `WEBSITE_NPM_DEFAULT_VERSION` should override all other mechanisms to select the npm version. Are you seeing otherwise?

suwatch:
@ahmelsayed  the #1745 fix may not be complete.  we may need to also fix the [selector script] (https://github.com/projectkudu/kudu/blob/master/Kudu.Core/Scripts/selectNodeVersion.js#L36) in addition.

ahmelsayed:
what's the scenario that this script is used for that would affect npm?

suwatch:
The script is used during deployment in deploy.cmd to resolve the npm path to be used when calling `npm install --production` on package.json for nodejs deployment.

PR #1780, opened 2015-10-29
Update selectNodeVersion.js to use WEBSITE_NPM_DEFAULT_VERSION. Fixes…

ahmelsayed:
… #1772 

This matched this commit 7b32c189b50a26ab4289f7052c19bd12bfd11e6d

### DIFF
--- a/Kudu.Core/Scripts/selectNodeVersion.js
+++ b/Kudu.Core/Scripts/selectNodeVersion.js
@@ -49,7 +49,7 @@ function getDefaultNpmVersion(nodeVersionPath) {
         var npmLinkPath = path.resolve(nodeVersionPath, 'npm.txt');
         // Determine if there's a link to npm at the node path
         if (!existsSync(npmLinkPath)) {
-            return;
+            return null;
         }
         var npmVersion = fs.readFileSync(npmLinkPath, 'utf8').trim();
         return npmVersion;
--- a/Kudu.Core/Scripts/selectNodeVersion.js
+++ b/Kudu.Core/Scripts/selectNodeVersion.js
@@ -34,13 +34,26 @@ function resolveNpmPath(npmRootPath, npmVersion) {
 
 
 function getDefaultNpmVersion(nodeVersionPath) {
-    var npmLinkPath = path.resolve(nodeVersionPath, 'npm.txt');
-    // Determine if there's a link to npm at the node path
-    if (!existsSync(npmLinkPath)) {
-        return;
+    var appSettingNpmVersion = process.env.WEBSITE_NPM_DEFAULT_VERSION;
+
+    // extract the current node version from the path
+    var nodePathSplited = nodeVersionPath.split(path.sep).filter(function(e) { return e; });
+    var currentNodeVersion = nodePathSplited[nodePathSplited.length - 1];
+
+    if (appSettingNpmVersion) {
+        return appSettingNpmVersion;
+    } else if (currentNodeVersion === '4.1.2') {
+        // This is to preserve parity with kudu's behavior to fix issues with ASP.NET 5 use of npm
+        return '3.3.6';
+    } else {
+        var npmLinkPath = path.resolve(nodeVersionPath, 'npm.txt');
+        // Determine if there's a link to npm at the node path
+        if (!existsSync(npmLinkPath)) {
+            return;
+        }
+        var npmVersion = fs.readFileSync(npmLinkPath, 'utf8').trim();
+        return npmVersion;
     }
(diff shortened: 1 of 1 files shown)


Kudu.Core/Scripts/selectNodeVersion.js
```diff
@@ -34,13 +34,26 @@ function resolveNpmPath(npmRootPath, npmVersion) {
...
-        return;
+    var appSettingNpmVersion = process.env.WEBSITE_NPM_DEFAULT_VERSION;
+
+    // extract the current node version from the path
+    var nodePathSplited = nodeVersionPath.split(path.sep).filter(function(e) { return e; });
+    var currentNodeVersion = nodePathSplited[nodePathSplited.length - 1];
+
+    if (appSettingNpmVersion) {
+        return appSettingNpmVersion;
+    } else if (currentNodeVersion === '4.1.2') {
```
shrimpy: we also added v4.2.1 and v5.0 and both map to NPM 3.3.9. Do we need to do similar special handling for them as well?

shrimpy: take my comment back. npm version will be in npm.txt file. else case should take care of it.

ahmelsayed: If they already map to a 3.x npm, then no. The problem with 4.1.2 is that it was the default node version for sites, but it maps to some 2.x version of NPM. And a vanilla ASP.NET 5 app was running into path length issue with that, so we force npm to be 3.3.6

Kudu.Core/Scripts/selectNodeVersion.js
```diff
@@ -34,13 +34,26 @@ function resolveNpmPath(npmRootPath, npmVersion) {
...
 
 function getDefaultNpmVersion(nodeVersionPath) {
-    var npmLinkPath = path.resolve(nodeVersionPath, 'npm.txt');
-    // Determine if there's a link to npm at the node path
-    if (!existsSync(npmLinkPath)) {
-        return;
+    var appSettingNpmVersion = process.env.WEBSITE_NPM_DEFAULT_VERSION;
+
+    // extract the current node version from the path
+    var nodePathSplited = nodeVersionPath.split(path.sep).filter(function(e) { return e; });
```
shrimpy: is it necessary to have the "filter" statement?

ahmelsayed: I wanted to filter out empty split results. If the path ends in `\` then the last element in the array will be an empty string. I'm not sure if the path here has a trailing slash or not though!

Kudu.Core/Scripts/selectNodeVersion.js
```diff
@@ -34,13 +34,26 @@ function resolveNpmPath(npmRootPath, npmVersion) {
...
+    if (appSettingNpmVersion) {
+        return appSettingNpmVersion;
+    } else if (currentNodeVersion === '4.1.2') {
+        // This is to preserve parity with kudu's behavior to fix issues with ASP.NET 5 use of npm
+        return '3.3.6';
+    } else {
+        var npmLinkPath = path.resolve(nodeVersionPath, 'npm.txt');
+        // Determine if there's a link to npm at the node path
+        if (!existsSync(npmLinkPath)) {
+            return;
```
shrimpy: this function is expected to return a string. but here we return void. Shall we return null instead? or throw error?

Kudu.Core/Scripts/selectNodeVersion.js
shrimpy: i see.

Kudu.Core/Scripts/selectNodeVersion.js
ahmelsayed: good question. let me take a look

ahmelsayed: Updated to return null. It was there originally and the only place it's being used is checking for null/undefined, but I agree that probably returning null is the right thing to do!

shrimpy: k. looks good to me.

suwatch:
I wonder if we can get to a point where we don't need this selectNodeVersion.js script and deploy.cmd simply relies on whatever is in the path.   The path will be determined at only one code path say in NodeJSBuilder.

davidebbo:
@suwatch doing this in `NodeJSBuilder` is not possible, as that would break any custom deployment script. It needs to either happen for all deployments, regardless of Node (may be tricky), or needs to exist inside the script. BTW, note that despite its name, `selectNodeVersion` also generates the web.config if there isn't one.

PR #1782, opened 2015-10-30
use VS2015 msbuild if available

suwatch:


davidebbo:
:shipit: What's with the static? New warnings?

suwatch:
Yes, VS2015 FxCop caught it.   Note that Kudu-CI (has VS2015) will start using new msbuild to generate the Kudu bits.  Wonder whether there is a regression risk.  IMO, there should not be.

PR #1790, opened 2015-11-06
Add support for crash dumps. Fixes #1785

### LINKED ISSUE #1785: Add support for crash dumps
Today, our Process Explorer lets you take a dump of its current state. Often users also need a crash dump.

We can achieve this via `procdump -t`. e.g. UI could be:
- Go to process explorer
- Right click a process and choose 'enable Crash Dump'
- Then on crash it would go in some well known folder

ahmelsayed:
The dumps go under `D:\home\data\CrashDumps`

### DIFF
--- a/Common/Constants.cs
+++ b/Common/Constants.cs
@@ -90,5 +90,7 @@ namespace Kudu
 
         //Setting for VC++ for node builds
         public const string VCVersion = "2015";
+
+        public const string CrashDumps = "CrashDumps";
     }
 }
--- a/Kudu.Services.Web/App_Start/NinjectServices.cs
+++ b/Kudu.Services.Web/App_Start/NinjectServices.cs
@@ -457,6 +457,7 @@ namespace Kudu.Services.Web.App_Start
             routes.MapHttpProcessesRoute("one-process-thread", "/{processId}/threads/{threadId}", new { controller = "Process", action = "GetThread" }, new { verb = new HttpMethodConstraint("GET") });
             routes.MapHttpProcessesRoute("all-modules", "/{id}/modules", new { controller = "Process", action = "GetAllModules" }, new { verb = new HttpMethodConstraint("GET") });
             routes.MapHttpProcessesRoute("one-process-module", "/{id}/modules/{baseAddress}", new { controller = "Process", action = "GetModule" }, new { verb = new HttpMethodConstraint("GET") });
+            routes.MapHttpProcessesRoute("one-process-dump-on-crash", "/{id}/dump/oncrash", new { controller = "Process", action = "TakeMemoryDumpOnCrash" }, new { verb = new HttpMethodConstraint("POST") });
 
             // Runtime
             routes.MapHttpRouteDual("runtime", "diagnostics/runtime", new { controller = "Runtime", action = "GetRuntimeVersions" }, new { verb = new HttpMethodConstraint("GET") });
--- a/Kudu.Services.Web/Content/Scripts/ProcessExplorer.js
+++ b/Kudu.Services.Web/Content/Scripts/ProcessExplorer.js
@@ -503,6 +503,13 @@ var Process = (function () {
         });
     };
 
+    Process.prototype.enableCrashDump = function () {
+        return $.ajax({
+            url: appRoot + "api/processes/" + this._json.id + "/dump/oncrash",
+            type: "POST"
+        });
+    };
+
     Process.getIdFromHref = function (href) {
         return parseInt(href.substr(href.lastIndexOf("/") + 1));
     };
@@ -863,6 +870,14 @@ function overrideRightClickMenu() {
                     $("li")[0].click();
                     $("li").blur();
                     break;
+                case "crashdump":
+                    $("#proc-loading").show();
+                    process.enableCrashDump().done(function () {
+                        return processExplorerSetupAsync();
+                    }).fail(function () {
+                        return processExplorerSetupAsync();
+                    });
+                    break;
             }
         },
         items: {
@@ -874,6 +889,7 @@ function overrideRightClickMenu() {
                     "dump2": { name: "Full Dump" }
                 }
             },
+            "crashdump": { name: "Enable Crash Dump" },
             "sep1": "---------",
             "properties": { name: "Properties" }
         },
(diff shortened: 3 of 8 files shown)


Kudu.Services/Diagnostics/ProcessController.cs
```diff
@@ -274,6 +276,48 @@ public HttpResponseMessage MiniDump(int id, int dumpType = 0, string format = nu
             }
         }
 
+        [HttpPost]
+        public async Task<HttpResponseMessage> TakeMemoryDumpOnCrash(int id)
+        {
+            using (_tracer.Step("ProcessController.TakeMemoryDumpOnTerminate"))
```
mathewc: Shouldn't this tracer log "OnTerminate" match the method name "OnCrash" ?

Common/Constants.cs
```diff
@@ -90,5 +90,7 @@ public static TimeSpan MaxAllowedExecutionTime
 
         //Setting for VC++ for node builds
         public const string VCVersion = "2015";
+
+        public const string CrashDumps = "CrashDumps";
```
mathewc: Maybe this folder should be just "dumps"? If we decide to write other types of dumps there in the future?

Kudu.Services.Web/App_Start/NinjectServices.cs
```diff
@@ -457,6 +457,7 @@ public static void RegisterRoutes(IKernel kernel, RouteCollection routes)
             routes.MapHttpProcessesRoute("one-process-thread", "/{processId}/threads/{threadId}", new { controller = "Process", action = "GetThread" }, new { verb = new HttpMethodConstraint("GET") });
             routes.MapHttpProcessesRoute("all-modules", "/{id}/modules", new { controller = "Process", action = "GetAllModules" }, new { verb = new HttpMethodConstraint("GET") });
             routes.MapHttpProcessesRoute("one-process-module", "/{id}/modules/{baseAddress}", new { controller = "Process", action = "GetModule" }, new { verb = new HttpMethodConstraint("GET") });
+            routes.MapHttpProcessesRoute("one-process-dump-on-crash", "/{id}/dump/oncrash", new { controller = "Process", action = "TakeMemoryDumpOnCrash" }, new { verb = new HttpMethodConstraint("POST") });
```
mathewc: You have the "dump" route path separate from "oncrash". Is that because you anticipate other dump types for a process? Or could this be "{id}/crashdump"? Also, the some of the routes use {id} while others use {processId}.

Kudu.Services/Diagnostics/ProcessController.cs
```diff
@@ -274,6 +276,48 @@ public HttpResponseMessage MiniDump(int id, int dumpType = 0, string format = nu
...
+                    var process = GetProcessById(id);
+                    var externalCommandFactory = new ExternalCommandFactory(_environment, _settings, _environment.RootPath);
+                    var exe = externalCommandFactory.BuildExternalCommandExecutable(_environment.CrashDumpsPath, _environment.WebRootPath, NullLogger.Instance);
+                    var outputStream = new MemoryStream();
+                    var errorStream = new MemoryStream();
+
+                    var runTask = Task.Run(() => exe.ExecuteAsync(_tracer, $"{ResolveProcDumpPath()} -accepteula -t {id}", outputStream, errorStream, idleManager: new IdleManager(TimeSpan.MaxValue, _tracer)));
+
+                    // procDump can fail to attach to a process if there is another instance of procDump (or any debugger) that's already attached to the same process
+                    await Task.Delay(TimeSpan.FromSeconds(1));
```
mathewc: Where does the 1 second delay come from?

Kudu.Services.Web/Content/Scripts/ProcessExplorer.js
```diff
@@ -863,6 +870,14 @@ function overrideRightClickMenu() {
                     $("li")[0].click();
                     $("li").blur();
                     break;
+                case "crashdump":
+                    $("#proc-loading").show();
+                    process.enableCrashDump().done(function () {
+                        return processExplorerSetupAsync();
+                    }).fail(function () {
+                        return processExplorerSetupAsync();
```
mathewc: Does the user get any visual indication that the crash dump registration failed?

Kudu.Services.Web/App_Start/NinjectServices.cs
ahmelsayed: yes the `{id}` vs `{processId}` part is ... unfortunate. I can have a separate commit that makes them consistent.

I made it that way mostly because we already had a `/dump` route that takes a dump on demand. so I thought I'd add a `/oncrash` to it. Now that I'll be adding all REST operations to it, not sure which route should we go with exactly. Especially that the dump is actually taken on any process termination, not just crash.....

Kudu.Services.Web/Content/Scripts/ProcessExplorer.js
ahmelsayed: No, they'll only see `procdump.exe` not being launched after the processes refresh. I'll look into displaying the error. The code in `ProcessExplorer.js` is pretty old and was hoping I won't have to mess with it too much :)

Kudu.Services/Diagnostics/ProcessController.cs
ahmelsayed: Indeed

Kudu.Services/Diagnostics/ProcessController.cs
ahmelsayed: It was arbitrary since the whole workflow is asynchronous. The way `procdump.exe -t {pid}` works is that it waits for ever until the process `{pid}` terminates, and it takes a dump then. However it can fail when you first launch it due to things like access denied, won't happen in the sandbox, or "debugger already attached", which could happen if the process you're trying to monitor is already being debugged by any other debugger. I kind of just picked 1 second to wait and see if it terminates within that then report the error, otherwise I'll assume everything is fine.

Now that I think about it, I think we may not need to handle any failure case at all here. Just let the tracer write any errors to kudu trace logs and always report success if the process just starts without any unexpected errors. What do you think?

ahmelsayed:
I'll close this PR since the code will change significantly. Will send another one when [this](https://github.com/projectkudu/kudu/issues/1785#issuecomment-154262966) is ready

PR #1791, opened 2015-11-06
Fixes #1786, validate input for trace leve, and auto-correct if data already corrupted

### LINKED ISSUE #1786: Diagnostics setting api should validate the input such as TraceLevel.
In addition, it should attempt to auto-correct the file if contained invalid value.

shrimpy:


Kudu.Core/Settings/DiagnosticsSettingsManager.cs
```diff
@@ -87,6 +93,7 @@ private void SaveSettings(DiagnosticsSettings diagnosticsSettings)
                 FileSystemHelpers.EnsureDirectory(Path.GetDirectoryName(_path));
             }
 
+            diagnosticsSettings.EnsureValid();
```
mathewc: Why does this have to be done after deserialization? Won't the TraceEnum converter validate these settings for us on deserialization? If not, we could have our own converter that does. That way, when we add new settings we don't have to remember to add the setting key name to your static list.

Kudu.Contracts/Settings/DiagnosticsSettings.cs
```diff
@@ -114,5 +124,72 @@ public bool RemoveSetting(string key)
...
+            {
+                object value = this.GetSetting(key);
+                if (!Enum.IsDefined(typeof(TraceEventType), value))
+                {
+                    throw new ArgumentException(string.Format(CultureInfo.InvariantCulture, "Invalid Key: '{0}', Value: '{1}'", key, value));
+                }
+            }
+        }
+
+        public static bool AutoCorrect(ITracer tracer, string rawContent, out string newRawContent)
```
mathewc: This method seems really complicated, and requires maintenance to remember to add new settings to a static list and to add new codeblocks to this method.

Kudu.Core/Settings/DiagnosticsSettingsManager.cs
```diff
@@ -69,6 +69,12 @@ private DiagnosticsSettings ReadSettings()
                 {
                     string fileContent = null;
                     OperationManager.Attempt(() => fileContent = FileSystemHelpers.ReadAllTextFromFile(_path));
+                    if (DiagnosticsSettings.AutoCorrect(_tracer, fileContent, out fileContent))
+                    {
+                        // if there is corrupted value, reset it back to file system
+                        FileSystemHelpers.WriteAllTextToFile(_path, fileContent);
+                    }
+
                     return JsonConvert.DeserializeObject<DiagnosticsSettings>(fileContent);
```
suwatch: We might be over-doing this (given we fix the setter and this would become rare).  I would do the following.  
```CSharp
try
{
    return JsonConvert.DeserializeObject<DiagnosticsSettings>(fileContent);
}
catch (Exception ex)
{
    _tracer.TraceError(ex);
    FileSystemHelpers.SafeDelete(_path);
    return new DiagnosticsSettings();
}
```

Kudu.Core/Settings/DiagnosticsSettingsManager.cs
```diff
@@ -87,6 +93,7 @@ private void SaveSettings(DiagnosticsSettings diagnosticsSettings)
                 FileSystemHelpers.EnsureDirectory(Path.GetDirectoryName(_path));
             }
 
+            diagnosticsSettings.EnsureValid();
             string fileContent = JsonConvert.SerializeObject(diagnosticsSettings);
```
suwatch: Looking at the code, diagnosticsSettings.EnsureValid() is not necessary as JsonConvert.SerializeObject(diagnosticsSettings); and DiagnosticsSettings class member type would have enforce it already.  No?  Try simulate it - let me know if you can get pass this without exception.

suwatch: I also reviewed Azure codes and still couldn't find how this could have happenned from the normal diagnostics api usages.

Kudu.Core/Settings/DiagnosticsSettingsManager.cs
shrimpy: the reason i have auto correction is, there are corrupted data exist in some user`s app. we would like to automatically fix those broken data.

for number value, serialization won`t cache issue.
````
           var result = (TraceEventType)2;
           Console.WriteLine(result);                      // will print "Error"
````

the error we see from controller, the XmlWriter must have a customer XML serializer to make sure value is define in enum

suwatch: Try running `JsonConvert.DeserializeObject<DiagnosticsSettings>(fileContent); ` with corrupted fileContent.  It will throw, no?

Kudu.Core/Settings/DiagnosticsSettingsManager.cs
shrimpy: this is to prevent rubbish value, where user can call UpdateSetting("AzureDriveTraceLevel", 9999)

since we are doing validation, i think will be good to also prevent this kind of invalid value

````
           var result = (TraceEventType)2;
            Console.WriteLine(result);                         // will print Error

            var result2 = (TraceEventType)9999;
            Console.WriteLine(result2);                       // will print 9999
````

Kudu.Core/Settings/DiagnosticsSettingsManager.cs
shrimpy: Nope! that is why i put in the validation.

image

Kudu.Contracts/Settings/DiagnosticsSettings.cs
shrimpy: let me improve a little bit, so that new setting values do not need to add new codeblocks (if they have the same default value)

Kudu.Core/Settings/DiagnosticsSettingsManager.cs
shrimpy: RE: Won't the TraceEnum converter validate these settings for us on deserialization?

[shrimpy] See below "9999" value is also get pass deserialization, which we do not want it to happen. We will need to do a IsDefined check just for safety.

image


RE: If not, we could have our own converter that does. That way, when we add new settings we don't have to remember to add the setting key name to your static list.


[shrimpy] do you mean to have a converter like "StringEnumConverter", and attribute on Property level?

the issue is when one setting value is wrong, we will also want to update its co-related setting values.

e.g if AzureTableTraceLevel is corrupted, we will want to reset AzureTableEnabled as well.
for a customer converter, we will need to enforce the order of deescalation, might a bit over engineering for this issue.

suwatch:
Look good.

PR #1795, opened 2015-11-10
VFS API: return file/directory created time `crtime`

duncansmart:
Note it's not `ctime` as that has different semantics (See http://www.linux-faqs.info/general/difference-between-mtime-ctime-and-atime "A common mistake is that ctime is the file creation time ...")


shrimpy:
@duncansmart 
thanks for your contribution. merged in https://github.com/projectkudu/kudu/commit/d04b61e33456b34656bcd1058361155270fe9807

Issue #634, opened 2013-05-27
Remove HEAD.lock and index.lock where necessary


suwatch:
HEAD.lock is the Git mechanism to synchronize git operations.   The issue is sometime the file is left over and subsequent git operations failed (thinking other git operation is in flight).   Since Kudu has our own lock (deployment lock), we could try to delete this lock in some cases.   One excerpt for discussion (in case of SubModule).

`I tried removed the HEAD.lock (site\repository\.git\modules\PublicSubModule\HEAD.lock) manually and it works fine.`
`The command that failed was “git.exe submodule update --init –recursive”.`

PR #1799, opened 2015-11-12
Fixes #1798, clear leftover branch locks.

### LINKED ISSUE #1798: clear up master.lock
One customer report he failed to push content to his local git repo. after investigate we found the "master.lock" file was there causing the issue.

we have logic go clean other lock file, we may as well as to do the same to clean up master.lock.

shrimpy:


Kudu.Core/SourceControl/Git/GitExeRepository.cs
```diff
@@ -367,19 +367,28 @@ public void ClearLock()
...
+            // Always clear lock file under ".git/refs/heads"
+            var branchLockFiles = Path.Combine(_gitExe.WorkingDirectory, ".git", "refs", "heads");
 
             if (!Directory.Exists(lockFilesPath))
             {
                 return;
             }
 
-            var lockFiles = Directory.EnumerateFiles(lockFilesPath, "*.lock", SearchOption.AllDirectories)
+            List<string> lockFiles = Directory.EnumerateFiles(lockFilesPath, "*.lock", SearchOption.AllDirectories)
```
davidebbo: Why do we need to search the entire .git folder recursively? It could be quite expensive. Don't we know the exact location of the various lock files?

ahmelsayed: Can this use `FileSystemHelpers`?

shrimpy: @davidebbo 

looked at the [original issue](https://github.com/projectkudu/kudu/issues/634), there will be multiple HEAD.lock in different folder (module folder), so we will have to go thru all folder recursively.


@ahmelsayed 
DirectoryBase from IFileSystem doesn`t have EnumerateFiles API, i think we can implement it, but the existing test case is running against real file system, we are fine with testing. i will leave it as it is for now, we can add it later if we really want to mock the file system.

davidebbo: Ah sorry, didn't realize we were already doing that.

davidebbo:
:shipit:

Kudu.Core/SourceControl/Git/GitExeRepository.cs
shrimpy: yep.. i just change "var" to "List<string>", since i will do "AddRange" later, thought this will be easier to read the code when we look back the code later.

Issue #1765, opened 2015-10-23
Unable to run functional tests


sitereactor:
I have been following this guide https://github.com/projectkudu/kudu/wiki/Getting-started in order to get the functional tests to run locally, but have been unsuccessful. All tests continue to fail even though I have been through all the steps in the guide above.

It looks like all the folders are created (ie. `C:\Kudu-Test-Files\KuduApps`), and I can see that the site(s) are created in IIS (kudu_*). But for a test like `AppChangesShouldTriggerBuildTests` I see in an `System.Net.Http.HttpRequestExceptionResponse status code does not indicate success: 401 (Unauthorized).` error, which I assume should not happen. Below is the full stack trace.

Any pointers as to what could be wrong in my setup? (I'm running VS2013 as Admin, and I've tried granting "Network Service" full control over the Kudu.Services.Web folder within the solution).

```
System.AggregateExceptionOne or more errors occurred.
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at System.Threading.Tasks.Task.Wait()
   at Kudu.TestHarness.ApplicationManager.Run(String testName, Action`1 action) in ApplicationManager.cs: line 242
   at Kudu.FunctionalTests.AppChangesShouldTriggerBuildTests.PushAppChangesShouldTriggerBuild() in GitRepositoryManagementTests.cs: line 329
System.AggregateExceptionOne or more errors occurred.
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at Kudu.TestHarness.SitePool.<CreateApplicationInternal>d__c.MoveNext() in SitePool.cs: line 116
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at Kudu.TestHarness.SitePool.<CreateApplicationAsync>d__0.MoveNext() in SitePool.cs: line 45
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

Issue #1638, opened 2015-07-29
Extending the logstream


sitereactor:
I guess this is kind of a feature request, as it doesn't seem possible OOTB. But would it be possible to extend the [logstream](https://github.com/projectkudu/kudu/wiki/Diagnostic-Log-Stream) to look at files outside the /LogFiles folder?
If you use [Umbraco](https://github.com/umbraco/Umbraco-CMS) on Azure WebApps it would be nice if the UmbracoTraceLogs.txt file (which resides in /wwwroot/App_Data/Logs) could be exposed to the /logstream just like /logstream/kudu/deployment, /logstream/application, etc.

The LogManager is restricted to that one LogFiles folder https://github.com/projectkudu/kudu/blob/master/Kudu.Services/Diagnostics/LogStreamManager.cs#L52 which there might be a very good reason for, but would it be possible to add some kind of extension points to allow other files to be "monitored" by the logstream?

davidebbo:
Indeed, it's not something that's supported today. Would you expect the same `/api/logstream` request to stream a mix of content from the LogFiles folder and other folders? Or separate `/api/logstream` calls that each ask for a different thing?

I guess the question is whether this configuration should exist on the URL or should be something pre-configured via some kind of site setting.

sitereactor:
I'd prefer it to be application specific, so you could do something like `/api/logstream/umbraco` - if other CMS applications have their logs written to the filesystem it would make sense if you could do something like `/api/logstream/orchard` or `/api/logstream/drupal` and it would know the path for that application or you could configure your application in an easy way (thinking you could specify endpoint and filesystem path - so for Umbraco it would be `/api/logstream/umbraco` and path `/site/wwwroot/App_Data/logs/UmbracoTraceLogs.txt`). 
I think that would tie in nicely to the rest of the diagnostics since the application specific log(s) might have different and valuable information that you don't get through IIS logs, deployment logs, trace logs, etc.

Makes sense?

warrenbuckley:
Hi @davidebbo I brought this up with @sitereactor as I was wanting to build a little NodeJS CLI utility for their Umbraco platform/service. With this it would allow more context specific logs to the Umbraco application.

And like @sitereactor approach suggests it would be a nice solution so any folder could be added to the fileWatcher that contains log files .txt etc & any other services using Kudu with their application could see more specific & contextual logs/errors

davidebbo:
[...]

PR #1800, opened 2015-11-13
Initial attempt of Extending the logstream #1638

sitereactor:
This PR implements the idea of extending the logstream as discussed in https://github.com/projectkudu/kudu/issues/1638

With this in place I can add an AppSetting with Key `SCM_LOGSTREAM_UMBRACO` and Value `site\wwwroot\app_data\logs`, which enables the endpoint `/api/logstream/umbraco` and a logstream that outputs everything from the logfiles within the `logs` folder.

I still have issues getting the functional tests to run as per #1765 so I haven't added any functional tests (yet). Out of the 3 `LogStreamManager` tests only the "NotFound" one works on my machine. I'd be happy to add a test for this scenario, but I'd need a bit of guidance on what to include and how to specify setting keys as part of a test.

### DIFF
--- a/Kudu.Contracts/Settings/SettingsKeys.cs
+++ b/Kudu.Contracts/Settings/SettingsKeys.cs
@@ -13,6 +13,7 @@
         public const string TraceLevel = "SCM_TRACE_LEVEL";
         public const string CommandIdleTimeout = "SCM_COMMAND_IDLE_TIMEOUT";
         public const string LogStreamTimeout = "SCM_LOGSTREAM_TIMEOUT";
+        public const string LogStreamCustomLog = "SCM_LOGSTREAM_";
         public const string GitUsername = "SCM_GIT_USERNAME";
         public const string GitEmail = "SCM_GIT_EMAIL";
         public const string ScmType = "ScmType";
@@ -35,9 +36,5 @@
         public const string SiteExtensionsFeedUrl = "SCM_SITEEXTENSIONS_FEED_URL";
         public const string DisableDeploymentOnPush = "SCM_DISABLE_DEPLOY_ON_PUSH";
         public const string TouchWebConfigAfterDeployment = "SCM_TOUCH_WEBCONFIG_AFTER_DEPLOYMENT";
-        public static string CustomLogStream(string streamKey)
-        {
-            return string.Concat("SCM_LOGSTREAM_", streamKey.ToUpperInvariant());
-        }
     }
 }
--- a/Kudu.Contracts/Settings/SettingsKeys.cs
+++ b/Kudu.Contracts/Settings/SettingsKeys.cs
@@ -35,5 +35,9 @@
         public const string SiteExtensionsFeedUrl = "SCM_SITEEXTENSIONS_FEED_URL";
         public const string DisableDeploymentOnPush = "SCM_DISABLE_DEPLOY_ON_PUSH";
         public const string TouchWebConfigAfterDeployment = "SCM_TOUCH_WEBCONFIG_AFTER_DEPLOYMENT";
+        public static string CustomLogStream(string streamKey)
+        {
+            return string.Concat("SCM_LOGSTREAM_", streamKey.ToUpperInvariant());
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


Kudu.Contracts/Settings/SettingsKeys.cs
```diff
@@ -35,5 +35,9 @@ public static class SettingsKeys
         public const string SiteExtensionsFeedUrl = "SCM_SITEEXTENSIONS_FEED_URL";
         public const string DisableDeploymentOnPush = "SCM_DISABLE_DEPLOY_ON_PUSH";
         public const string TouchWebConfigAfterDeployment = "SCM_TOUCH_WEBCONFIG_AFTER_DEPLOYMENT";
+        public static string CustomLogStream(string streamKey)
```
suwatch: We usually don't do it here.  Instead, we put all setting helper in [extensions](https://github.com/projectkudu/kudu/blob/master/Kudu.Contracts/Settings/DeploymentSettingsExtension.cs).

Kudu.Services/Diagnostics/LogStreamManager.cs
```diff
@@ -279,21 +282,29 @@ private string ParseRequest(HttpContext context)
...
+            bool isHttp = String.Equals(firstPath, "http", StringComparison.OrdinalIgnoreCase);
+            if (isHttp || isApplication)
+            {
                 FileSystemHelpers.EnsureDirectory(Path.Combine(_logPath, firstPath));
+                return Path.Combine(_logPath, routePath);
             }
-            else
+
+            // in case a custom log file path exist
+            string path = _settings.GetValue(SettingsKeys.CustomLogStream(firstPath));
```
suwatch: It would be great to add comments what is the custom path example.   For instance, it could be relative to root such as `foo`, and the outcome will be `d:\home\foo`.  It could be absolute value - such as `d:\home\bar`.

Kudu.Services/Diagnostics/LogStreamManager.cs
```diff
@@ -279,21 +282,29 @@ private string ParseRequest(HttpContext context)
...
+            if (!string.IsNullOrEmpty(path))
             {
-                bool isHttp = String.Equals(firstPath, "http", StringComparison.OrdinalIgnoreCase);
-                if (isHttp)
-                {
-                    FileSystemHelpers.EnsureDirectory(Path.Combine(_logPath, firstPath));
-                }
+                var customLogPath = Path.Combine(_environment.RootPath, path.TrimStart('\\'));
```
suwatch: It should take the remaining portions of routePath into account, no?

suwatch:
It would be great to write either a unit or e2e test for this.   To write a unit test, you may want to factor out the path resolution to a function and test that function in unit test fashion.

Kudu.Contracts/Settings/SettingsKeys.cs
sitereactor: Alright I'll move it to extensions. Thanks!

Kudu.Services/Diagnostics/LogStreamManager.cs
sitereactor: Yes, iI'll a note. It's relative to the root, so you'd enter `site\wwwroot\foo` to have it listen to a folder within the website.

Kudu.Services/Diagnostics/LogStreamManager.cs
sitereactor: I only use the routepath to identify the correct settings key. It's not used for the path to the log files folder. As it could be anywhere within the site I think it makes sense that the "enitre" relative path comes from the setting.

sitereactor: Well, the first part of the routepath that is.

sitereactor:
@suwatch I have updated the pull request according to your initial comments. I'm still missing the unit test or end to end test, and I was hoping you could guide me a bit here. 
If I write a unit test for the part of the code that determines which log to stream (trace by default, application if that is provided in the Url or a custom log if both url part + settings key is present) - do you want it separated out as a class or maybe as a public or internal method in the `LogStreamMananger` ?

sitereactor:
oh, and btw I signed the CLA but not sure why it has been updated to cla-signed. Thought that happened automatically when I signed it.

warrenbuckley:
@suwatch @davidebbo any update on this PR please?
As it seems that @sitereactor is awaiting a response from you to help move this along?

Thanks,
Warren 😄

davidebbo:
@sitereactor I don't remember where we left this. Is it in a state where you think it's ready, or were there more things to solve first?

mi-hol:
@sitereactor @davidebbo this PR seems orphaned

davidebbo:
Yes, closing it for now.

PR #1808, opened 2015-11-20
Fix logo height issue with edge 13

shibayan:
Issue with Edge 13 (Windows 10 TH2) And I think the Edge of the problem...

kudu_logo


davidebbo:
Yes, I see the issue as well with my TH2 box.

Have you tested that things still look the same in Chrome/IE with the fix? If so, it should be good to go.

shibayan:
Yes. I deployed private kudu on Web Apps and confirmed the display in IE / Chrome.

kudu_chrome

davidebbo:
Thanks!

davidebbo:
The change is now live in Azure.

PR #1811, opened 2015-11-20
Update logic for detecting ASP.NET 5 web applications for RC1 release

### LINKED ISSUE #1810: ASP.NET 5 RC1 breaks logic for determining web application projects
[AspNet5Helper](https://github.com/projectkudu/kudu/blob/a64937bdf4b5d329a16a6eebb97ddb7838b9e2df/Kudu.Core/Infrastructure/AspNet5Helper.cs#L14) determines whether an ASP.NET 5 project is a web application by the presence of `webroot` in project.json. However RC1 moves `webroot` to hosting.json ([see here](https://github.com/aspnet/Announcements/issues/94)), hence this logic is broken for projects targeting RC1. This causes deployment of RC1 projects to fail.

ahmelsayed:
Fixes #1810


davidebbo:
This won't break beta8 and earlier, right?

ahmelsayed:
I don't think so. As long as there is a `wwwroot` folder, which has been there for a while. The only time it'll break beta8, I think if someone changed the name of wwwroot in their project.json, but according to them, that was very rare for people to do that.

PR #1816, opened 2015-11-27
Replaced gif spinner with old school ASCII spinner

snobu:
Well... it's a __Debug__ Console so i thought an ajax gif loader wasn't really cutting it.
Introducing the old school ASCII spinner!

spinner

//TO DO: Stop execution while div is hidden.

### DIFF
--- a/Kudu.Services.Web/DebugConsole/Default.cshtml
+++ b/Kudu.Services.Web/DebugConsole/Default.cshtml
@@ -37,7 +37,7 @@
                     <button class="btn btn-info box-border" style="display:none" data-bind="visible: isTransferInProgress(), click: showCopyProgressModal">
                         <span id="copy-percentage" title="percentage copy"></span>
                     </button>
-                    <i data-bind="visible: koprocessing()" class="fa fa-spinner fa-spin" title="Please wait"></i>
+                    <i data-bind="visible: koprocessing()" class="old-school-spinner" title="Please wait"></i>
                     <i data-bind="visible: errorText, attr: {title: errorText}" class="alert-warning glyphicon glyphicon-exclamation-sign" style="display: none"></i>
                 </div>
             </div>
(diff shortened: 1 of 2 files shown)


Kudu.Services.Web/DebugConsole/Default.cshtml
```diff
@@ -184,3 +184,28 @@
...
+    frames.forEach(function(frame) {
+        setTimeout(function() {
+            spinbox.text(frame);
+        }, timeout * i);
+        i++;
+    });
+
+    setTimeout(function() {
+        spinner();
+    }, timeout * (frames.length));
```
shrimpy: why not replace "timeout * (frames.length)" with a actual number, since it won`t change.

Kudu.Services.Web/DebugConsole/Default.cshtml
```diff
@@ -184,3 +184,28 @@
 <script src="~/content/scripts/jquery-console/jquery.console.js"></script>
 <script src="~/content/scripts/ace-init.js"></script>
 <script src="~/content/scripts/HelperScript.js"></script>
+
+<script>
+// Old school ASCII spinner
+var i = 0;
```
shrimpy: i think this "var i" is unnecessary?

shrimpy:
@snobu what is the intention of replacing the GIF with ASCII spinner? constantly access DOM tree doesn`t seem good for perf.

snobu:
If i'm not the only one who thinks the ol' spinner thing is a good idea, two options unfold:

1. I can make the JS only trigger when div is visible.
2. I can make it a gif and remove all JS code

shrimpy:
I will vote for option 2 make it a GIF or use CSS animation.

suwatch:
`i thought an ajax gif loader wasnt really cutting it.`

Could you elaborate pros and cons?  Any imminent issue to require this fix?  Just to get the idea as I am no expert.

snobu:
I think that sounded more serious that it should have. Just trying to make that spinner stand out a bit, we've all seen rotating gif wheels on and on and on and we can break through the monotony with a circa 1982 text spinner :)

I'll try to do it the CSS3 way and if that fails, then gif.

snobu:
CSS3 works but, as expected, every browser has a mind of its own and injects a 1-2 second delay before the animation starts, which renders moot the whole point of having a wait indicator for brief moments.

For the sake of completion here's the fiddle:
http://jsfiddle.net/d8bd15u4/6/

I'll make a gif. 1987 technology never ceases to impress me.

snobu:
How about this
kpacman2

instead of `fa-spinner`?
fa

davidebbo:
Love it :)

snobu:
Continues in https://github.com/projectkudu/kudu/pull/1848

PR #1822, opened 2015-12-02
[Ace] Version up to 1.2.2 and code cleanup

snobu:
### [Ace] Version up to 1.2.2 and code cleanup ###

`editor.setValue()` now expects a string - changed that accordingly.

Removed `var statusbarObj = Object.create(statusbar)` - no need to newup a statusbarObj, we can use `statusbar` directly.

#### What's new in Ace 1.2.2 that at least two people care about: ####

- Updated LESS and CSS modes
- Vim mode enhancements
- Fix CTRL-F now binds to `Find` - no longer cycles between `Find` and `Find and Replace` - See https://github.com/ajaxorg/ace/issues/2653. I already know two people who care about this one!

Passes my PhantomJS functional tests - which given that nobody else knows what they are - isn't saying much :) I'll make them public soon.

#### Question for the Kudu audience: ####
Should i implement a CTRL-R as "Reload this file in the editor"? Might do wonders for log files that otherwise you'd need to tail with `tail -f php_error.log`.

### DIFF
--- a/Kudu.Services.Web/Content/Scripts/FileBrowser.js
+++ b/Kudu.Services.Web/Content/Scripts/FileBrowser.js
@@ -145,7 +145,7 @@ var copyObjectsManager = {
         }
 
         return perc;
-    }, 
+    },
     removeAtIndex: function(index) {
         delete this._copyProgressObjects[index];
     },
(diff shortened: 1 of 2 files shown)


shrimpy:
CTRL+R sounds like a good feature, +1 for it.

Issue #1705, opened 2015-09-24
Upload button


sansjunk:
Kudu is great for drag 'n drop upload but blame me for trying Kudu on Edge which doesn't support drag 'n drop yet! Can we please have an explicit old-style upload button for those who dare to try the bleeding edge :-)

davidebbo:
To be honest, I'm a bit hesitant to do work for something that is (hopefully!) a temporary limitation until they complete implementation.

sansjunk:
Fair enough!

christopheranderson:
Poking this issue. Drag and drop still isn't working on Edge for large zip files.

davidebbo:
@christopheranderson and the same file works in Chrome? It probably comes down to browser limitation. Frankly, I'm hesitant to invest in adding an upload button when there is an easy workaround with Chrome. And Edge did get better in that area through the years, so hopefully it'll continue to improve.

jvano:
Hi

Kudu will continue to run in Azure App Service. However, this repo will no longer be maintained.
If the problem persists and is related to running on Azure App Service, please open a support incident in Azure:
https://learn.microsoft.com/en-us/azure/azure-portal/supportability/how-to-create-azure-support-request

This way we can better track and assist you on this case

Thanks,

Joaquin Vano
Azure App Service

PR #1835, opened 2015-12-15
Add DataTransferItem feature detection for Edge

shibayan:
Fixed a problem that can't upload files using Microsoft Edge #1705 
I confirmed in IE11/Edge25/Chrome47 browsers :smile_cat:


ahmelsayed:
what about firefox? Also I don't quite understand why a check for `getAsEntry` is needed here? It just seems to make the upload case more restrictive right?

ahmelsayed:
Never mind I checked the code and I understand what's going on now. It looks good to me, I'll merge it today. Thanks @shibayan !

shibayan:
Thanks! :tada:

PR #1851, opened 2016-01-04
Return 200 instead of 202 when running WebJobs from ARM

davidebbo:
See comment in code

### DIFF
--- a/Kudu.Services/Jobs/JobsController.cs
+++ b/Kudu.Services/Jobs/JobsController.cs
@@ -179,7 +179,10 @@ namespace Kudu.Services.Jobs
             try
             {
                 _triggeredJobsManager.InvokeTriggeredJob(jobName, arguments, "External - " + Request.Headers.UserAgent);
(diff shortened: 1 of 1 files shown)


suwatch:
:shipit:

mathewc:
It'd be best if we had a regression test for this in place - that sends a mocked request that looks like an ARM request and verifies the behavior. Is there already a test suite that you could add one to?

davidebbo:
Afaik, we don't currently have any tests that exercise the ARM code path. Maybe others know otherwise?

mathewc:
There is a `JobsControllerTests` test suite that is already doing some unit tests mocking requests and verifying behavior. It should be easy to add a new test that sends a request with/without the ARM header and verifies the response code. We have some other tests that mock ARM requests, e.g `SiteExtensionInstallUninstallAsyncTest`

davidebbo:
Added ARM test: 0c73e75d3228b923a96511ed002799e92783f069

mathewc:
Awesome, thanks :)

PR #1852, opened 2016-01-06
Update post deployment env var to be 'SCM_POST_DEPLOYMENT_ACTIONS_PATH', and cleanup unused script.

shrimpy:
From wiki, "SCM_POST_DEPLOYMENT_ACTIONS_PATH" is the right environment variable to override default post deployment script folder path, while "POST_DEPLOYMENT_ACTIONS_DIR" is an internal environment variable to be used in deployment script.


davidebbo:
:shipit:

PR #1857, opened 2016-01-08
fix #1853 Caching the last /api/deployments

### LINKED ISSUE #1853: Caching the last /api/deployments and its etag to improve perf
Given the unchanged nature of deployment list, caching the last result of `/api/deployments` and its Etag would improve performance.   This reduced the number of file read in normal scenario to one.  We could also trace how long does it take to calculate etag (reading a file) for debugging purpose.

suwatch:


Kudu.Services/Deployment/DeploymentController.cs
```diff
@@ -296,7 +305,18 @@ public HttpResponseMessage GetDeployResults()
             {
                 using (_tracer.Step("DeploymentService.GetDeployResults"))
                 {
-                    IEnumerable<DeployResult> results = GetResults(Request).ToList();
+                    if (!currentEtag.Equals(cachedDeployments.Etag))
```
davidebbo: `EtagEquals` like above?

davidebbo: Oh, I see that one takes a Request. Ignore previous comment...

Kudu.Services/Deployment/DeploymentController.cs
```diff
@@ -296,7 +305,18 @@ public HttpResponseMessage GetDeployResults()
...
+                        cachedDeployments = new DeploymentsCacheItem
+                        {
+                            Results = GetResults(Request).ToList(),
+                            Etag = currentEtag
+                        };
+
+                        _cachedDeployments = cachedDeployments;
+                    }
+
+                    IEnumerable<DeployResult> results = cachedDeployments.Results;
```
davidebbo: Somewhat pointless variable. Probably just inline in next statement.

davidebbo:
Minor comment. :shipit:

Kudu.Services/Deployment/DeploymentController.cs
suwatch: Will do.

PR #1870, opened 2016-01-26
fix #1867 handle disk full issue

### LINKED ISSUE #1867: Improve certain GET apis to handle no free space or read-only.
For example, /api/sshkeys or /api/deployments failed miserably when no more available space (hitting quota) or disk is read-only (standby volume).  We should handle that gracefully.

suwatch:


Kudu.Core/Environment.cs
```diff
@@ -270,5 +273,40 @@ public static bool IsAzureEnvironment()
...
+        {
+            try
+            {
+                ulong freeBytes;
+                ulong totalBytes;
+                GetDiskFreeSpace(path, out freeBytes, out totalBytes);
+
+                var usage = Math.Round(((totalBytes - freeBytes) * 100.0) / totalBytes);
+                var color = usage > 97 ? "red" : (usage > 90 ? "orange" : "green");
+                return String.Format("<span style='color:{0}'>{1:#,##0} MB total; {2:#,##0} MB free</span>", color, totalBytes / (1024 * 1024), freeBytes / (1024 * 1024));
```
shrimpy: for safety, would be good to change String.Format to String.Format(***CultureInfo.InvariantCulture***...

shrimpy:
:shipit:

Kudu.Services.Web/Env.cshtml
```diff
@@ -1,95 +1,103 @@
...
+        <li>CLR version: @Environment.Version</li>
+        <li>System directory: @Environment.SystemDirectory</li>
+        <li>Current working directory: @Environment.CurrentDirectory</li>
+        <li>IIS command line: @Environment.CommandLine</li>
+        @{
+            var homePath = Environment.GetEnvironmentVariable("HOME");
+            <li>@homePath usage: @Html.Raw(Kudu.Core.Environment.GetFreeSpaceHtml(homePath))</li>
+            if (Kudu.Core.Environment.IsAzureEnvironment())
+            {
+                <li>d:\local usage: @Html.Raw(@Kudu.Core.Environment.GetFreeSpaceHtml("d:\\local"))</li>
```
shrimpy: do we have environment variable for "d:\\local"?

Kudu.Core/Environment.cs
suwatch: alright.

Kudu.Services.Web/Env.cshtml
suwatch: Not exactly d:\local.   I think it is ok to hardcode given we check for Azure env.   I will robust the codes to check for Directory.Exists in addition.

PR #1875, opened 2016-02-01
support functions sync triggers

suwatch:


Common/Constants.cs
```diff
@@ -92,5 +92,8 @@ public static TimeSpan MaxAllowedExecutionTime
 
         //Setting for VC++ for node builds
         public const string VCVersion = "2015";
+
+        public const string X_MS_SITE_RESTRICTED_JWT = "X-MS-SITE-RESTRICTED-JWT";
+        public const string HTTP_HOST = "HTTP_HOST";
```
shrimpy: all our constants are CamelCase, shall we update these two new contants to CamelCase?

suwatch: Let me fix that.

Kudu.Core/Infrastructure/OperationClient.cs
```diff
@@ -0,0 +1,52 @@
...
+            return await PostAsync<string>(path);
+        }
+
+        public async Task<HttpResponseMessage> PostAsync<T>(string path, T content = default(T))
+        {
+            var jwt = System.Environment.GetEnvironmentVariable(Constants.X_MS_SITE_RESTRICTED_JWT);
+            var host = System.Environment.GetEnvironmentVariable(Constants.HTTP_HOST);
+            using (_tracer.Step("POST " + path))
+            {
+                using (var client = new HttpClient())
```
shrimpy: i think this will send request just fine against PROD server. but for private stamp, we will get SSL error. shall we ignore SSL error by default?

````
using (var handler = new WebRequestHandler())
{
    handler.ServerCertificateValidationCallback += (sender, cert, chain, sslPolicyErrors) => true;
    ....
}
````

ahmelsayed: I don't that's a good idea.

suwatch: Good point both of you.  Let me think of something...

suwatch: I think it is less work and more secure for private azure deployment to get the proper issued certificate.

ahmelsayed: You can also always install the private cert in the cert store of the machine running the tests to make it trusted.

shrimpy:
:shipit:

PR #1880, opened 2016-02-02
caching list webjobs call fixes #1855

### LINKED ISSUE #1855: Call to Api/WebJobs/ takes more than 20 seconds sometimes leading to the GeoMaster timing out and Ibiza WebJobs blade failing to load, need to improve performance

shobak101:


Kudu.Core/Jobs/TriggeredJobsScheduler.cs
```diff
@@ -94,6 +94,7 @@ private void OnJobChanged(string jobName)
                 triggeredJobSchedule.Dispose();
                 _triggeredJobsSchedules.Remove(jobName);
             }
+            ((TriggeredJobsManager) _triggeredJobsManager).ListJobRefreshNeeded = true;
```
davidebbo: Won't that break any unit test that mocks the interface?

Kudu.Core/Jobs/ContinuousJobsManager.cs
```diff
@@ -18,6 +18,19 @@ public class ContinuousJobsManager : JobsManagerBase<ContinuousJob>, IContinuous
     {
         private const string StatusFilesSearchPattern = ContinuousJobStatus.FileNamePrefix + "*";
         private const string Localhost = "127.0.0.1";
+        public bool ListJobRefreshNeeded
```
davidebbo: Use auto-prop

Kudu.Core/Jobs/ContinuousJobsManager.cs
```diff
@@ -18,6 +18,19 @@ public class ContinuousJobsManager : JobsManagerBase<ContinuousJob>, IContinuous
...
+            get
+            {
+                return _listJobRefreshNeeded;
+            }
+            set
+            {
+                _listJobRefreshNeeded = value;
+            }
+        }
+        private bool _listJobRefreshNeeded = true;
```
ahmelsayed: What is this used for? and why have a backing field?

Kudu.Core/Jobs/ContinuousJobsManager.cs
davidebbo: Also, why not just set the cache to null when we need a refresh and do away with this separate prop?

Kudu.Core/Jobs/TriggeredJobsManager.cs
```diff
@@ -31,7 +44,14 @@ public TriggeredJobsManager(ITraceFactory traceFactory, IEnvironment environment
 
         public override IEnumerable<TriggeredJob> ListJobs()
         {
-            return ListJobsInternal();
+            if (TriggeredJobCache == null || ListJobRefreshNeeded)
```
suwatch: var triggerJobCache = TriggerJobCache;
...
return triggerJobCache;

Kudu.Core/Jobs/TriggeredJobsScheduler.cs
suwatch: this could be just set cache to null.

suwatch: The cache object should simply be static class member, no?

PR #1881, opened 2016-02-02
caching web jobs list fixes #1855

### LINKED ISSUE #1855: Call to Api/WebJobs/ takes more than 20 seconds sometimes leading to the GeoMaster timing out and Ibiza WebJobs blade failing to load, need to improve performance

shobak101:
sorry had some merge conflicts, created new branch.

### DIFF
--- a/Kudu.Core/Jobs/ContinuousJobsManager.cs
+++ b/Kudu.Core/Jobs/ContinuousJobsManager.cs
@@ -14,7 +14,7 @@ using Kudu.Contracts.Tracing;
 
 namespace Kudu.Core.Jobs
 {
-    public class ContinuousJobsManager : JobsManagerBase<ContinuousJob>, IContinuousJobsManager, IDisposable
+    public class ContinuousJobsManager : JobsManagerBase<ContinuousJob>, IContinuousJobsManager
     {
         private const string StatusFilesSearchPattern = ContinuousJobStatus.FileNamePrefix + "*";
         private const string Localhost = "127.0.0.1";
@@ -23,12 +23,10 @@ namespace Kudu.Core.Jobs
 
         private readonly Dictionary<string, ContinuousJobRunner> _continuousJobRunners = new Dictionary<string, ContinuousJobRunner>(StringComparer.OrdinalIgnoreCase);
 
-        private JobsFileWatcher _jobsFileWatcher;
-
         public ContinuousJobsManager(ITraceFactory traceFactory, IEnvironment environment, IDeploymentSettingsManager settings, IAnalytics analytics)
             : base(traceFactory, environment, settings, analytics, Constants.ContinuousPath)
         {
-            _jobsFileWatcher = new JobsFileWatcher(JobsBinariesPath, OnJobChanged, null, ListJobNames, traceFactory, analytics);
+            RegisterExtraEventHandlerForFileChange(OnJobChanged);
         }
 
         public override ContinuousJob GetJob(string jobName)
@@ -110,7 +108,7 @@ namespace Kudu.Core.Jobs
 
         protected override void OnShutdown()
         {
-            _jobsFileWatcher.Stop();
+            JobsFileWatcher.Stop();
 
             foreach (ContinuousJobRunner continuousJobRunner in _continuousJobRunners.Values)
             {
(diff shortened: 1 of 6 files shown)


Kudu.Core/Jobs/TriggeredJobsManager.cs
```diff
@@ -31,7 +32,18 @@ public TriggeredJobsManager(ITraceFactory traceFactory, IEnvironment environment
...
+            var cache = TriggeredJobCache;
+            if (cache == null)
+            {
+                cache = ListJobsInternal();
+                TriggeredJobCache = cache;
+            }
+            return cache;
+        }
+
+        public void InvalidateCache ()
```
suwatch: not needed, right?

shobak101: yes, removed

Kudu.Core/Jobs/ContinuousJobsManager.cs
```diff
@@ -32,7 +33,13 @@ public ContinuousJobsManager(ITraceFactory traceFactory, IEnvironment environmen
...
         public override IEnumerable<ContinuousJob> ListJobs()
         {
-            return ListJobsInternal();
+            var cache = ContinuousJobCache;
+            if (cache == null)
+            {
+                cache = ListJobsInternal();
+                ContinuousJobCache = cache;
+            }
+            return cache;
```
ahmelsayed: I think you can get rid of the temp variable `cache` here. Same in the triggered

ahmelsayed: oh I take that back. That's to handle race conditions with the change notification isn't it.

Kudu.Core/Jobs/TriggeredJobsManager.cs
```diff
@@ -31,7 +32,18 @@ public TriggeredJobsManager(ITraceFactory traceFactory, IEnvironment environment
...
         {
-            return ListJobsInternal();
+            var cache = TriggeredJobCache;
+            if (cache == null)
+            {
+                cache = ListJobsInternal();
+                TriggeredJobCache = cache;
+            }
+            return cache;
+        }
```
davidebbo: This method looks identical to the Continuous one. Shouldn't it just be in the base as an `IEnumerable<TJob>` to avoid dup?

Kudu.Core/Jobs/ContinuousJobsManager.cs
shobak101: yes :)

Kudu.Core/Jobs/TriggeredJobsManager.cs
shobak101: the base declared this method as abstract and it had some implementation before, I am not sure why it was, I will think about it sometime and see if it makes sense to change it.

shobak101: the watchers for both continuous and triggered jobs are not symmetric in implementation, for continuous it is in the same class but for the triggered it is in the scheduler. that led to the necessity of having the cache object as an internal static so that the scheduler watcher can null it when needed, such a thing can not work if it was pulled in a common class, so I think this should stay as is.

suwatch: I think what you could do is to move this logic to base class ListJobsInternal().   Then the cache object (static member) would also be in the base class.  The challenge is it might have to be declared as object and, when used, casted accordingly.  Make sense?

Kudu.Core/Jobs/JobsManagerBase.cs
```diff
@@ -77,9 +83,49 @@ protected JobsManagerBase(ITraceFactory traceFactory, IEnvironment environment,
...
+        /// Process job file changed and invoke any extra file watcher event handlers
+        /// </summary>
+        private void OnJobChanged(string jobName)
+        {
+            JobListCache = null;
+            if (FileWatcherExtraEventHandlers != null)
+            {
+                foreach (Action<string> action in FileWatcherExtraEventHandlers)
+                {
+                    action.Invoke(jobName);
```
suwatch: Instead of EventHandlers/Callback model, I wonder if we can make OnJobChanged protected virtual - derived class just override it?

shobak101: the reason for this is that the scheduler is the one that needs an event handler to be called not one of the derived manager classes, and the logic in there can not be just moved to triggeredJobManager because it is specific to the scheduler

shobak101:
opened another pull request to avoid unnecessary merges

PR #1888, opened 2016-02-11
Add initial lfs support and set PATH process level

davidebbo:


davidebbo:
Also, part of this is:

- name the remote origin instead of external (no one remembers reason)
- Don't delete remote. This is needed for lfs to work

davidebbo:
Reviewed by @suwatch

PR #1889, opened 2016-02-12
allow knob to skip SSL cert validation

suwatch:


Setup/DeployPrivateKudu.cmd
```diff
@@ -1,48 +1,58 @@
-@if "%_echo%" neq "1" @echo off
```
davidebbo: What's the deal with this file? Accidental?

Kudu.Core/Infrastructure/OperationClient.cs
```diff
@@ -48,5 +50,13 @@ public OperationClient(ITracer tracer)
                 }
             }
         }
+
+        public static void SkipSslValidation()
```
davidebbo: Maybe change the method name to convey that it doesn't always disable it. Like `SkipSslValidationIfNeeded`. Or something less ugly :)

Kudu.Core/Infrastructure/OperationClient.cs
```diff
@@ -48,5 +50,13 @@ public OperationClient(ITracer tracer)
...
+        {
+            if (System.Environment.GetEnvironmentVariable(SettingsKeys.SkipSslValidation) == "1")
+            {
+                ServicePointManager.ServerCertificateValidationCallback = delegate { return true; };
```
davidebbo: That's cool, didn't know you could omit all the params in the delegate!

Kudu.Core/Infrastructure/OperationClient.cs
suwatch: No preference on name.   I think what you proposed `SkipSslValidationIfNeeded` is fine.   Will fix it.

Setup/DeployPrivateKudu.cmd
suwatch: It is not related to this Ssl validation.  I will separate to another commit.   

As for what it was about, this is the script helping installing private kudu to a site.  Currently we look for curl.exe at `%ProgramFiles(x86)%\git\bin\curl.exe`, and if not exists, `%ProgramW6432%\git\mingw64\bin\curl.exe`.  I added one extra location `%ProgramFiles(x86)%\git\mingw32\bin\curl.exe`.   This would support people installing Git either 32bits or 64bits favors.

davidebbo: But why is the diff showing every line? It's usually the sign of some line ending issue.

Issue #1891, opened 2016-02-12
Sync Triggers needs to respect Disabled functions


mathewc:
Function metadata also supports a "disabled" attribute as follows:

``` javascript
{
  "disabled": true,
  "bindings": {
    "input": [
      {
        "type": "timerTrigger",
        "schedule": "05:00:00"
      }
    ]
  }
}
```

The function runtime does not run a disabled function (e.g. it won't be polling the queue, running the timer, etc.). Details on format can be found in the wiki here: https://github.com/Azure/azure-webjobs-sdk-script/wiki/function.json.

davidebbo:
It'd be so much easier if there was a way to do this from the Functions app instead of Kudu.

mathewc:
I don't understand your comment David. I created this issue because currently the Kudu API is the one choosing which functions to report back with and it needs to take this into account. The Functions app portal will have a toggle button mapping to this setting.

davidebbo:
I meant that ideally we'd find a way for this logic to exists in the functions app instead. Right now, we're in a bit of a strange situation as Functions effectively have a split management API, some of it in Kudu and some of it in Functions app.

jvano:
Hi

Kudu will continue to run in Azure App Service. However, this repo will no longer be maintained.
If the problem persists and is related to running on Azure App Service, please open a support incident in Azure:
https://learn.microsoft.com/en-us/azure/azure-portal/supportability/how-to-create-azure-support-request

This way we can better track and assist you on this case

Thanks,

Joaquin Vano
Azure App Service

PR #1892, opened 2016-02-13
Add REST APIs for managing functions

ahmelsayed:


Kudu.Contracts/Functions/IFunctionManager.cs
```diff
@@ -2,11 +2,20 @@
...
 using Kudu.Core.SourceControl;
+using Kudu.Contracts.Jobs;
+using Newtonsoft.Json.Linq;
 
 namespace Kudu.Core.Functions
 {
     public interface IFunctionManager
     {
         Task SyncTriggers();
+        Task<FunctionEnvelope> CreateOrUpdate(string name, FunctionEnvelope functionEnvelope);
```
mathewc: Nit: Consider using the "Async" naming pattern for all the async functilons (e.g. SyncTriggersAsync, etc.)

Kudu.Contracts/Jobs/FunctionEnvelope.cs
```diff
@@ -0,0 +1,37 @@
+﻿using Kudu.Contracts.Infrastructure;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using System;
+
+namespace Kudu.Contracts.Jobs
+{
+    public class FunctionEnvelope : INamedObject
```
mathewc: Unless "Envelope" is a naming that is already established for other types, you might consider "FunctionDescriptor" instead.

Kudu.Contracts/Jobs/FunctionTemplate.cs
```diff
@@ -0,0 +1,29 @@
...
+    public class FunctionTemplate
+    {
+        [JsonProperty(PropertyName = "id")]
+        public string Id { get; set; }
+
+        [JsonProperty(PropertyName = "trigger")]
+        public string Trigger { get; set; }
+
+        [JsonProperty(PropertyName = "inputs")]
+        public IEnumerable<FunctionTemplateInput> Inputs { get; set; }
```
mathewc: I guess none of our templates currently use Outputs?

Kudu.Contracts/Functions/IFunctionManager.cs
ahmelsayed: I'd like to live in a world where all functions are a sync by default :)

Kudu.Contracts/Jobs/FunctionTemplate.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Kudu.Contracts.Jobs
+{
+    public class FunctionTemplate
+    {
+        [JsonProperty(PropertyName = "id")]
+        public string Id { get; set; }
+
+        [JsonProperty(PropertyName = "trigger")]
+        public string Trigger { get; set; }
```
mathewc: The way the runtime handles trigger inputs, they're just another input type treated specially, so there is simply an array of Inputs (one of which is a trigger), which opens the door in the future for the possibility of multiple triggers. See: https://github.com/Azure/azure-webjobs-sdk-script/blob/master/src/WebJobs.Script/Description/FunctionMetadata.cs

Kudu.Contracts/Jobs/FunctionTemplate.cs
```diff
@@ -0,0 +1,29 @@
...
+        public string Language { get; set; }
+    }
+
+    public class FunctionTemplateInput
+    {
+        [JsonProperty(PropertyName = "name")]
+        public string Name { get; set; }
+
+        [JsonProperty(PropertyName = "input_type")]
+        public string InputType { get; set; }
```
mathewc: nit: Just "Type"?

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -49,52 +52,291 @@ public FunctionManager(IEnvironment environment, ITraceFactory traceFactory)
...
+            var functionDir = Path.Combine(_environment.FunctionsPath, name);
+
+            // Assert templateId and value are both present
+            if (!string.IsNullOrEmpty(functionEnvelope?.TemplateId))
+            {
+                // check to create template.
+                if (FileSystemHelpers.DirectoryExists(functionDir) &&
+                    FileSystemHelpers.GetFileSystemEntries(functionDir).Any())
+                {
+                    throw new InvalidOperationException($"Function {name} already exist");
```
mathewc: nit: exists

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -49,52 +52,291 @@ public FunctionManager(IEnvironment environment, ITraceFactory traceFactory)
...
+            var configList = await Task.WhenAll(
+                    FileSystemHelpers
+                    .GetDirectories(_environment.FunctionsPath)
+                    .Select(d => Path.Combine(d, Constants.FunctionsConfigFile))
+                    .Where(FileSystemHelpers.FileExists)
+                    .Select(async f => { try { return await GetFunctionConfig(Path.GetFileName(Path.GetDirectoryName(f))); } catch { return null; } }));
+            return configList.Where(c => c != null);
+        }
+
+        public async Task<FunctionEnvelope> GetFunctionConfig(string name)
```
mathewc: Maybe add a TryGetFunctionConfig overload that does the error / exception handling so you don't have to inline that above?

Kudu.Contracts/Jobs/FunctionTemplate.cs
ahmelsayed: It's not that none of them uses outputs, but more that the selection of templates in the UI currently is based on the trigger type and language rather than anything else. The whole template story is very shaky hence the shaky APIs. I think we would need to move to sone sort of auto generated templates that cover all the possible combinations of inputs, triggers, and outputs? Maybe?

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -49,52 +52,291 @@ public FunctionManager(IEnvironment environment, ITraceFactory traceFactory)
...
+            }
+
+            throw new FileNotFoundException($"Host file ({path}) does not exist");
+        }
+
+        public async Task<JObject> PutHostConfig(JObject content)
+        {
+            var path = Path.Combine(_environment.FunctionsPath, Constants.FunctionsHostConfigFile);
+            await FileSystemHelpers.WriteAllTextToFileAsync(path, JsonConvert.SerializeObject(content));
+            return await GetHostConfig();
```
mathewc: Why does this have to go through a re-read via GetHostConfig?

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -49,52 +52,291 @@ public FunctionManager(IEnvironment environment, ITraceFactory traceFactory)
...
+        }
+
+        public FunctionEnvelope CreateFunctionConfig(string configContent, string functionName)
```
mathewc: Recommend getting the naming in sync. Either "FunctionDescriptor CreateFunctionDescriptor(...)" or "FunctionEnvelope CreateFunctionEnvelope".

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -49,52 +52,291 @@ public FunctionManager(IEnvironment environment, ITraceFactory traceFactory)
...
+                new FunctionTemplate { Id = "TimerTrigger", Language = "JavaScript", Trigger = "Timer" },
+                new FunctionTemplate { Id = "WebHook-Generic", Language = "JavaScript", Trigger = "WebHook-Generic" },
+                new FunctionTemplate { Id = "WebHook-GitHub", Language = "JavaScript", Trigger = "WebHook-GitHub" }
+            };
+        }
+
+        public void DeleteFunction(string name)
+        {
+            FileSystemHelpers.DeleteDirectorySafe(GetFunctionPath(name), ignoreErrors: false);
+            FileSystemHelpers.DeleteFileSafe(GetFunctionSampleDataFile(name));
```
mathewc: Nit: Add a "Path" suffix to these two methods, as you already have in "GetFunctionPath"?

mathewc: Do we want to leave the function Log directory, or delete that as well?

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -49,52 +52,291 @@ public FunctionManager(IEnvironment environment, ITraceFactory traceFactory)
...
+            {
+                return path;
+            }
+
+            throw new FileNotFoundException($"Function ({path}) does not exist");
+        }
+
+        private static async Task<IEnumerable<GitHubContent>> GetTemplatesFromGithub()
+        {
+            using (var client = GetHttpClient())
```
mathewc: Can't this client be reused for the lifetime of this class? It's properties don't appear to change.

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -49,52 +52,291 @@ public FunctionManager(IEnvironment environment, ITraceFactory traceFactory)
...
+        public async Task<JArray> GetTriggerInputs(ITracer tracer)
         {
             JArray inputs = new JArray();
-            foreach (var functionJson in EnumerateFunctionFiles())
+            foreach (var functionJson in await ListFunctionsConfig())
             {
                 try
                 {
-                    var json = JObject.Parse(FileSystemHelpers.ReadAllText(functionJson));
-                    var binding = json.Value<JObject>("bindings");
```
mathewc: This would be a great place / time to address https://github.com/projectkudu/kudu/issues/1891 while you're in here :)

Kudu.Services/Functions/FunctionController.cs
```diff
@@ -22,6 +25,100 @@ public FunctionController(IFunctionManager manager, ITraceFactory traceFactory)
...
+            }
+        }
+
+        [HttpPut]
+        public async Task<HttpResponseMessage> PutHostSettings()
+        {
+            var tracer = _traceFactory.GetTracer();
+            using (tracer.Step("FunctionsController.PutHostSettings()"))
+            {
+                return Request.CreateResponse(HttpStatusCode.Created, await _manager.PutHostConfig(await Request.Content.ReadAsAsync<JObject>()));
```
mathewc: Shouldn't we only return 201 created if we created a new resource, distinguished from if we've updated an existing resource? Or simply return a 200?

Kudu.Services/Functions/FunctionController.cs
```diff
@@ -22,6 +25,100 @@ public FunctionController(IFunctionManager manager, ITraceFactory traceFactory)
             _traceFactory = traceFactory;
         }
 
+        [HttpPut]
```
mathewc: Just curious: Why do we have these http verb attributes in addition to the HttpMethodConstraints above? Isn't one or the other sufficient?

Kudu.Services/Functions/FunctionController.cs
```diff
@@ -22,6 +25,100 @@ public FunctionController(IFunctionManager manager, ITraceFactory traceFactory)
...
+        public async Task<HttpResponseMessage> CreateOrUpdate(string name)
+        {
+            var tracer = _traceFactory.GetTracer();
+            using (tracer.Step($"FunctionsController.CreateOrUpdate({name})"))
+            {
+                try
+                {
+                    var functionEnvelope = await Request.Content.ReadAsAsync<FunctionEnvelope>();
```
mathewc: Might consider allowing WebApi handle the content deserialization for you. Example [here](https://github.com/Azure/azure-webjobs-sdk-script/blob/master/src/WebJobs.Script.WebHost/Controllers/AdminController.cs#L33).

Kudu.Contracts/Jobs/FunctionEnvelope.cs
ahmelsayed: We use the term envelope when referring to [ARM requests](https://github.com/projectkudu/kudu/blob/master/Kudu.Services/Arm/ArmUtils.cs#L15), not sure if this constitutes a naming pattern. I don't have a preference either way.

Kudu.Core/Functions/FunctionManager.cs
ahmelsayed: It's just the config property that gets written to disk, while the descriptor is always generated dynamically.

Kudu.Core/Functions/FunctionManager.cs
ahmelsayed: removed.

Kudu.Core/Environment.cs
```diff
@@ -270,6 +271,22 @@ public string SiteExtensionSettingsPath
...
+            {
+                return this.WebRootPath;
+            }
+        }
+
+        public string AppBaseUrlPrefix
+        {
+            get
+            {
+                return HttpContext.Current?.Request?.Url?.GetLeftPart(UriPartial.Authority);
```
davidebbo: Won't that cause it to be null in some cases? This doesn't seem expected based on usage. If not, should it throw instead?

Also, please add comment above it to show example of what that looks like, as `GetLeftPart` is not super descriptive.

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -49,52 +51,214 @@ public FunctionManager(IEnvironment environment, ITraceFactory traceFactory)
...
+
+        public async Task<FunctionEnvelope> GetFunctionConfigAsync(string name)
+        {
+            var path = Path.Combine(GetFunctionPath(name), Constants.FunctionsConfigFile);
+            if (FileSystemHelpers.FileExists(path))
+            {
+                return CreateFunctionConfig(await FileSystemHelpers.ReadAllTextFromFileAsync(path), name);
+            }
+
+            throw new FileNotFoundException($"Function ({path}) does not exist");
```
davidebbo: Shouldn't they error use `name` instead of `path`? Path is kind of impl detail at API level.

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -49,52 +51,214 @@ public FunctionManager(IEnvironment environment, ITraceFactory traceFactory)
...
+
+        public async Task<JObject> GetHostConfigAsync()
+        {
+            var path = Path.Combine(_environment.FunctionsPath, Constants.FunctionsHostConfigFile);
+            if (FileSystemHelpers.FileExists(path))
+            {
+                return JObject.Parse(await FileSystemHelpers.ReadAllTextFromFileAsync(path));
+            }
+
+            throw new FileNotFoundException($"Host file ({path}) does not exist");
```
davidebbo: We need to solve whose responsibility it is to create this file initially.

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -49,52 +51,214 @@ public FunctionManager(IEnvironment environment, ITraceFactory traceFactory)
         public bool IsFunctionEnabled()
         {
             // this should read appSettings instead
-            var hostJson = Path.Combine(_environment.WebRootPath, HostJsonFile);
+            var hostJson = Path.Combine(_environment.WebRootPath, Constants.FunctionsHostConfigFile);
```
davidebbo: `WebRootPath` should be `FunctionsPath`? You have the same Path.Combine 3 times, so maybe a helper.

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -49,52 +51,214 @@ public FunctionManager(IEnvironment environment, ITraceFactory traceFactory)
...
+        }
+
+        public async Task<JObject> PutHostConfigAsync(JObject content)
+        {
+            var path = Path.Combine(_environment.FunctionsPath, Constants.FunctionsHostConfigFile);
+            await FileSystemHelpers.WriteAllTextToFileAsync(path, JsonConvert.SerializeObject(content));
+            return await GetHostConfigAsync();
+        }
+
+        public FunctionEnvelope CreateFunctionConfig(string configContent, string functionName)
```
davidebbo: `private`

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -49,52 +51,214 @@ public FunctionManager(IEnvironment environment, ITraceFactory traceFactory)
...
-                var functionJson = Path.Combine(functionPath, FunctionJsonFile);
-                if (FileSystemHelpers.FileExists(functionJson))
+                return functionPath;
+            }
+            else if (functionFiles.Length == 1)
+            {
+                // if there is only a single file, that file is primary
+                return functionFiles[0];
+            }
+            else
```
davidebbo: I hate this `else` and the indentation it causes :)

davidebbo: But since it's copied from elsewhere, you can leave it :(

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -49,52 +51,214 @@ public FunctionManager(IEnvironment environment, ITraceFactory traceFactory)
...
+                if (string.IsNullOrEmpty(functionPrimary))
+                {
+                    // TODO: should this be an error?
+                    return functionPath;
+                }
+                return functionPrimary;
             }
         }
+
+        private Uri FilePathToVfsUri(string filePath, bool isDirectory = false)
```
davidebbo: Arguably more of a vfs utility than functions specific. Don't we have other APIs that return vfs paths?

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -49,52 +51,214 @@ public FunctionManager(IEnvironment environment, ITraceFactory traceFactory)
...
+
+        private string GetFunctionPath(string name)
+        {
+            var path = Path.Combine(_environment.FunctionsPath, name);
+            if (FileSystemHelpers.DirectoryExists(path))
+            {
+                return path;
+            }
+
+            throw new FileNotFoundException($"Function ({path}) does not exist");
```
davidebbo: `name`

Kudu.Core/Functions/FunctionManager.cs
davidebbo: Also, it's more correct to say that it's the function *config* that doesn't exist. If the function itself doesn't `GetFunctionPath` above will throw.

Kudu.Services/Diagnostics/LogStreamManager.cs
```diff
@@ -555,7 +555,9 @@ public ProcessRequestAsyncResult(HttpContext context, AsyncCallback callback, ob
 
                 _context.Response.Buffer = false;
                 _context.Response.BufferOutput = false;
-                _context.Response.ContentType = "text/plain";
+                _context.Response.ContentType = _context.Request.Headers["FunctionsPortal"] != null
+                    ? "custom-functions/stream"
+                    : "text/plain";
```
davidebbo: This was tricky stuff, so comment would be good.

Kudu.Services/Diagnostics/ProcessController.cs
```diff
@@ -468,7 +468,15 @@ private Process GetProcessById(int id)
         {
             try
             {
-                return id <= 0 ? Process.GetCurrentProcess() : Process.GetProcessById(id);
+                switch (id)
+                {
+                    case 0:
+                        return Process.GetCurrentProcess();
+                    case -1:
+                        return Process.GetProcessesByName("w3wp").FirstOrDefault(s => s.Id != Process.GetCurrentProcess().Id) ?? Process.GetCurrentProcess();
```
davidebbo: So it's the Main site, but if Main is not running you return the scm? Seems a bit funky. Needs comment to explain reasoning :)

Kudu.Core/Functions/FunctionManager.cs
ahmelsayed: @davidebbo What do you think the right behavior here should be? Should we delete the log directory for a function when that function is deleted? I think it makes sense.

Kudu.Services/Functions/FunctionController.cs
ahmelsayed: I never know with WebAPI. Sometimes I need to have it even though I have the HttpMethodConstraints, but I'm not sure. I see it in all of kudu's controllers though.

Kudu.Core/Functions/FunctionManager.cs
davidebbo: Delete it. While I could see arguments to keep it for posterity, it's going to create garbage over time and does more harm than good. In most cases, when you delete a Function, you stop caring about it.

Kudu.Services/Functions/FunctionController.cs
davidebbo: Let's not sweat it now. If we think it can be simplified throughout Kudu, we can do a separate pass for it.

Kudu.Core/Environment.cs
ahmelsayed: I did it like that to mirror how the [webjobs manager does it](https://github.com/projectkudu/kudu/blob/master/Kudu.Core/Jobs/JobsManagerBase.cs#L458-L470). We shouldn't be throwing a null ref though, right? I can throw InvalidOperation I think.

Kudu.Core/Functions/FunctionManager.cs
ahmelsayed: @mathewc What do you think? I think this is something that the runtime should create if it doesn't exist.

Kudu.Core/Environment.cs
davidebbo: Yes, definitely not null ref. That's just rude :)

Kudu.Core/Functions/FunctionManager.cs
ahmelsayed: I couldn't find any. I was looking at how the WebJobs does it, [but it seems to build the urls one by one](https://github.com/projectkudu/kudu/blob/ahmels-functions/Kudu.Core/Jobs/JobsManagerBase.cs#L427-L435).

davidebbo: Ok, just stay with what you have.

Kudu.Services/Diagnostics/ProcessController.cs
ahmelsayed: Do we want to get this change in? I needed it in the private kudu because the script host loads a native dll for edge js that get's locked and I can't update it.

davidebbo: Let's get it in, with 'Not Found' if no main w3wp.

PR #1896, opened 2016-02-17
#1891 Sync Triggers needs to respect Disabled functions

suwatch:


Kudu.Core/Infrastructure/OperationClient.cs
```diff
@@ -35,7 +35,17 @@ public OperationClient(ITracer tracer)
         public async Task<HttpResponseMessage> PostAsync<T>(string path, T content = default(T))
         {
             var jwt = System.Environment.GetEnvironmentVariable(Constants.X_MS_SITE_RESTRICTED_JWT);
+            if (String.IsNullOrEmpty(jwt))
```
davidebbo: This part is orthogonal to _Sync Triggers needs to respect Disabled functions_, right?

suwatch: Not really.  OperationClient is simply a HttpClient to call back to FE based using JWT.   It is payload agnostic.  There are currently 2 scenarios using OperationClient - 1. auto swap, 2. sync function trigger.

davidebbo:
:shipit:

PR #1898, opened 2016-02-17
Downgrade kudu to target .net 4.5.2 to run on mono

shrimpy:


Kudu.Services.Web/Web.config
```diff
@@ -10,9 +10,17 @@
     <add key="webpages:Version" value="3.0.0.0"/>
     <add key="webpages:Enabled" value="true"/>
   </appSettings>
+  <!--
+    For a description of web.config changes see http://go.microsoft.com/fwlink/?LinkId=235367.
```
davidebbo: I would yank all these comments. Other than that, looks good.

shrimpy: sure, cleaning up now.

PR #1899, opened 2016-02-17
fixing race condition with caching list of webjobs

shobak101:


Kudu.Services/Jobs/JobsController.cs
```diff
@@ -35,8 +35,8 @@ public JobsController(ITriggeredJobsManager triggeredJobsManager, IContinuousJob
         [HttpGet]
         public HttpResponseMessage ListAllJobs()
         {
-            IEnumerable<ContinuousJob> continuousJobs = _continuousJobsManager.ListJobs();
-            IEnumerable<TriggeredJob> triggeredJobs = _triggeredJobsManager.ListJobs();
+            IEnumerable<ContinuousJob> continuousJobs = _continuousJobsManager.ListJobs(false);
```
ahmelsayed: I think we try to use named parameters in kudu when the parameter is boolean.

Kudu.Core/Jobs/JobLogger.cs
```diff
@@ -68,6 +68,20 @@ protected string GetStatusFilePath()
         public void ReportStatus<TJobStatus>(TJobStatus status) where TJobStatus : class, IJobStatus
         {
             ReportStatus(status, logStatus: true);
+            if (status.GetType() == typeof(ContinuousJobStatus))
```
ahmelsayed: Maybe a comment explaining why we have to clear the cache after we report the status?

shobak101: done, also moved it to the more general reportstatus(status, bool)

Kudu.Services/Jobs/JobsController.cs
shobak101: done and in all other places too

Kudu.Core/Jobs/JobLogger.cs
davidebbo: I'm not seeing the change. Did you push?

The conditionals don't feel very polymorphic though, compare to calling virtual method which does the right thing in each derived class.

Kudu.Core/Jobs/JobsFileWatcher.cs
```diff
@@ -122,7 +122,7 @@ private void StartWatcher(object state)
                     _fileSystemWatcher.EnableRaisingEvents = true;
 
                     // Refresh all jobs
-                    IEnumerable<string> jobNames = _listJobNames();
+                    IEnumerable<string> jobNames = _listJobNames(true);
```
davidebbo: It's not clear looking at the call what this `true` means.

shobak101: fixed that by adding a named parameter to all places where this is called

Kudu.Core/Jobs/JobLogger.cs
shobak101: yes pushed, and tests passed. Suwat took a look and had no comments. yes it makes more sense to have that in a method for each, will do that quickly

Kudu.Core/Jobs/JobsManagerBase.cs
```diff
@@ -100,7 +102,10 @@ public void RegisterExtraEventHandlerForFileChange(Action<string> action)
 
         private void OnJobChanged(string jobName)
         {
-            JobListCache = null;
+            lock (jobsListCacheLockObj)
```
davidebbo: Would be cleaner to have a `ClearJobListCache()` helper to avoid having to lock from so many places.

Though I'm not sure I understand the purpose of the `lock` here since the `ListJobs` logic uses a local.

shobak101: well the lock is needed because the nulling can happen during the collection of information about jobs list, if that happens we end up with an invalid object that gets assigned to the list and there is no way of knowing that an invalidation happened during information collection. 

the local variable is not needed at this point actually so will clean that up, will add a method to invalidate the cache. also locking should not be a big problem as it happens on a site level, so the number of listing and changing jobs is not expected to be huge.

davidebbo: Ah ok, understood, thanks!

shobak101: Did some rebasing before pushing, can not send another update to this CR, closing and sending another one addressing David's comments

shobak101:
Did some rebasing before pushing, can not send another update to this CR, closing and sending another one addressing David's comments

Issue #1894, opened 2016-02-16
Unable to run custom post deployment script


litera:
I have an Asp.net Web application on Azure and I've also created a post deployment script, that used to run up until the end of the year (last time I was deploying was 23/12/2015) and the next time in this year when I deployed (first time in new year 11/01/2016) it wasn't running any more.

### What I have
1. I have `.deployment` file in my solution root (also git repo root)

    ```
    [config]
    SCM_POST_DEPLOYMENT_ACTIONS_PATH=deployment
    ```
2. I have folder `\deployment` in my root where I keep my special post deployment `cmd` file
3. I have set application setting in Azure portal to also define `SCM_POST_DEPLOYMENT_ACTIONS_PATH = deployment` although I think I only need to define it once (in file or app settings) even though I don't know which one takes precedence - but this is irrelevant as they both define the same value

And this used to work just fine as initially described. After *KuduSync.NET* copied changed files it automatically executed my custom post deployment script as shown in this log excerpt:
```
...
Copying file: '...'
Executing - 'D:\home\site\repository\deployment\PostDeployment.cmd'
... [my script output here]
'D:\home\site\repository\deployment\PostDeployment.cmd' executed successfully
Finished successfully.
```
But after the new year this is the output:
```
...
Copying file: '...'
Finished successfully.
```

### What I tried
1. I tried changing `SCM_POST_DEPLOYMENT_ACTIONS_PATH` to `POST_DEPLOYMENT_ACTIONS_DIR` in both my file and in Azure app settings
2. I tried setting path to `site\repository\deployment`
3. I removed setting from `.deployment` file keeping only Azure app setting
4. Changed path to be relative to deployed project instead of solution from `SCM_POST_DEPLOYMENT_ACTIONS_PATH=deployment` to `SCM_POST_DEPLOYMENT_ACTIONS_PATH=..\deployment`  

None of these worked in a sense that my custom script would even run. It doesn't get executed by Kudu at all. No error, no nothing. I don't see any changes that it would make nor do I see any output in the logs... **What am I doing wrong?**

davidebbo:
You describe how you set things up, but don't give info about the observed behavior. Did it cause a failure? Or did the flag simply get ignored? Does it work if you don't set the flag and drop a script under `site\deployments\tools\PostDeploymentActions`? Not saying you should be doing that, but to validate if at least that works.

litera:
[...]

PR #1901, opened 2016-02-18
Update to consume relative path for post deployment script environment variable

shrimpy:


ahmelsayed:
Is this a breaking change?

davidebbo:
@ahmelsayed it's the other way around: it fixes a recent breaking change :)

ahmelsayed:
I see

Kudu.Core/Deployment/Generator/ExternalCommandBuilder.cs
```diff
@@ -129,7 +128,7 @@ public IList<string> GetPostBuildActionScripts()
             var defaultPath = Path.Combine(Environment.DeploymentToolsPath, PostDeploymentActions);
             if (!string.IsNullOrEmpty(System.Environment.GetEnvironmentVariable(SettingsKeys.PostDeploymentActionsDirectory)))
```
davidebbo: Use `DeploymentSettings.GetValue()` instead or it won't work with `.deployment`.

davidebbo: Also, store it in variable to avoid double fetching.

shrimpy: yep, will do. just confirmed from our old code (reset and deploy), we do support settings from .deployment file, but we only expect absolute file path.

i will update the logic to honor settings from both environment variable and deployment setting, relative path and absolute path.

davidebbo: If old logic only supported absolute file path, what caused the regression in #1894?

davidebbo:
:shipit:

PR #1902, opened 2016-02-20
Add support for including code files with functions

davidebbo:
Mostly for @ahmelsayed, as this is going into his branch and not master.

The idea is to support:

```
{
  "config": {

  },
  "files":{
      "foo.js": "bar",
      "hello.js": "This is a test"
  }
}
```

### DIFF
--- a/Kudu.Contracts/Jobs/FunctionEnvelope.cs
+++ b/Kudu.Contracts/Jobs/FunctionEnvelope.cs
@@ -33,5 +33,8 @@ namespace Kudu.Contracts.Jobs
 
         [JsonProperty(PropertyName = "config")]
         public JObject Config { get; set; }
+
+        [JsonProperty(PropertyName = "files")]
+        public JObject Files { get; set; }
     }
 }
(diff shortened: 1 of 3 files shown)


Kudu.Contracts/Jobs/FunctionEnvelope.cs
```diff
@@ -33,5 +33,8 @@ public class FunctionEnvelope : INamedObject
 
         [JsonProperty(PropertyName = "config")]
         public JObject Config { get; set; }
+
+        [JsonProperty(PropertyName = "files")]
+        public JObject Files { get; set; }
```
ahmelsayed: Why not `Dictionary<string, string>`?

ahmelsayed: No we don't have any unite tests yet :) I know I'll add some

ahmelsayed:
No we don't have any unite tests yet :) I know I'll add some

Kudu.Contracts/Jobs/FunctionEnvelope.cs
davidebbo: Changed as suggested.

davidebbo:
I also added ARM support to the controller.

suwatch:
just happen to glance around the implementation.   Unrelated to this change, when doing GetTemplatesFromGithub or DownloadFromGithub, we should have tracing around it.  It will help us diagnose issues.

davidebbo:
We'll probably be yanking that GitHub logic and manage the templating from the outside instead (making it more of a normal creation).

PR #1907, opened 2016-02-25
Use Consolas as the command window font

lindydonna:
Put Consolas in font-family list before Lucida Console, since Lucida Console is an ugly font, and Consolas is beautiful.

### DIFF
--- a/Kudu.Services.Web/Content/Styles/FileBrowser.css
+++ b/Kudu.Services.Web/Content/Styles/FileBrowser.css
@@ -86,7 +86,7 @@ div.upload-unzip-show {
 }
 
 div.zip-upload-text {
-    font-family: 'Lucida Console', Consolas, 'Courier New', monospace;
+    font-family: Consolas, 'Lucida Console', 'Courier New', monospace;
     font-size: large;
     position: relative;
     text-align: center;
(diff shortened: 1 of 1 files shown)


davidebbo:
Ok, so current is:

image

And with your change:

image

Not sure I can even tell the difference! :)

suwatch:
All my command shell font is Consolas.  Hence, I personally like this change.

lindydonna:
@davidebbo The first one is ugly, the second one is nice! :)  I'm a font person.

davidebbo:
Ok, it's in! But I'll have to challenge you guys to a side by side test to see if you're not completely full of it.

jeme:
diff

One is slightly wider than the other and a few letters are a bit taller, that seems to be about it... (Notice how they blend perfectly on the first couple of letters)... >.<...

davidebbo:
Yep, it's slightly narrower, That's the only difference :)

lindydonna:
Thanks for merging!  @davidebbo Feel free to give me a test anytime--I cant tell the difference between those fonts quite easily. 

One of my friends is even more crazy about fonts than me, and can tell the difference between Arial and Helvetica. (I can't)

PR #1908, opened 2016-02-25
filter additional path to reduce noise in kudu traces

suwatch:


suwatch:
Discover these noises while investigating customer issue with setup continuous deployment.

davidebbo:
:shipit:

PR #1920, opened 2016-03-08
Create directory if it doesn't exist in LogStreamingManager if the re…

ahmelsayed:
…quest to stream logs was coming from functions portal

### DIFF
--- a/Kudu.Services/Diagnostics/LogStreamManager.cs
+++ b/Kudu.Services/Diagnostics/LogStreamManager.cs
@@ -18,6 +18,7 @@ using Kudu.Services.Infrastructure;
 
 using Environment = System.Environment;
 using System.Diagnostics.CodeAnalysis;
+using Kudu.Services.Diagnostics;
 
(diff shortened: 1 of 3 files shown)


Kudu.Services/Diagnostics/LogStreamManager.cs
```diff
@@ -74,10 +75,14 @@ public IAsyncResult BeginProcessRequest(HttpContext context, AsyncCallback cb, o
             });
 
             string path = ParseRequest(context);
-            if (!Directory.Exists(path))
+            if (!Directory.Exists(path) && !context.Request.IsFunctionPortalRequest())
```
davidebbo: Why special case?

ahmelsayed: I figured it's technically a breaking change. Not sure why someone would've relied on the old behavior, but if you think it's okay to always create the folder if it doesn't exist, then I can change it to do that.

PR #1927, opened 2016-03-11
Fix #1926 sync triggers even no inputs

### LINKED ISSUE #1926: Sync Triggers should still call to register even no input triggers
Currently, we don't call to register triggers if [no input triggers](https://github.com/projectkudu/kudu/blob/master/Kudu.Core/Functions/FunctionManager.cs#L47).   This is not entirely correct.  Scale controller is also interested in the fact that container has no trigger.

suwatch:


shrimpy:
:shipit:

PR #1936, opened 2016-03-17
fix initial put function may not have host.json

suwatch:
host.json is created by runtime.  During initial put of function, it may not exist.  As a result, no trigger is synched.   Change to use env instead.


davidebbo:
:shipit:

davidebbo:
Merged

PR #1942, opened 2016-03-23
webjob startup issue led to no job running

suwatch:
The key fix is to throw if IOException when building job.   This caused the start up codes in JobsFileWatcher to retry.    The rest of the fix is to add traces to help investigate future issue.

### DIFF
--- a/Kudu.Core/Jobs/JobsManagerBase.cs
+++ b/Kudu.Core/Jobs/JobsManagerBase.cs
@@ -117,12 +117,14 @@ namespace Kudu.Core.Jobs
         {
             lock (jobsListCacheLockObj)
             {
-                if (JobListCache == null || forceRefreshCache)
+                var jobList = JobListCache;
+                if (jobList == null || forceRefreshCache)
                 {
-                    JobListCache = ListJobsInternal();
+                    jobList = ListJobsInternal();
+                    JobListCache = jobList;
                 }
+                return jobList;
             }
-            return JobListCache;
         }
 
         internal static void ClearJobListCache()
@@ -308,6 +310,13 @@ namespace Kudu.Core.Jobs
                 // Return a job representing an error for no runnable script file found for job
                 if (nullJobOnError)
                 {
+                    if (ex is IOException)
+                    {
+                        // unexpected IOException given simply reading files/folders.
+                        // just rethrow and let the upper layers handle it.
+                        throw;
+                    }
+
                     return null;
                 }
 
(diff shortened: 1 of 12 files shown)


Kudu.Core/Jobs/JobsManagerBase.cs
```diff
@@ -117,12 +117,14 @@ public IEnumerable<TJob> ListJobs(bool forceRefreshCache)
...
+                var jobList = JobListCache;
+                if (jobList == null || forceRefreshCache)
                 {
-                    JobListCache = ListJobsInternal();
+                    jobList = ListJobsInternal();
+                    JobListCache = jobList;
                 }
+                return jobList;
             }
-            return JobListCache;
```
suwatch: the fix here is we return JobListCache outside the lock.   This could be null (I ran into nullref myself) as it is being set to null by other code path.   Use local variable to make it more robust.  In fact, we could remove the `lock(jobsListCacheLockObj)`.

Kudu.Core/Jobs/JobsManagerBase.cs
```diff
@@ -308,6 +310,13 @@ protected TJob BuildJob(DirectoryInfoBase jobDirectory, bool nullJobOnError = tr
                 // Return a job representing an error for no runnable script file found for job
                 if (nullJobOnError)
                 {
+                    if (ex is IOException)
+                    {
+                        // unexpected IOException given simply reading files/folders.
+                        // just rethrow and let the upper layers handle it.
+                        throw;
+                    }
```
suwatch: This is the key fix.   This rethrows unexpected IOException and let the upper layers retry.

Kudu.Core/Jobs/TriggeredJobSchedule.cs
```diff
@@ -50,6 +52,8 @@ public void Reschedule(DateTime lastRun, Schedule schedule = null)
                 Logger.LogInformation("Next schedule expected in " + nextInterval);
             }
 
+            _analytics.JobEvent(TriggeredJob.Name.Fuzz(), $"Reschedule {nextInterval}", TriggeredJob.JobType, String.Empty);
```
suwatch: This is the place where we reschedule the next run based on cron expression.   This should help us debug.

Kudu.Core/Jobs/JobsManagerBase.cs
davidebbo: Nice catch!

Kudu.Core/Jobs/TriggeredJobSchedule.cs
davidebbo: Is it really worth fuzzing the job name? Given that it belongs in ARM path's, it's not 'sensitive'.

davidebbo:
Look great, and should really help us debug this next time if some issues remain.

davidebbo:
:shipit:

davidebbo:
This is in, right? Can close?

PR #1943, opened 2016-03-23
Functions rbac

davidebbo:
Note that most of the `GetFunctionSecretsAsync` is just moved from below (changed to async).

There is also a small unrelated change to fix a data folder portal issue.

### DIFF
--- a/Kudu.Services.Web/Tracing/TraceModule.cs
+++ b/Kudu.Services.Web/Tracing/TraceModule.cs
@@ -26,6 +26,7 @@ namespace Kudu.Services.Web.Tracing
         private static Regex[] _rbacWhiteListPaths = new[]
         {
             new Regex(@"^/api/siteextensions(/|$)", RegexOptions.IgnoreCase),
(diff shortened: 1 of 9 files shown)


Kudu.Services.Web/Tracing/TraceModule.cs
```diff
@@ -26,6 +26,7 @@ public class TraceModule : IHttpModule
         private static Regex[] _rbacWhiteListPaths = new[]
         {
             new Regex(@"^/api/siteextensions(/|$)", RegexOptions.IgnoreCase),
+            new Regex(@"^/api/functions((/|$)([^/]*|$)){0,2}(/|$)$", RegexOptions.IgnoreCase),
```
suwatch: Consider adding test [here](https://github.com/projectkudu/kudu/blob/master/Kudu.Core.Test/TraceModuleFacts.cs#L8).

davidebbo: Added tests, thanks.

suwatch:
:shipit:

PR #1945, opened 2016-03-24
use capital case for Message to show up in MDS

suwatch:
Caught while verifying in production and message didn't show up at the right column.   This is consistent pattern with other event (see KuduException for instance).


davidebbo:
:shipit:

PR #1951, opened 2016-03-28
Run kudu on mono

shrimpy:
- Run kudu on mono
- Solve issue that deployment lock not working on kudu.exe
- Update PathUtility to aware of linux
- Grant full permission to post receive hook
- Hide unsupported Kudu feature from Linux enviroment.

### DIFF
--- a/Common/Constants.cs
+++ b/Common/Constants.cs
@@ -1,4 +1,5 @@
 using System;
+using System.IO;
 
 namespace Kudu
 {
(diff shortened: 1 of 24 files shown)


Common/Constants.cs
```diff
@@ -22,7 +23,7 @@ public static class Constants
         public const string SiteFolder = @"site";
         public const string LogFilesPath = @"LogFiles";
         public const string ApplicationLogFilesDirectory = "Application";
-        public const string TracePath = LogFilesPath + @"\kudu\trace";
+        public readonly static string TracePath = Path.Combine(LogFilesPath, "kudu", "trace");
```
davidebbo: Why this change? Having a true const seems preferable.

shrimpy: When running on mono, it doesn`t like the path separator '\', and we end up creating a folder name with "LogFiles\kudu\trace" instead of nested folders. using "Path.Combine" it will use the right separator when running on different environment.

PR #1959, opened 2016-04-06
Enabe basic Kudu console on Linux

shrimpy:
Screenshot:

image


Kudu.Services.Web/DebugConsole/WindowsConsole.cshtml
```diff
@@ -0,0 +1,190 @@
+﻿@{
```
shrimpy: No code change, this file is a copy of the original Default.cshtml

PR #1962, opened 2016-04-12
handle corrupted job status file

suwatch:
Treat corrupted job status file as initial job run state (as if no job status at all).

### DIFF
--- a/Kudu.Core/Jobs/JobsManagerBase.cs
+++ b/Kudu.Core/Jobs/JobsManagerBase.cs
@@ -422,7 +422,7 @@ namespace Kudu.Core.Jobs
 
         protected TJobStatus GetStatus<TJobStatus>(string statusFilePath) where TJobStatus : class, IJobStatus, new()
         {
-            return JobLogger.ReadJobStatusFromFile<TJobStatus>(TraceFactory, statusFilePath) ?? new TJobStatus();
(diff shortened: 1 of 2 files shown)


Kudu.Core/Jobs/JobsManagerBase.cs
```diff
@@ -422,7 +422,7 @@ private static string GetJobSettingsPath(DirectoryInfoBase jobDirectory)
 
         protected TJobStatus GetStatus<TJobStatus>(string statusFilePath) where TJobStatus : class, IJobStatus, new()
         {
-            return JobLogger.ReadJobStatusFromFile<TJobStatus>(TraceFactory, statusFilePath) ?? new TJobStatus();
+            return JobLogger.ReadJobStatusFromFile<TJobStatus>(TraceFactory, statusFilePath);
```
davidebbo: No sure I understand why it's wrong to return a `new` object when `ReadJobStatusFromFile` returns null.

davidebbo: Ok, saw your comments in mail.

davidebbo:
:shipit:

suwatch:
https://github.com/projectkudu/kudu/commit/63e4c4003bf856ca23d0c319b024684d7765df87

PR #1965, opened 2016-04-13
Enable basic Kudu console on other OS.

shrimpy:


Kudu.Services.Web/DebugConsole/Default.cshtml
```diff
@@ -1,186 +1,8 @@
-﻿@{
-    Layout = "~/_Layout.cshtml";
-    Page.Title = "Diagnostic Console";
+﻿@if (Kudu.Core.Helpers.OSDetecter.IsOnWindows())
+{
+    @RenderPage("~/DebugConsole/WindowsConsole.cshtml");
```
davidebbo: `@RenderPage("WindowsConsole.cshtml")` would be cleaner to avoid baking assumptions on the folder name.

Kudu.Services.Web/Default.cshtml
```diff
@@ -73,26 +73,29 @@
...
-            <a href="api/processes">Processes and mini-dumps</a>
-        </li>
-        <li>
-            <a href="api/diagnostics/runtime">Runtime versions</a>
-        </li>
-        <li>
-            Site Extensions: <a href="api/siteextensions">installed</a> | <a href="api/extensionfeed">feed</a>
-        </li>
-        <li>
             <a href="api/scm/info">Source control info</a>
```
davidebbo: The ordering of 'Source control info' is different from before. Not sure if that's intentional.

Kudu.Services.Web/_Layout.cshtml
```diff
@@ -42,7 +54,7 @@
             </div>
             <div class="collapse navbar-collapse">
                 <ul class="nav navbar-nav">
-                    <li><a href="~/Env">Environment</a></li>
+                    <li><a href="~/Env.cshtml">Environment</a></li>
```
davidebbo: Why doesn't /env just work? That's a small regression on Windows if we force the extension.

Kudu.Services/Infrastructure/UriHelper.cs
```diff
@@ -1,9 +1,29 @@
...
     {
+        public static Uri GetBaseUri(HttpRequestMessage request)
+        {
+            if (request == null)
+            {
+                throw new ArgumentNullException("request");
+            }
+
+            IEnumerable<string> disguisedHostValues = new List<string>();
+            if (request.Headers.TryGetValues("DISGUISED-HOST", out disguisedHostValues) && disguisedHostValues.Count() > 0)
```
davidebbo: This needs some explaining. I don't know what `DISGUISED-HOST` is.

Kudu.Services/Infrastructure/UriHelper.cs
```diff
@@ -1,9 +1,29 @@
...
 
 namespace Kudu.Services.Infrastructure
 {
     public static class UriHelper
     {
+        public static Uri GetBaseUri(HttpRequestMessage request)
+        {
+            if (request == null)
+            {
+                throw new ArgumentNullException("request");
```
davidebbo: We don't typically do argument checks for things we only call ourselves. This is not a framework.

Kudu.Services/Infrastructure/UriHelper.cs
shrimpy: FrontedEnd will always pass this header to Kudu (confirmed with Jenny), and it always carry the right host name.
 
when running Kudu on mono in a container, container is bind to a local ip, so we cannot rely on "request.RequestUri", which will be "127.0.0.1:xxxx". hence this fix. i will update code to have more comments.

Kudu.Services/Infrastructure/UriHelper.cs
shrimpy: sure. might be over defensive here. will take it out.

shrimpy:
Close this PR, new changes are in https://github.com/projectkudu/kudu/pull/1978

PR #1967, opened 2016-04-13
adding lock object to JobmanagerBase ListJobs to fix a possible race Condition

shobak101:
if 2 requests come in while the cache is null, both will run in parallel and because listjobsInternal is a long running call (up to 20 seconds) the variation in the time the call finishes is big. Thus a new change induced call could finish earlier than another old change induced call, which will make the cache have an invalid state. Locking makes sure only one runs at a time, which simplifies the problem and makes sure no such race conditions happen. also locking is happening per site so no fear of it consuming a lot of time or slowing responses very much.


suwatch:
:shipit:

davidebbo:
@shobak101 Is this something we want to get in with next RU? If so, please get it in all the way so we get a build. Thanks!

PR #1969, opened 2016-04-15
Fixed deployment log uri and enable ssh client.

shrimpy:
- Fixed deployment log uri and enable ssh client.
- Update to use bufferless stream.

### DIFF
--- a/Kudu.Core/Environment.cs
+++ b/Kudu.Core/Environment.cs
@@ -6,6 +6,7 @@ using System.Runtime.InteropServices;
 using System.Security;
 using Kudu.Core.Infrastructure;
 using System.Web;
+using Kudu.Core.Helpers;
 
 namespace Kudu.Core
 {
@@ -95,7 +96,17 @@ namespace Kudu.Core
             _siteExtensionSettingsPath = Path.Combine(SiteRootPath, Constants.SiteExtensionsCachePath);
             _diagnosticsPath = Path.Combine(SiteRootPath, Constants.DiagnosticsPath);
             _locksPath = Path.Combine(SiteRootPath, Constants.LocksPath);
-            _sshKeyPath = Path.Combine(rootPath, Constants.SSHKeyPath);
+
+            string siteName = System.Environment.GetEnvironmentVariable("WEBSITE_SITE_NAME");
+            if (OSDetecter.IsOnWindows() || rootPath.TrimEnd(new char[] { '/', '\\' }).EndsWith(siteName, StringComparison.OrdinalIgnoreCase))
+            {
+                _sshKeyPath = Path.Combine(rootPath, Constants.SSHKeyPath);
+            }
+            else
+            {
+                // in linux, rootPath is "/home", while .ssh folder need to under "/home/{user}"
+                _sshKeyPath = Path.Combine(rootPath, System.Environment.GetEnvironmentVariable("WEBSITE_SITE_NAME"), Constants.SSHKeyPath);
+            }
             _scriptPath = Path.Combine(binPath, Constants.ScriptsPath);
             _nodeModulesPath = Path.Combine(binPath, Constants.NodeModulesPath);
             _logFilesPath = Path.Combine(rootPath, Constants.LogFilesPath);
--- a/Kudu.Core/Permissions/PermissionHandler.cs
+++ b/Kudu.Core/Permissions/PermissionHandler.cs
@@ -1,0 +1,34 @@
+﻿using System.IO;
+using Kudu.Contracts.Permissions;
+using Kudu.Contracts.Settings;
+using Kudu.Core.Deployment;
+using Kudu.Core.Deployment.Generator;
+using Kudu.Core.Infrastructure;
+
+namespace Kudu.Core.Permissions
+{
+    /// <summary>
+    /// This handler is meant for Linux enviroment only
+    /// </summary>
+    public class PermissionHandler : IPermissionHandler
+    {
+        private IEnvironment _environment;
+        private IDeploymentSettingsManager _settings;
+        private ILogger _logger;
+
+        public PermissionHandler(IEnvironment env, IDeploymentSettingsManager settings, ILogger logger)
+        {
+            _environment = env;
+            _settings = settings;
+            _logger = logger;
+        }
+
+        public void Chmod(string permission, string filePath)
+        {
+            var folder = new DirectoryInfo(filePath).Parent;
+            var exeFactory = new ExternalCommandFactory(_environment, _settings, null);
+            Executable exe = exeFactory.BuildCommandExecutable("/bin/chmod", folder.FullName, _settings.GetCommandIdleTimeout(), _logger);
+            exe.Execute("{0} {1}", permission, filePath);
+        }
+    }
+}
--- a/Kudu.Services.Web/App_Start/NinjectServices.cs
+++ b/Kudu.Services.Web/App_Start/NinjectServices.cs
@@ -1,9 +1,9 @@
 using System;
 using System.Collections.Generic;
+using System.Configuration;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
-using System.Net;
 using System.Net.Http.Formatting;
 using System.Web;
 using System.Web.Hosting;
@@ -11,6 +11,7 @@ using System.Web.Http;
 using System.Web.Routing;
 using Kudu.Contracts.Infrastructure;
 using Kudu.Contracts.Jobs;
+using Kudu.Contracts.Permissions;
 using Kudu.Contracts.Settings;
 using Kudu.Contracts.SiteExtensions;
 using Kudu.Contracts.SourceControl;
@@ -19,9 +20,12 @@ using Kudu.Core;
 using Kudu.Core.Commands;
 using Kudu.Core.Deployment;
 using Kudu.Core.Deployment.Generator;
+using Kudu.Core.Functions;
+using Kudu.Core.Helpers;
 using Kudu.Core.Hooks;
 using Kudu.Core.Infrastructure;
 using Kudu.Core.Jobs;
+using Kudu.Core.Permissions;
 using Kudu.Core.Settings;
 using Kudu.Core.SiteExtensions;
 using Kudu.Core.SourceControl;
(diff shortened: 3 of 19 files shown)


Kudu.Core/Environment.cs
```diff
@@ -95,7 +96,17 @@ public class Environment : IEnvironment
             _siteExtensionSettingsPath = Path.Combine(SiteRootPath, Constants.SiteExtensionsCachePath);
             _diagnosticsPath = Path.Combine(SiteRootPath, Constants.DiagnosticsPath);
             _locksPath = Path.Combine(SiteRootPath, Constants.LocksPath);
-            _sshKeyPath = Path.Combine(rootPath, Constants.SSHKeyPath);
+
+            string siteName = System.Environment.GetEnvironmentVariable("WEBSITE_SITE_NAME");
+            if (OSDetecter.IsOnWindows() || rootPath.TrimEnd(new char[] { '/', '\\' }).EndsWith(siteName, StringComparison.OrdinalIgnoreCase))
```
davidebbo: Not sue I understand. When would the root path end with the site name? Probably need comments with examples to illustrate.

Kudu.Core/Environment.cs
```diff
@@ -95,7 +96,17 @@ public class Environment : IEnvironment
...
+            {
+                _sshKeyPath = Path.Combine(rootPath, Constants.SSHKeyPath);
+            }
+            else
+            {
+                // in linux, rootPath is "/home", while .ssh folder need to under "/home/{user}"
+                _sshKeyPath = Path.Combine(rootPath, System.Environment.GetEnvironmentVariable("WEBSITE_SITE_NAME"), Constants.SSHKeyPath);
```
davidebbo: Use `siteName` since you have it from above.

davidebbo: The comment says `{user}`, but the code uses the site name. Maybe they're the same? If so, comment should clarify.

shrimpy: as intend to use "siteName" and not getting enviroment var again. will update. thanks for catching this.

yes, user name and site name always the same. will add more comment.

Kudu.Core/Environment.cs
shrimpy: when run in container, there is code to map root (%HOME% environment variable) to /home.

since Kudu is an open source project, normally ROOT for a user is /home/{user}, e.g when i tested on a linux machine but not using container. 

i make it easier to test kudu on both our container environment and normal Linux environment.

let me know if you think this is over kill. or i will add more comment and keep it.

Kudu.Core/Permissions/PermissionHandler.cs
```diff
@@ -0,0 +1,34 @@
...
+using Kudu.Contracts.Permissions;
+using Kudu.Contracts.Settings;
+using Kudu.Core.Deployment;
+using Kudu.Core.Deployment.Generator;
+using Kudu.Core.Infrastructure;
+
+namespace Kudu.Core.Permissions
+{
+    /// <summary>
+    /// This handler is meant for Linux enviroment only
```
davidebbo: But looking at the code, it looks like the class is also instantiated on Windows, no?

Kudu.Core/Environment.cs
shrimpy: hmm. i think i didn`t think thru properly. during testing username and site name will be different. i will take it out.

Kudu.Core/Permissions/PermissionHandler.cs
```diff
@@ -0,0 +1,34 @@
...
+        public PermissionHandler(IEnvironment env, IDeploymentSettingsManager settings, ILogger logger)
+        {
+            _environment = env;
+            _settings = settings;
+            _logger = logger;
+        }
+
+        public void Chmod(string permission, string filePath)
+        {
+            var folder = new DirectoryInfo(filePath).Parent;
```
davidebbo: Shouldn't it be `FileInfo` if it's a file?

davidebbo: `Path.GetDirectoryName` might be more what you want.

Kudu.Core/Permissions/PermissionHandler.cs
shrimpy: i would think Ninject will delay instantiate it if no one call it.

Kudu.Core/Permissions/PermissionHandler.cs
shrimpy: right, will update.

Kudu.Services.Web/App_Start/NinjectServices.cs
```diff
@@ -720,11 +724,24 @@ private static IEnvironment GetEnvironment(IDeploymentSettingsManager settings =
             string root = PathResolver.ResolveRootPath();
             string siteRoot = Path.Combine(root, Constants.SiteFolder);
             string repositoryPath = Path.Combine(siteRoot, settings == null ? Constants.RepositoryPath : settings.GetRepositoryPath());
+            string binPath = HttpRuntime.BinDirectory;
+
+            if (!string.IsNullOrWhiteSpace(binPath) && !OSDetecter.IsOnWindows())
+            {
+                int binIdx = binPath.LastIndexOf("Bin", StringComparison.Ordinal);
```
davidebbo: That will match any instance of 'Bin' in the full path. e.g. it could be part of the site name. Doesn't seem right.

shrimpy: binPath is from "HttpRuntime.BinDirectory", it always pointing to Kudu`s bin path, always end with ..../Bin, unless we change Kudu`s folder structure.
why we need to worry about site name here?

Kudu.Services/Deployment/DeploymentController.cs
```diff
@@ -111,8 +111,9 @@ public void Delete(string id)
                             {
                                 CreateDeployment(deployResult, result.Value<string>("details"));
 
-                                deployResult.Url = Request.RequestUri;
-                                deployResult.LogUrl = UriHelper.MakeRelative(Request.RequestUri, "log");
+                                Uri baseUri = UriHelper.MakeRelative(UriHelper.GetBaseUri(Request), Request.RequestUri.AbsolutePath);
```
davidebbo: I struggle to understand this line. Why isn't baseUri just `UriHelper.GetBaseUri(Request)`? That's what the method name would imply.

shrimpy: i think I should rename the variable to be deploymentUri to make it more clear. 

e.g if final url is 
````
https://kudutry.scm.azurewebsites.net/api/deployments/ef52ec67fc9574e726955a9cbaf7bcba791e4e95/log
````

baseUri is inted to be 
````
https://kudutry.scm.azurewebsites.net/api/deployments/ef52ec67fc9574e726955a9cbaf7bcba791e4e95
````

Kudu.Core/Permissions/PermissionHandler.cs
davidebbo: I don't see how it could do that. All the classes that take in an `IPermissionHandler` will receive an instance of this class even if they never call `Chmod`.

Generally, I feel there is way too much added Ninject and parameter passing everywhere just to call Chmod is a couple places. Can't Chmod just be a static helper called conditionally, without affecting anything else?

Kudu.Services/Deployment/DeploymentController.cs
davidebbo: I find that having specific examples as comment showing what string handling is doing really helps read through the code.

Kudu.Core/Environment.cs
davidebbo: I do agree that it's worth making it runnable both on normal Linux and in container, to help testing.

Kudu.Core/Permissions/PermissionHandler.cs
shrimpy: static function was my first implementation, but since i will need Enviroment, DeploymentSettingsManager and Logger, thought to follow the pattern to use Ninject. 

Let me refactor it to be static function.

Kudu.Services.Web/App_Start/NinjectServices.cs
davidebbo: Concern was that code tests if it finds a match, implying it may not. But if it doesn't end with bin, it may end up finding some unrelated 'bin' in the middle of the path. Shouldn't it test for EndsWith instead?

Kudu.Core/Permissions/PermissionHandler.cs
davidebbo: Since the context should never affect what it does, maybe you can have a singleton instance that gets created early in Ninject init (not request context), and then is used everywhere? Well, not sure if that would mess up the logger. Maybe discuss with @suwatch. But it does seem a bit overkill for such a simple need.

Kudu.Services.Web/App_Start/NinjectServices.cs
shrimpy: yes, i do EndWIth at line 736. use LastIndexOf is to make sure i cover both "bin" and "bin/" cases

Kudu.Core/Permissions/PermissionHandler.cs
suwatch: If the intention is simply Chmod of the id_rsa and its associates (to workaround certain OS), having IPermissionHandler might be a bit overkilled (if I read the code correctly).    We could something along this line.

1) to back fill old sites, you could do similarly [here](https://github.com/projectkudu/kudu/blob/6886065d11e1f3b37f52dcc65a69c7eb4de3d95d/Kudu.Services.Web/App_Start/NinjectServices.cs#L552). 

2) For new file, also do it where we initially create [the id_rsa file](https://github.com/projectkudu/kudu/blob/6886065d11e1f3b37f52dcc65a69c7eb4de3d95d/Kudu.Core/SSHKey/SSHKeyManager.cs).

Is this class useful for other scenarios?

shrimpy: for now it is just one. just to prepare we might have other command, e.g chown
I think I will change to static function. later when there is a better need then we can revisit this.

suwatch: Another idea is to do it lazily when needed.   For instance, right before we [fetch] (https://github.com/projectkudu/kudu/blob/6886065d11e1f3b37f52dcc65a69c7eb4de3d95d/Kudu.Services/ServiceHookHandlers/FetchHandler.cs#L238) or with the fetch itself.   You will have everything needed in the context.

I think this is the only place we actually use SSH.

Kudu.Core/Environment.cs
suwatch: WEBSITE_SITE_NAME is the same for all slots.   Do use `APP_POOL_ID = ~1suwatchslot01__fb0e` (and trim ~1 prefix) instead.

shrimpy:
Close this PR, new changes are in https://github.com/projectkudu/kudu/pull/1978

PR #1976, opened 2016-04-25
adding cache expiry logic to jobs list cache and invalidating the cac…

shobak101:
…he when job deletion api called


Kudu.Core/Jobs/JobsManagerBase.cs
```diff
@@ -67,6 +67,10 @@ public abstract class JobsManagerBase<TJob> : JobsManagerBase, IJobsManager<TJob
         protected JobsFileWatcher JobsWatcher { get; set; }
         internal static IEnumerable<TJob> JobListCache { get; set; }
 
+        private DateTime JobListCacheExpiryDate;
```
davidebbo: Fields are normally named `_jobListCacheExpiryDate`

davidebbo:
Other than minor naming issue, looks good.

PR #1980, opened 2016-04-29
add tool paths to %path% to anticipate git 2.8.1 installation

suwatch:
The key change is to add more tool path to %path%.   Scanning thru codes, I noticed that many api relying on where Git folder was (some hardcoded and expected 32 bits being installed).   

I did a clean up so we can tolerate either 32 or 64 bits of git installation (in public Kudu scenario).


davidebbo:
:shipit:

Issue #1984, opened 2016-05-04
host.json may have invalid json


develohpanda:
It is possible that the host.json file contain invalid json - there is currently no error handling for this and an unhandled exception is thrown when parsing.

mathewc:
Is this issue intended for the azure-webjobs-sdk-script repo? I see your PR https://github.com/Azure/azure-webjobs-sdk-script/pull/348 for this.

develohpanda:
Nope this issue is still relevant here, malformed host.json still not handled. in the closed PR, @davidebbo mentioned behavior needs improvement.

davidebbo:
Fixed per #1985

PR #1985, opened 2016-05-04
Better handling of invalid host.json

develohpanda:
Fixes issue #1984 - gracefully handles host.json file having invalid json with a cleaner error message as opposed to what is shown in the issue, that boils up to the portal in Azure. Now user will explicitly see that they are seeing issues because their host.json is invalid.

### DIFF
--- a/Kudu.Core/Functions/FunctionManager.cs
+++ b/Kudu.Core/Functions/FunctionManager.cs
@@ -205,9 +205,28 @@ namespace Kudu.Core.Functions
 
         public async Task<JObject> GetHostConfigAsync()
         {
-            return FileSystemHelpers.FileExists(HostJsonPath)
-                ? JObject.Parse(await FileSystemHelpers.ReadAllTextFromFileAsync(HostJsonPath))
-                : new JObject();
+            var host = await TryGetHostConfigAsync();
+            if (host == null)
+            {
+                throw new FileNotFoundException("Host.json is invalid");
+            }
+            return host;
+        }
+
+        private async Task<JObject> TryGetHostConfigAsync()
+        {
+            try
+            {
+                return FileSystemHelpers.FileExists(HostJsonPath)
+                    ? JObject.Parse(await FileSystemHelpers.ReadAllTextFromFileAsync(HostJsonPath))
+                    : new JObject();
+            }
+            catch
+            {
+                // no-op
+            }
+
+            return null;
         }
 
         public async Task<JObject> PutHostConfigAsync(JObject content)


Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -205,9 +205,28 @@ private static bool IsFunctionEnabled
 
         public async Task<JObject> GetHostConfigAsync()
         {
-            return FileSystemHelpers.FileExists(HostJsonPath)
-                ? JObject.Parse(await FileSystemHelpers.ReadAllTextFromFileAsync(HostJsonPath))
-                : new JObject();
+            var host = await TryGetHostConfigAsync();
+            if (host == null)
+            {
+                throw new FileNotFoundException("Host.json is invalid");
```
davidebbo: `FileNotFoundException` doesn't seem like the right exception type here since the file is there.

davidebbo:
I think the pop up error you see in the portal actually doesn't come from this API, but comes from the portal calling `/admin/host/status` into the [function runtime](https://github.com/Azure/azure-webjobs-sdk-script) (which is not Kudu). So I don't think this fix in itself will change the behavior. Though I do agree the behavior should be improved.

develohpanda:
If you try with an invalid function.json file, the pop up shows the exact message from the exception thrown in GetFunctionConfigAsync() - I was using the same model for GetHostConfigAsync(). What would be a more appropriate exception to use in this case (or create a new exception type)?

develohpanda:
Also, hitting a broken function over an http request gives a stacktrace that leads to this method. Once at a computer again I can share more detail.

davidebbo:
Strange, I actually tried deploying a private Kudu build with your change and it did not affect the behavior for me (getting the same pop up with when having an empty host.json).

develohpanda:
Interesting - what if you change the message on GetFunctionConfigAsync and use an invalid function.json? Does that update popup message?

develohpanda:
Some further investigation - GetHostStatus() in this function is what routes the error. This returns 200 (hence portal not crashing), but the response is shown below, which is the exact exception thrown in the method changed in this pull request.

image
 I will investigate again in a couple of hours.

davidebbo:
Right, that's what I meant above when I said the issue comes from the portal calling `/admin/host/status` into the function runtime. So you agree this is not the right fix and we should close it?

develohpanda:
We can close it, but how the invalid json is handled here will affect how the ```/admin/host/status``` endpoint handles the response, right? Lets close PR but leave the issue open, and I will look deeper into the function runtime to really pinpoint the location. If it does end up at kudu again, I will set up another PR with some more information. Thanks for your help!

davidebbo:
No, I would not expect them to relate. `/admin/host/status` does its own thing, not using the Kudu API.

@ahmelsayed, do you know when this `/api/functions/config` Kudu API is even used?

ahmelsayed:
I don't think it's used anywhere now. The plan was to consume it in the portal and offer UI for the host global settings but we haven't that work yet.

davidebbo:
Even though the API is not used, may as well take this to improve its error handling for when we'll use it. I merged it.

develohpanda:
Hey @davidebbo, bugged merge? The PR still says closed with un-merged commits.

davidebbo:
That's because I rebased it over latest master commit.

PR #1986, opened 2016-05-04
Set ScriptHref to null if we can't find a script file for the function

ahmelsayed:


Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -296,17 +296,13 @@ private string GetFunctionScriptPath(string functionName, JObject functionInfo)
...
-                    // TODO: should this be an error?
-                    return functionPath;
-                }
                 return functionPrimary;
             }
         }
 
         private Uri FilePathToVfsUri(string filePath, bool isDirectory = false)
         {
+            if (string.IsNullOrEmpty(filePath)) return null;
```
suwatch: nit using {  and } for readability.

suwatch: is there a reason not not fix the caller instead (never passing null)?  callee should throw if null or empty etc.   easier to find bug that way.

davidebbo:
@ahmelsayed what's the status of this PR?

davidebbo:
@ahmelsayed Can you update status on this one?

ahmelsayed:
obsolete by #2008

PR #2002, opened 2016-05-13
Support nodejs deployment on mono

shrimpy:


Kudu.Core/Environment.cs
```diff
@@ -106,8 +106,8 @@ public class Environment : IEnvironment
             {
                 // in linux, rootPath is "/home", while .ssh folder need to under "/home/{user}", 
                 // and username and site name is always the same in actual deployment
-                string siteName = System.Environment.GetEnvironmentVariable("APP_POOL_ID");
-                siteName = NormalizeAppPoolId(siteName);
+                string siteName = System.Environment.GetEnvironmentVariable("WEBSITE_SITE_NAME");
```
shrimpy: Need to revert back to use "WEBSITE_SITE_NAME". On mono LWAS pass site name via "WEBSITE_SITE_NAME" environment valuable. We do not have "APP_POOL_ID"

Kudu.Core/Helpers/EnvironmentHelper.cs
```diff
@@ -0,0 +1,27 @@
...
+            {
+                int binIdx = binPath.LastIndexOf("Bin", StringComparison.Ordinal);
+                if (binIdx >= 0)
+                {
+                    string subStr = binPath.Substring(binIdx);
+                    // make sure file path is end with ".....Bin" or "....Bin/"
+                    if (subStr.Length < 5 && binPath.EndsWith(subStr, StringComparison.OrdinalIgnoreCase))
+                    {
+                        // real bin folder is lower case, but in mono, value is "Bin" instead of "bin"
+                        binPath = binPath.Substring(0, binIdx) + subStr.ToLowerInvariant();
```
davidebbo: Not sure I understand. If the folder name is 'bin', why is the value 'Bin'? Where is that coming from?

davidebbo: Oh, I see this was moved from another file. Anyway, it's Linux specific, so not a risk to Windows.

davidebbo:
:shipit:

PR #2003, opened 2016-05-13
Update to use 7zip to create Kudu.zip

shrimpy:


davidebbo:
One thing to watch out for is that public and private Kudu build different zips. Ideally we want to target both.

shrimpy:
@davidebbo right, private Kudu have a slightly different logic. I just check the code in private Kudu, they use the same code to create zip file while pass different parameters (which contains version number)

i think current PR is ok, i just need to adjust the code a little bit to when merge to private Kudu.

davidebbo:
:shipit:

PR #2011, opened 2016-05-18
Support building ASP.NET Core projects with MSBuild if we have a VS solution

ahmelsayed:
~~This is bending getting msbuild to work with `set NUGET_PACKAGES=%HOME%\.nuget`~~


davidebbo:
:shipit:

PR #2012, opened 2016-05-19
Return the run uri when invoking triggered WebJob

davidebbo:
Previously, starting a run returned an empty body. Now it returns the run object, letting the caller have the ID. I think the odd of someone making assumptions on the empty body is small enough that it is not a concern.

### DIFF
--- a/Kudu.Services/Jobs/JobsController.cs
+++ b/Kudu.Services/Jobs/JobsController.cs
@@ -182,7 +182,8 @@ namespace Kudu.Services.Jobs
 
                 // Return a 200 in the ARM case, otherwise a 202 can cause it to poll on /run, which we don't support
                 // For non-ARM, stay with the 202 to reduce potential impact of change
-                return Request.CreateResponse(ArmUtils.IsArmRequest(Request) ? HttpStatusCode.OK : HttpStatusCode.Accepted);
+                return Request.CreateResponse(ArmUtils.IsArmRequest(Request) ? HttpStatusCode.OK : HttpStatusCode.Accepted,
+                    ArmUtils.AddEnvelopeOnArmRequest(_triggeredJobsManager.GetLatestJobRun(jobName), Request));
             }
             catch (JobNotFoundException)
             {
--- a/Kudu.Services/Jobs/JobsController.cs
+++ b/Kudu.Services/Jobs/JobsController.cs
@@ -1,6 +1,7 @@
 ﻿using Kudu.Contracts;
 using Kudu.Contracts.Jobs;
 using Kudu.Contracts.Tracing;
+using Kudu.Core;
 using Kudu.Core.Hooks;
 using Kudu.Core.Infrastructure;
 using Kudu.Core.Jobs;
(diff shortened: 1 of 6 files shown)


Kudu.Services/Jobs/JobsController.cs
```diff
@@ -182,7 +182,8 @@ public HttpResponseMessage InvokeTriggeredJob(string jobName, string arguments =
 
                 // Return a 200 in the ARM case, otherwise a 202 can cause it to poll on /run, which we don't support
                 // For non-ARM, stay with the 202 to reduce potential impact of change
-                return Request.CreateResponse(ArmUtils.IsArmRequest(Request) ? HttpStatusCode.OK : HttpStatusCode.Accepted);
+                return Request.CreateResponse(ArmUtils.IsArmRequest(Request) ? HttpStatusCode.OK : HttpStatusCode.Accepted,
+                    ArmUtils.AddEnvelopeOnArmRequest(_triggeredJobsManager.GetLatestJobRun(jobName), Request));
```
suwatch: Get latest run is still approximate and could race with other triggers.   A more precise fix would be for this [API](https://github.com/projectkudu/kudu/blob/a6261ff068546cb9c1fd7c02e4fdb8f80d7fe338/Kudu.Core/Jobs/TriggeredJobRunner.cs#L44) to return logger.Id.   This id gets bubbled up.   It should not be many file changes I think.

davidebbo: That's actually how I started, and it was easy to bubble up the ID. But then I was not sure exactly what to return back to client:

- Some new minimal object that just has the ID?
- A full run object where I only populate the ID?

davidebbo: Oh wait, I could just use the ID to get the full run object :) Let me try that.

davidebbo:
@suwatch ok, updated to return the true run id. Please take a look.

Kudu.Services/Jobs/JobsController.cs
```diff
@@ -178,11 +178,12 @@ public HttpResponseMessage InvokeTriggeredJob(string jobName, string arguments =
...
             try
             {
-                _triggeredJobsManager.InvokeTriggeredJob(jobName, arguments, "External - " + Request.Headers.UserAgent);
+                string runId = _triggeredJobsManager.InvokeTriggeredJob(jobName, arguments, "External - " + Request.Headers.UserAgent);
 
                 // Return a 200 in the ARM case, otherwise a 202 can cause it to poll on /run, which we don't support
                 // For non-ARM, stay with the 202 to reduce potential impact of change
-                return Request.CreateResponse(ArmUtils.IsArmRequest(Request) ? HttpStatusCode.OK : HttpStatusCode.Accepted);
+                return Request.CreateResponse(ArmUtils.IsArmRequest(Request) ? HttpStatusCode.OK : HttpStatusCode.Accepted,
+                    ArmUtils.AddEnvelopeOnArmRequest(_triggeredJobsManager.GetJobRun(jobName, runId), Request));
```
suwatch: that would do.  Or if you want to avoid touching file system again, the location header representing a link url to specific run history would suffice.   This is what ARM async operation usually does.

davidebbo: Interesting. I'll look into that as well. Though note that in the ARM case, we're explicitely *not* doing the async pattern and we return 200. But presumably a location header would be harmless?

suwatch: With 200, I am pretty sure ARM would ignore location header.

davidebbo:
@suwatch alright, another round using just location header!

suwatch:
Looks good.  You may want to fix [this](https://github.com/projectkudu/kudu/blob/a6261ff068546cb9c1fd7c02e4fdb8f80d7fe338/Kudu.Client/Jobs/RemoteJobsManager.cs#L71) and enhance existing webjobtests to cover it.

davidebbo:
Ok, I added basic test coverage and pushed.

PR #2017, opened 2016-05-20
Missing request tracing on Mono

shrimpy:
**Issue:**
On mono, we will lost all request trace, e.g when perform a deployment we only see trace from kudu.exe, all **highlighted** trace will be missing since TraceModule is not invoked

**Solution:**
Register HttpModule via from code instead of Web.config

trace
        **├── 2016-05-20T21-44-17_abd311_002_GET_node3.git-info-refs_200_2s.xml
        ├── 2016-05-20T21-44-19_abd311_003_POST_node3.git-git-receive-pack_200_16s.xml**
        ├── 2016-05-20T21-44-23_abd311_001_kudu.exe_12s.xml
        **└── abd3118e3c27-04517727-330e-4ae0-ba00-71dcfdde45f9.txt**


davidebbo:
:shipit:

PR #2021, opened 2016-05-24
Do not cleanup lock file when running on Mono

shrimpy:


Kudu.Core/Infrastructure/LockFile.cs
```diff
@@ -229,13 +230,18 @@ public void Release()
         // it does not handled IOException due to 'file in used'.
         private void DeleteFileSafe()
```
shrimpy: technically, DeleteFileSafe is not needed
we just want to see the lock folder clean where there is no active lock.

davidebbo:
:shipit:

PR #2026, opened 2016-05-30
Fix WebJobs that have spaces in the file name

davidebbo:
WebJobs were broken if the file name has spaces. e.g. "Foo Bar.exe" or "Foo Bar.cmd". This came up [here](https://social.msdn.microsoft.com/Forums/en-US/ac6b901f-dd93-43fb-8caf-46961263e753/where-does-blank-blankspace-in-exe-name-come-from?forum=windowsazurewebsitespreview).


shrimpy:
:shipit:

Issue #1797, opened 2015-11-11
Can we ship dig.exe in Kudu?


snobu:
This is probably more a Web Apps thing, but here it goes anyway:

![image](https://cloud.githubusercontent.com/assets/6472374/11091046/bd8e5c40-887f-11e5-9cc1-e6eeb4906de3.png)

Works fine, we just need `dig.exe` and a bunch of .DLLs from this BIND9 package:
https://www.isc.org/downloads/file/bind-9-10-3/?version=win-32-bit

davidebbo:
Probably something that could go under `d:\devtools` rather than Kudu. There are various things in there already.

ahmelsayed:
we could also [put the tools there on the path](https://github.com/projectkudu/kudu/blob/bc3b8917f5d6baaf65af8bb9f630ac6d34bf9cdd/Kudu.Core/Deployment/Generator/ExternalCommandFactory.cs#L92-L109) for the console. Right now you have to know where they are or type the full path.

davidebbo:
Yes, we should do this. Note that it's in multiple folders, though `d:\devtools\sysinternals` is where the meat is.

snobu:
Is this a safe way to tell which libs dig.exe is loading?
![image](https://cloud.githubusercontent.com/assets/6472374/11100416/c8dcbc74-88b9-11e5-9a0e-d9a18080701e.png)

It does seem to work with just this short list:
```
libisc.dll
libdns.dll
libeay32.dll
libbind9.dll
libisccfg.dll
liblwres.dll
```

On Linux there's `strace` which does this:
![image](https://cloud.githubusercontent.com/assets/6472374/11100535/7f0a3d8c-88ba-11e5-97db-628b7748675d.png)

Is there an "strace" for Windows?
I know Process Explorer (Sysinternals) does just that but it's not CLI.

`listdlls.exe` seems very similar to what `strace` does but i can't get it to work.
`/accepteula` does nothing.

```cmd
D:\devtools\sysinternals>listdlls.exe
listdlls.exe

D:\devtools\sysinternals> 

D:\devtools\sysinternals>listdlls.exe /accepteula
listdlls.exe /accepteula

D:\devtools\sysinternals>
```

It does pop up an EULA on first run on my desktop.

davidebbo:
Kudu Process Explorer should give correct data. Note that many of the listed libs are system things, so you don't need to include them. You can get the full path to the lib if you click the More button.

snobu:
[...]

PR #2035, opened 2016-06-07
Put deployments\tools earlier in the path so it overrides other things

davidebbo:


snobu:
Unrelated, but still somewhat on topic, could you guys give `dig` another look to be included in the sandbox? nslookup is ..sub optimal.. I'm nice.

Portable binaries gathered here and tested OK in App Service sandbox:
https://github.com/snobu/minimaldig4kudu

davidebbo:
That's #1797. Sorry, that didn't get traction, maybe because adding things machine level is harder than adding them to Kudu :)

snobu:
That's okay. Nothing wrong with `nslookup` if you don't use it and use `nameresolve.exe` or PowerShell's `Resolve-DnsName` instead :) Those actually use the OS APIs to resolve so when you're in a VNET you're asking the right servers.

PR #2037, opened 2016-06-08
Kudu home page fixes

davidebbo:
Shaw runtime version + don't show uptime milliseconds

### DIFF
--- a/Kudu.Services.Web/Default.cshtml
+++ b/Kudu.Services.Web/Default.cshtml
@@ -76,7 +76,7 @@
             <strong>Site up time</strong>
         </div>
         <div>
-            @Kudu.Services.Web.Tracing.TraceModule.UpTime
+            @Kudu.Services.Web.Tracing.TraceModule.UpTime.ToString(@"dd\.hh\:mm\:ss")
         </div>
     </div>
     <div class="row">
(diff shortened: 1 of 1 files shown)


Kudu.Services.Web/Default.cshtml
```diff
@@ -1,11 +1,31 @@
...
+        Assembly assembly;
+        try
+        {
+            assembly = Assembly.Load("Microsoft.Web.Hosting, Version=7.1.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35");
+            var fileVersionInfo = System.Diagnostics.FileVersionInfo.GetVersionInfo(assembly.Location);
+            return fileVersionInfo.ProductVersion;
+        }
+        catch
+        {
+            return null;
```
shrimpy: shall we return empty string or a nicely error message instead of display null?

davidebbo: It's not an error. It's the normal behavior on non Azure machines.

shrimpy: if we return null. on the page, will it display "null" or empty? maybe should return empty string or "-" instead?

davidebbo: See below, the whole section is omitted when it's null.

shrimpy: ah. didn`t pay attention to "appServiceVersion" in line 11. thx

shrimpy:
:shipit:

Issue #2001, opened 2016-05-12
OneDrive deployment:The given path's format is not supported.


snobu:
When setting up OneDrive deployment in Ibiza, deployment is successful however a bunch of XML traces are written to `LogFiles\kudu\trace`.

###### D:\home\LogFiles\kudu\trace\2016-05-12T10-23-43_ea40a8_070_UnexpectedException.xml

```
<exception>
System.NotSupportedException: The given path's format is not supported.
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share)
   at System.IO.Abstractions.FileWrapper.Open(String path, FileMode mode, FileAccess access, FileShare share)
   at Kudu.Core.Infrastructure.FileSystemHelpers.AppendAllTextToFile(String path, String content)
   at Kudu.Core.Tracing.XmlTracer.WriteStartElement(String title, IDictionary`2 attribs)
</exception>
```

Any idea where in the code this is thrown?

shrimpy:
Have a feeling onedrive might send us encoded file path cause invalid file path issue.

You can start with https://github.com/projectkudu/kudu/blob/c3e58b6b958295538d1a4ed9e2f06cfcab8b91a5/Kudu.Services/FetchHelpers/OneDriveHelper.cs

snobu:
@shrimpy, one quick question:

https://github.com/projectkudu/kudu/blame/c3e58b6b958295538d1a4ed9e2f06cfcab8b91a5/Kudu.Services/FetchHelpers/OneDriveHelper.cs#L181

```
  "RepositoryUrl": "https://api.onedrive.com/v1.0/drive/special/approot:/kududev"
```

My `kududev` folder sits in my OneDrive under `Apps\Azure Web Apps\`
This `/special` OneDrive API ..there has to be something special going on with it. Is the magical path `Apps\Azure Web Apps` something well-known shared between the "Azure Web Sites" OneDrive app and OneDrive API and the path we ask for in the JSON payload is always relative to that one?

shrimpy:
[...]

PR #2040, opened 2016-06-10
Fix OneDrive tracing exceptions

snobu:
Fixes https://github.com/projectkudu/kudu/issues/2001

### DIFF
--- a/Kudu.Core/Infrastructure/StringExtensions.cs
+++ b/Kudu.Core/Infrastructure/StringExtensions.cs
@@ -34,11 +34,11 @@ namespace Kudu.Core
         // Does not check for reserved words.
         public static string ToSafeFileName(this string str)
         {
+            str = str.Trim()
+                     .Replace(' ', '-');
             foreach (char c in Path.GetInvalidFileNameChars())
             {
-                str = str.Trim()
-                         .Replace(c, '-')
-                         .Replace(' ', '-');
+                str = str.Replace(c, '-');
             }
             return str;
         }
--- a/Kudu.Core/Infrastructure/StringExtensions.cs
+++ b/Kudu.Core/Infrastructure/StringExtensions.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Globalization;
+using System.IO;
 
 namespace Kudu.Core
 {
@@ -27,6 +28,28 @@ namespace Kudu.Core
         public static string EscapeHashCharacter(this string str)
         {
             return str.Replace("#", Uri.EscapeDataString("#"));
+        }
+
+        // Returns a string you can safely use as valid Windows or Linux file name.
+        // Does not check for reserved words.
+        public static string ToSafeFileName(this string str)
+        {
+            foreach (char c in Path.GetInvalidFileNameChars())
+            {
+                str = str.Trim()
+                         .Replace(c, '-')
+                         .Replace(' ', '-');
+            }
+            return str;
+        }
+        
+        public static string Truncate(this string str, int maxLength)
+        {
+            if (string.IsNullOrEmpty(str))
+            {
+                return str;
+            }
+            return str.Length <= maxLength ? str : str.Substring(0, maxLength);
         }
     }
 }
--- a/Kudu.Services/FetchHelpers/OneDriveHelper.cs
+++ b/Kudu.Services/FetchHelpers/OneDriveHelper.cs
@@ -62,7 +62,7 @@ namespace Kudu.Services.FetchHelpers
             ChangeSet changeSet = null;
             string cursor = _settings.GetValue(CursorKey);
             ChangesResult changes = null;
-            // We truncate cursor value in trace filename but keep in unharmed in the trace content
+            // We truncate cursor value in filename but keep it unharmed in the trace content
             using (_tracer.Step("Getting delta changes with cursor: {0}...", cursor.Truncate(5)))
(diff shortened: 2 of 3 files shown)


snobu:
Is this something people usually write a test for?

Kudu.Core/Infrastructure/StringExtensions.cs
```diff
@@ -28,5 +29,27 @@ public static string EscapeHashCharacter(this string str)
...
+
+        // Returns a string you can safely use as valid Windows or Linux file name.
+        // Does not check for reserved words.
+        public static string ToSafeFileName(this string str)
+        {
+            foreach (char c in Path.GetInvalidFileNameChars())
+            {
+                str = str.Trim()
+                         .Replace(c, '-')
+                         .Replace(' ', '-');
```
davidebbo: Slightly odd to do the space replace inside the loop since it's invariant.

snobu: Right! I'll pull that out.

snobu: Fixed in https://github.com/projectkudu/kudu/pull/2040/commits/d1e78434b9aa38f708239990b1f87bd8081816b3

Kudu.Services/FetchHelpers/OneDriveHelper.cs
```diff
@@ -62,7 +62,7 @@ internal ILogger Logger
             ChangeSet changeSet = null;
             string cursor = _settings.GetValue(CursorKey);
             ChangesResult changes = null;
-            using (_tracer.Step("Getting delta changes with cursor: {0} ...", cursor))
+            using (_tracer.Step("Getting delta changes with cursor: {0}...", cursor.Truncate(5)))
```
shrimpy: why this is necessary?

Kudu.Core/Infrastructure/StringExtensions.cs
```diff
@@ -28,5 +29,27 @@ public static string EscapeHashCharacter(this string str)
         {
             return str.Replace("#", Uri.EscapeDataString("#"));
         }
+
+        // Returns a string you can safely use as valid Windows or Linux file name.
+        // Does not check for reserved words.
+        public static string ToSafeFileName(this string str)
+        {
+            str = str.Trim()
+                     .Replace(' ', '-');
```
shrimpy: why we filter out "space"? thought it should be a valid char for file name

Kudu.Services/FetchHelpers/OneDriveHelper.cs
snobu: That OneDrive __cursor__ thing is VERY long and breaks the UI like this:

image

Kudu.Core/Infrastructure/StringExtensions.cs
snobu: No other trace seems to contain a space so i put it in there for consistency.

traces

Kudu.Services/FetchHelpers/OneDriveHelper.cs
shrimpy: isn`t this line just the content of the log, not the file name?
````
using (_tracer.Step("Getting delta changes with cursor: {0}...", cursor.Truncate(5)))
````

snobu: No. Whole thing after _tracer.Step(  gets in the file name. That's what was producing the exception in the first place.

shrimpy: then what is the content inside that file? same as in title?

shrimpy: ah.. i see. this is the very first trace, that is why file name same as title content.
i would suggest to change to below, so that the cursor will go into content instead of showing up in file name.

````
            using (_tracer.Step("Getting delta changes"))
            using (_tracer.Step("cursor: {0} ...", cursor))
            {
                changes = await GetChanges(info.TargetChangeset.Id, info.AccessToken, info.RepositoryUrl, cursor);
            }
````

snobu: Here it is:
image

image

shrimpy: @snobu, do you mind update the code as I suggest above and give it a try to see what is the file name and content?Will be good to keep the full cursor value, since it will benefit for debugging.

Kudu.Core/Infrastructure/StringExtensions.cs
shrimpy: i see. thx for the update.

Kudu.Services/FetchHelpers/OneDriveHelper.cs
snobu: Good idea. I'll do that.

snobu: Fixed in https://github.com/projectkudu/kudu/pull/2040/commits/e0e95d54acae05896e456e1154660deca178613e
image

shrimpy:
@snobu, thanks again for your contribution, change has been rebased and merged into master https://github.com/projectkudu/kudu/commit/51c50b3d901e0e92eca847b22ef4190dedd18c3c

PR #2046, opened 2016-06-15
fix #2045 null ref in LogStreamManager

### LINKED ISSUE #2045: NullRef and Cyclic error in logstream
There are potential issues in [logstream](https://github.com/projectkudu/kudu/blob/master/Kudu.Services/Diagnostics/LogStreamManager.cs).
    
- NullRef issue.  When we initialize log stream, we only assign the `_logFiles` variable after we have [gather](https://github.com/projectkudu/kudu/blob/4835fe551457804d5a02055bb7ccc97b7b9f3609/Kudu.Services/Diagnostics/LogStreamManager.cs#L151) existing log file infos.   Since watcher has been started, there could be event raised in before the gathering finish.   This could lead to nullref.  The fix should be trivial by initializing `_logFiles` to an empty dictionary first.

```
   at System.IO.FileInfo.get_Length()
   at Kudu.Services.Performance.LogStreamManager.Initialize(String path)
2016-06-13T08:52:52      Error occurred, type: error, text: Too many changes at once in directory:D:\home\LogFiles., stackTrace:
2016-06-13T08:52:52      Error occurred, type: error, text: Too many changes at once in directory:D:\home\LogFiles., stackTrace:
2016-06-13T08:52:52      Error occurred, type: error, text: Too many changes at once in directory:D:\home\LogFiles., stackTrace:
2016-06-13T08:52:52    Error occurred, type: error, text: Object reference not set to an instance of an object., stackTrace:    at Kudu.Services.Performance.LogStreamManager.OnRenamed(Object sender, RenamedEventArgs e)
   at Kudu.Services.Performance.LogStreamManager.<>c__DisplayClass27_0`2.<DoSafeAction>b__0(T1 t1, T2 t2)
```
  - Cyclic error issue.  When file watcher encounters an error, we also trace that error.  The issue was, by tracing, this contributes to more file change notification.  Once we get into `Too many changes`, this becomes cyclic errors.  The fix would be to find a different way to trace this without causing extra change notification.

suwatch:


Kudu.Services/Diagnostics/LogStreamManager.cs
```diff
@@ -433,11 +435,6 @@ private void OnRenamed(object sender, RenamedEventArgs e)
 
         private void OnError(object sender, ErrorEventArgs e)
         {
-            using (_tracer.Step("FileSystemWatcher.OnError"))
-            {
-                _tracer.TraceError(e.GetException());
-            }
-
```
suwatch: For simplicity, just don't trace it.   Instead, send error to client straight (codes a few lines below).

davidebbo:
:shipit: I merged it.

Issue #1435, opened 2015-01-17
Source control deployment from a tag as well a branch


danlister:
It looks like Azure Websites can only deploy from a branch when linked to a repository from providers such as Bitbucket. My current workaround is to create the tag as a branch. I have tried the following with no success:
- `v1.0`
- `tags/v1.0`
- `refs/tags/v1.0`

This feature would be especially helpful when having multiple deployment slots within an Azure Website. Thanks for all your hard work :)

davidebbo:
Ok, here is the deal. The commands we run to do the fetch look like this:

    git remote add -t test external https://github.com/KuduApps/RepoWithMultipleBranches
    git fetch external --progress

But if test is a tag and not a branch, it fails with:

    fatal: Couldn't find remote ref refs/heads/test

I was hoping it would just work for either, but I guess that's not how `git remote add -t` works.

davidebbo:
If I try `refs/tags/firstcommit`, it fails with:

    fatal: Couldn't find remote ref refs/heads/refs/tags/firstcommit

We'll need to find a different git sequence to make that work. But note that we want to do a fetch from just that branch/tag, and not for the who remote repo (for perf/space reasons).

davidebbo:
We could just do:

    git remote add external https://github.com/KuduApps/RepoWithMultipleBranches
    git fetch external test

But I seem to recall we had some reason for using the quirkier `git remote add -t` syntax. I'm just not sure what it was...

alercunha:
Is there any news on this issue? I have a project in GitHub which creates a release as a tag not a branch. 

This would be a nice feature to add. When working with Prod environments it's key to have them pointing to a fixed tag. GitHub unfortunately doesn't give me an option to create a release as a branch.

davidebbo:
Sorry, short answer is that there hasn't been progress on this. I do hope we will get to it as it is a valid feature.

skaggsp:
I'd like to throw in my support for deploy-by-tag instead of by branch.

danlister:
Hey @davidebbo. I was just wondering if this feature request has been discussed/planned at all?

davidebbo:
Still no progress on this one, sorry! I would like to have it addressed, but getting it on the schedule with all the other priorities is the hard part.

davidebbo:
[...]

PR #2049, opened 2016-06-16
Added check to LibGit2SharpRepository to see if we are checking out a…

danlister:
… tag instead of a branch #1435 

To deploy a tag, the branch name passed to the function must start with "refs/tags/". From there, we can assume we want to deploy a tag instead of a branch. After testing locally, it seemed to work okay. I was able to deploy a master branch for example to Kudu, change the deployment source to a tag, create a tag locally, push to Kudu and see a successful deployment of the tag reference.

I also then created further commits on the master branch and pushed to Kudu. The tag reference was only ever deployed and not the further commits. My last test was to delete and re-create a tag, pushing to Kudu after. The deployment for the application deployed the updated tag reference.

I imagine I've not covered all bases but it's a start (I hope)

### DIFF
--- a/Kudu.Core/SourceControl/Git/LibGit2SharpRepository.cs
+++ b/Kudu.Core/SourceControl/Git/LibGit2SharpRepository.cs
@@ -226,25 +226,39 @@ echo $i > pushinfo
                         // This will only retrieve the "master"
                         repo.Network.Fetch(remote);
                     }
-
+                    
                     using (tracer.Step("LibGit2SharpRepository Update"))
                     {
-                        // Optionally set up the branch tracking configuration
-                        var trackedBranch = repo.Branches[trackedBranchName];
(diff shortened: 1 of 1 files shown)


davidebbo:
Didn't get to play with it much, but the change seems unlikely to regress anything given the conditional. We should add coverage though.

Thanks for your contrib!

davidebbo:
@danlister this is now deployed. Can you give it a try in Azure and make sure everything works as you expect? Thanks!

danlister:
@davidebbo It seems to work as expected. Below were my test steps:

- Navigate to a Web App on management.azure.com
- Change the "Branch to Deploy" configuration property to a tag
- Navigate back to portal.azure.com
- Re-deploy the App Service

davidebbo:
Perfect, thanks!

PR #2055, opened 2016-06-23
Remove support for pre-installed site extensions

davidebbo:
This was added for Kudu/Monaco/DaaS because there was no Ibiza UI, so we were relying on Site Extension UI to enable them.

But now, they all have first class support, so I think we should get rid of all this extra complexity.

When reviewing, add `?w=1` to the URL so it ignores blanks.

### DIFF
--- a/Kudu.Core/SiteExtensions/SiteExtensionManager.cs
+++ b/Kudu.Core/SiteExtensions/SiteExtensionManager.cs
@@ -52,42 +52,6 @@ namespace Kudu.Core.SiteExtensions
 
         private static readonly string _toBeDeletedDirectoryPath = System.Environment.ExpandEnvironmentVariables(@"%HOME%\data\Temp\SiteExtensions");
 
-        private static readonly Dictionary<string, SiteExtensionInfo> _preInstalledExtensionDictionary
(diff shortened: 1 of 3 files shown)


Kudu.Core/SiteExtensions/SiteExtensionManager.cs
```diff
@@ -202,82 +153,36 @@ public SiteExtensionManager(IContinuousJobsManager continuousJobManager, ITrigge
...
         }
 
         // <inheritdoc />
         public async Task<SiteExtensionInfo> GetLocalExtension(string id, bool checkLatest = true)
         {
             ITracer tracer = _traceFactory.GetTracer();
-            SiteExtensionInfo info = GetPreInstalledExtension(id);
-            if (info != null && info.ExtensionUrl != null)
+            UIPackageMetadata package = null;
+            using (tracer.Step("{0} is not a pre-installed package. Now querying from local repo.", id))
```
shrimpy: Comment "{0} is not a pre-installed package." no longer valid, please remove

shrimpy:
:shipit:  with one small comment

PR #2060, opened 2016-06-27
Hide disk space calculation on mono

shrimpy:
We are using below code to get freespace which is not existed in mono. 
On Mono we might have to use function from "Mono.Unix.DriveInfo namespace", which mean Kudu will take mono dependency in code. Need to spend more time for investigation to see if there is other way. 

As for now, I will just hide this feature on mono.

````
        [SuppressUnmanagedCodeSecurity]
        static class EnvironmentNativeMethods
        {
            [DllImport("kernel32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
            [return: MarshalAs(UnmanagedType.Bool)]
            public static extern bool GetDiskFreeSpaceEx(string path, out ulong freeBytes, out ulong totalBytes, out ulong diskFreeBytes);
        }
````


davidebbo:
:shipit:

PR #2061, opened 2016-06-27
logging failure to create status file to jobs log file

shobak101:


davidebbo:
:shipit:

shobak101:
turns out I can not do that because this class is the same class defining the log file operations, and log message is virtual, so can not call it from ctor, will replace the log to log in kudu trace

shobak101:
changes squashed and merged to master

PR #2063, opened 2016-06-28
Set NUGET_PACKAGES on start

davidebbo:
This helps scenarios that use Core do the right thing by default


ahmelsayed:
looks good. I opened https://github.com/projectkudu/KuduScript/pull/50 to go with it.

PR #2065, opened 2016-06-28
Exclude excluded Functions in sync triggers

mathewc:
See corresponding Functions commit here: https://github.com/Azure/azure-webjobs-sdk-script/commit/3e7309f0fbc131ba49e7b61ec04bda493be07619

### DIFF
--- a/Kudu.Core.Test/Functions/FunctionManagerTests.cs
+++ b/Kudu.Core.Test/Functions/FunctionManagerTests.cs
@@ -1,0 +1,103 @@
+﻿using System.Collections.Generic;
+using Kudu.Core.Functions;
+using Kudu.Core.Tracing;
+using Newtonsoft.Json.Linq;
+using Xunit;
+
+namespace Kudu.Core.Test.Functions
+{
+    public class FunctionManagerTests
+    {
+        [Theory]
+        [InlineData(null, null, false)]
+        [InlineData(false, null, false)]
+        [InlineData(null, false, false)]
+        [InlineData(false, false, false)]
+        [InlineData(null, true, true)]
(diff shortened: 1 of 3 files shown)


Kudu.Core.Test/Functions/FunctionManagerTests.cs
```diff
@@ -0,0 +1,81 @@
+﻿using System.Collections.Generic;
+using Kudu.Core.Functions;
+using Kudu.Core.Tracing;
+using Newtonsoft.Json.Linq;
+using Xunit;
+
+namespace Kudu.Core.Test.Functions
+{
+    public class FunctionManagerTests
```
mathewc: There were no unit tests for any of this code, so I'm adding some here. We really need to make sure we're adding tests with code.

davidebbo:
Note that @ahmelsayed is moving all the logic to the runtime. When that happen, I would expect this code to no longer be relevant. Please connect with him.

mathewc:
Yes, but for now, the code lives here, and we need it to be updated. Right @ahmelsayed ?

davidebbo:
I think that depends when we switch over. If it's around the corner, why do duplicate work? Anyway, I'll let you come to agreement with Ahmed on this one. I don't mind changes going into Kudu for transition if that's what it takes.

mathewc:
Yeah, I want to check this PR in. The runtime change is already in, so I don't want to block on keeping Kudu in sync.

davidebbo:
Ok. Looking is most of the diff because of variable renames rather than directly related? When can look at it directly in person if you want to speed it up.

davidebbo:
:shipit:

PR #2078, opened 2016-07-05
Fix #2073, Only trigger autoswap when deploymnet is success.

### LINKED ISSUE #2073: Autoswap of failed deployment
Hi, I've found that there's already been similar issue - #1813, however the described problem is still happening to me. Even if the deployment fails, the swap operation is performed and the version of latest successful deploy before the erroneous deploy is used. According to our logs, we're observing the following scenario:

1. Production slot running V2, Staging slot running V1.
2. Github deployment of V3 (erroneous) to staging slot.
3. Deployment fails, swap operation is triggered.
4. Production slot running V1, staging slot running V2.

This is how it looks like in Azure portal:

image

Compared to successful deployment, the `Running post deployment command(s)...` and `Deployment successful.` items of the deployment details are missing.

shrimpy:


suwatch:
:shipit:

PR #2095, opened 2016-08-03
Adding aria-live='polite' to jquery.console to support screen readers

ahmelsayed:
This is a very safe change. Just adding some metadata on the console divs to tell screen readers that they are alive and they should be monitored  Here is [MDN docs on how it works](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions)


davidebbo:
Since you're asking politely, :shipit:

PR #2107, opened 2016-08-23
added a random custom delay before deployment

watashiSHUN:
sample ".deployment" file configuration should look like this 
```
[config]
SCM_MAX_RANDOM_START_DELAY_IN_SECONDS = 40
```

sample trace file should look like this
```
<step title="custom max delay is set to 40s" date="2016-08-22T19:07:00.512" />
<!-- duration: 0ms -->
<step title="deployment deplay applied, Start sleeping for 29s" date="2016-08-22T19:07:00.512" />
<!-- duration: 29022ms -->
```

### DIFF
--- a/Kudu.Core/Deployment/DeploymentManager.cs
+++ b/Kudu.Core/Deployment/DeploymentManager.cs
@@ -23,7 +23,7 @@ namespace Kudu.Core.Deployment
     {
         public readonly static string DeploymentScriptFileName = OSDetector.IsOnWindows() ? "deploy.cmd" : "deploy.sh";
 
-        private static readonly Random _random = new Random( String.IsNullOrEmpty(System.Environment.GetEnvironmentVariable(Constants.HttpHost))? DateTime.Now.Millisecond : System.Environment.GetEnvironmentVariable(Constants.HttpHost).GetHashCode());
+        private static readonly Random _random = new Random(Guid.NewGuid().GetHashCode());
 
         private readonly ISiteBuilderFactory _builderFactory;
         private readonly IEnvironment _environment;
--- a/Kudu.Core/Deployment/DeploymentManager.cs
+++ b/Kudu.Core/Deployment/DeploymentManager.cs
@@ -23,7 +23,7 @@ namespace Kudu.Core.Deployment
     {
         public readonly static string DeploymentScriptFileName = OSDetector.IsOnWindows() ? "deploy.cmd" : "deploy.sh";
 
-        private static readonly Random _random = new Random();
+        private static readonly Random _random = new Random(System.Environment.GetEnvironmentVariable(Constants.HttpHost).GetHashCode());
 
         private readonly ISiteBuilderFactory _builderFactory;
         private readonly IEnvironment _environment;
@@ -555,8 +555,7 @@ namespace Kudu.Core.Deployment
                     else
                     {
                         tracer.Trace("{0} is set to {1}s", SettingsKeys.MaxRandomDelayInSec, maxDelay);
-                        Random rdn = new Random();
-                        int gap = rdn.Next(maxDelay);
+                        int gap = _random.Next(maxDelay);
                         using (tracer.Step("Randomization applied to {0}, Start sleeping for {1}s", maxDelay, gap))
                         {
                             logger.Log(Resources.Log_DelayingBeforeDeployment, gap);
--- a/Kudu.Core/Deployment/DeploymentManager.cs
+++ b/Kudu.Core/Deployment/DeploymentManager.cs
@@ -23,7 +23,7 @@ namespace Kudu.Core.Deployment
     {
         public readonly static string DeploymentScriptFileName = OSDetector.IsOnWindows() ? "deploy.cmd" : "deploy.sh";
 
-        private static readonly Random _random = new Random(System.Environment.GetEnvironmentVariable(Constants.HttpHost).GetHashCode());
+        private static readonly Random _random = new Random( String.IsNullOrEmpty(System.Environment.GetEnvironmentVariable(Constants.HttpHost))? DateTime.Now.Millisecond : System.Environment.GetEnvironmentVariable(Constants.HttpHost).GetHashCode());
 
         private readonly ISiteBuilderFactory _builderFactory;
         private readonly IEnvironment _environment;


Kudu.Core/Deployment/DeploymentManager.cs
```diff
@@ -544,6 +544,31 @@ private DeployResult GetResult(string id, string activeDeploymentId, bool isDepl
...
+                    int gap = rdn.Next(maxDelay);
+                    using (tracer.Step("deployment deplay applied, Start sleeping for {0}s",gap))
+                    {
+                        System.Threading.Thread.Sleep(new TimeSpan(0,0,gap));
+                    }
+                    /*
+                    var cpuCounter = new PerformanceCounter("Processor", "% Processor Time", "_Total");
+                    int usage = (int)cpuCounter.NextValue();
+                    _tracer.Trace("current cpu usage is {0}%", usage);
+                     "Access to the registry key &apos;Global&apos; is denied.*/
```
davidebbo: Clean out this commented logic.

Kudu.Core/Deployment/DeploymentManager.cs
```diff
@@ -544,6 +544,31 @@ private DeployResult GetResult(string id, string activeDeploymentId, bool isDepl
...
+                        tracer.Trace("entered invalid SCM_MAX_RANDOM_START_DELAY_IN_SECONDS value, use the default: 60");
+                    }else
+                    {
+                        tracer.Trace("custom max delay is set to {0}s", maxDelay);
+                    }
+                    Random rdn = new Random();
+                    int gap = rdn.Next(maxDelay);
+                    using (tracer.Step("deployment deplay applied, Start sleeping for {0}s",gap))
+                    {
+                        System.Threading.Thread.Sleep(new TimeSpan(0,0,gap));
```
shrimpy: Change to use "Task.Delay" would be better, not blocking the thread

Kudu.Core/Deployment/DeploymentManager.cs
```diff
@@ -544,6 +544,31 @@ private DeployResult GetResult(string id, string activeDeploymentId, bool isDepl
                 string repositoryRoot = _environment.RepositoryPath;
                 var perDeploymentSettings = DeploymentSettingsManager.BuildPerDeploymentSettingsManager(repositoryRoot, _settings);
 
+                string delayMaxInStr = perDeploymentSettings.GetValue(SettingsKeys.MaxRandomDelayInSec, false);
+                if (delayMaxInStr != null) {
+                    int maxDelay;
+                    if(!Int32.TryParse(delayMaxInStr, out maxDelay) || maxDelay < 0)
```
shrimpy: if given number is not valid, might just treat it the same as "null" ?

watashiSHUN: I was thinking of using the default value for user input error: 
```
SCM_MAX_RANDOM_START_DELAY_IN_SECONDS = true
SCM_MAX_RANDOM_START_DELAY_IN_SECONDS = 10s
SCM_MAX_RANDOM_START_DELAY_IN_SECONDS = 123.45
...
```
in otherwords, user still wants the delay, but the format is off
if it is the convention to treat it as "null", I can surely change it back

shrimpy: just personal suggestion no convention i think.  it will be easier to explain and follow if we only accept the exact format, either number or timespam format. 

otherwise user who use this feature might get confuse. e.g value was set to "120s", so was expecting more than 1 minutes delay for some occasion, however current code convert to 60 seconds max internally. if user do not read the log they won`t notice this changes. but if there is no delay, user will come back and revisit what they did wrong.

davidebbo: I'd say it's fine to ignore the setting if it has an incorrect value. It's an advanced setting, and it's not that hard to get it right. I believe that's what we do with other similar settings.

Kudu.Contracts/Settings/SettingsKeys.cs
```diff
@@ -36,5 +36,6 @@ public static class SettingsKeys
         public const string SiteExtensionsFeedUrl = "SCM_SITEEXTENSIONS_FEED_URL";
         public const string DisableDeploymentOnPush = "SCM_DISABLE_DEPLOY_ON_PUSH";
         public const string TouchWebConfigAfterDeployment = "SCM_TOUCH_WEBCONFIG_AFTER_DEPLOYMENT";
+        public const string MaxRandomDelayInSec = "SCM_MAX_RANDOM_START_DELAY_IN_SECONDS";
```
davidebbo: Leave out the `_IN_SECONDS` part. Our other settings like `SCM_COMMAND_IDLE_TIMEOUT` and `SCM_LOGSTREAM_TIMEOUT` are all in seconds, and don't have that explicitly in their name.

Kudu.Core/Deployment/DeploymentManager.cs
```diff
@@ -544,6 +544,25 @@ private DeployResult GetResult(string id, string activeDeploymentId, bool isDepl
...
+                    {
+                        tracer.Trace("Invalid {0} value, expect a positive integer, received {1}", SettingsKeys.MaxRandomDelayInSec, delayMaxInStr);
+                    }else
```
davidebbo: `else` on its own line. That should be the default formatting option in VS.

Kudu.Core/Deployment/DeploymentManager.cs
```diff
@@ -544,6 +544,25 @@ private DeployResult GetResult(string id, string activeDeploymentId, bool isDepl
...
+                    {
+                        tracer.Trace("{0} is set to {1}s", SettingsKeys.MaxRandomDelayInSec, maxDelay);
+                        Random rdn = new Random();
+                        int gap = rdn.Next(maxDelay);
+                        using (tracer.Step("Randomization applied to {0}, Start sleeping for {1}s", maxDelay,gap))
```
davidebbo: Here we also need to log something to the user visible deployment log. `tracer` is a separate thing that is more for internals, and customer doesn't normally see it. You probably just need to call `innerLogger.Log("...")`.

davidebbo:
For test coverage, you can easily modify the `DeploymentApis` test. After the `RunAsync` call, add a line like this:

    await appManager.SettingsManager.SetValue(SettingsKeys.MaxRandomDelayInSec, "10");

And in the `KuduAssert.VerifyUrl(result.LogUrl, cred);` line, add an extra param with the string to check for.

Kudu.Core/Deployment/DeploymentManager.cs
```diff
@@ -544,6 +544,27 @@ private DeployResult GetResult(string id, string activeDeploymentId, bool isDepl
...
+                    }
+                    else
+                    {
+                        tracer.Trace("{0} is set to {1}s", SettingsKeys.MaxRandomDelayInSec, maxDelay);
+                        Random rdn = new Random();
+                        int gap = rdn.Next(maxDelay);
+                        using (tracer.Step("Randomization applied to {0}, Start sleeping for {1}s", maxDelay, gap))
+                        {
+                            logger.Log(Resources.Log_DelayingBeforeDeployment, gap);
+                            await Task.Delay(new TimeSpan(0, 0, gap));
```
ahmelsayed: I think `TimeSpan.FromSeconds(gap)` would be clearer here.

davidebbo:
:shipit:

Kudu.Core/Deployment/DeploymentManager.cs
```diff
@@ -544,6 +544,27 @@ private DeployResult GetResult(string id, string activeDeploymentId, bool isDepl
                 string repositoryRoot = _environment.RepositoryPath;
                 var perDeploymentSettings = DeploymentSettingsManager.BuildPerDeploymentSettingsManager(repositoryRoot, _settings);
 
+                string delayMaxInStr = perDeploymentSettings.GetValue(SettingsKeys.MaxRandomDelayInSec, false);
+                if (delayMaxInStr != null)
```
ahmelsayed: we tend to use `string.IsNullOrEmpty(delayMaxInStr)` in kudu for string null checks. I know this is semantically different from `== null` but it's almost always what you want anyway.

Kudu.Core/Deployment/DeploymentManager.cs
```diff
@@ -544,6 +544,27 @@ private DeployResult GetResult(string id, string activeDeploymentId, bool isDepl
                 string repositoryRoot = _environment.RepositoryPath;
                 var perDeploymentSettings = DeploymentSettingsManager.BuildPerDeploymentSettingsManager(repositoryRoot, _settings);
 
+                string delayMaxInStr = perDeploymentSettings.GetValue(SettingsKeys.MaxRandomDelayInSec, false);
```
ahmelsayed: consider using a named parameter for the boolean type

watashiSHUN: the Boolean is useful when there are several setting providers, it is not very relevant in this case....
and there are code that use plain true/false as an argument:`MarkStatusComplete(currentStatus, success: false);`

Kudu.Core/Deployment/DeploymentManager.cs
```diff
@@ -544,8 +544,8 @@ private DeployResult GetResult(string id, string activeDeploymentId, bool isDepl
                 string repositoryRoot = _environment.RepositoryPath;
                 var perDeploymentSettings = DeploymentSettingsManager.BuildPerDeploymentSettingsManager(repositoryRoot, _settings);
 
-                string delayMaxInStr = perDeploymentSettings.GetValue(SettingsKeys.MaxRandomDelayInSec, false);
-                if (delayMaxInStr != null)
+                string delayMaxInStr = perDeploymentSettings.GetValue(SettingsKeys.MaxRandomDelayInSec, true);
```
ahmelsayed: why did it become true :) just curious. btw, by named parameter I meant
```csharp
perDeploymentSettings.GetValue(SettingsKeys.MaxRandomDelayInSec, onlyPerSite: true);
```
VS will tell you that it's not needed and you should consider removing it, but we like them in kudu. They make reading code easier :)

watashiSHUN: I was reading its implementation, it appears that if I put `true`, less conditions and less code will run, but since I don't really care about its value, I am going back to default

davidebbo:
:shipit:

shrimpy:
@watashiSHUN shall we close this PR or you are still working on it? I saw fix has already been merged into master. Same for https://github.com/projectkudu/kudu/pull/2113

davidebbo:
@watashiSHUN can this one be closed now?

PR #2112, opened 2016-08-25
Update Kudu to consume container username.

shrimpy:
- There was a change in Azure and need to update Kudu accordingly, hence this change.
 - Tested in private deployment


davidebbo:
:shipit:

PR #2118, opened 2016-08-29
Set LastWriteTime of files extracted from zip

duncansmart:


davidebbo:
I have this strange deja vu that this is a change that we talked about doing way back and somehow ended up not doing, but I can't remember why.

I think one concern is that zip files don't have concept of time zone, and that without knowing the time zone it was created it in, you can't reliably infer their true UTC time.

Still, on the surface it would seem fairly harmless.

duncansmart:
Well you could argue that it's better than nothing! :-)

davidebbo:
Yes, I agree. Unless there was something else that came up that I'm not remembering. Anyway, let's give this a try and see what blows up :)

Issue #2135, opened 2016-09-14
Fail to deploy from GitHub Enterprise


helaili:
Trying to deploy an Azure web app from GitHub Enterprise following [those steps](https://nsamteladze.wordpress.com/2015/07/19/continuous-deployment-from-github-enterprise-repository-to-azure-web-app/). I've connected through the Kudu console and confirmed my deploy key is valid by manually triggering a git clone and checking the that GitHub reports usage of this key. 

Note that before I did the manual clone, the key was never used besides trying to trigger multiple deployments. 

Looking at `api/vfs/LogFiles/kudu/trace/2016-09-14T10-36-55_f61f4d_532_Background_POST_deploy_1s.xml`, I get the message `Request failed with status code: 401` 

Full log : 

```

<step title="BackgroundTrace" date="2016-09-14T10:36:55.330" instance="f61f4d" url="/deploy" method="POST">
<step title="Performing fetch based deployment" date="2016-09-14T10:36:55.346">
<step title="Creating temporary deployment" date="2016-09-14T10:36:55.346"/>
<!--  duration: 47ms  -->
<step title="LibGit2SharpRepository Fetch" date="2016-09-14T10:36:55.581"/>
<!--  duration: 176ms  -->
</step>
<!--  duration: 477ms  -->
[...]

PR #2145, opened 2016-09-22
add logic to detect public repo with private GHE, issue #2135

watashiSHUN:
There is a case that I don't handle:

if a **public GHE** with a **public repo** POST to kudu
kudu will try to use `https`, however it is possible for the **public GHE** to have a self signed certificate (which is the case for my test personal GHE) the default behavior of `git` is
> fatal: unable to access https://123.123.123.123/GHErepo/ SSL certificate problem: self signed certificate

you can disable certificate check with a git command, but I don't know if we should commit to that


davidebbo:
Yeah, let's not worry about the self-signed cert case. I think in a real deployment companies would never use that as it's bad security practice. If it ever comes up, we'll tackle.

Kudu.Services/ServiceHookHandlers/GithubHandler.cs
```diff
@@ -1,10 +1,29 @@
...
+using Kudu.Core;
+using System;
+using Kudu.Core.Infrastructure;
+using System.IO;
 
 namespace Kudu.Services.ServiceHookHandlers
 {
     public class GitHubHandler : GitHubCompatHandler
     {
+        private IEnvironment environmentInject { get; set; } = null;
```
davidebbo: Casing: `EnvironmentInject`. However, for something like this we typically use private fields and not props. You can just call it `_environment`.

Kudu.Services/ServiceHookHandlers/GithubHandler.cs
```diff
@@ -17,6 +36,25 @@ public override DeployAction TryParseDeploymentInfo(HttpRequestBase request, JOb
             return DeployAction.UnknownPayload;
         }
 
+        protected override string determineSecurityProtocol(JObject repository)
+        {
+            string repositoryUrl = repository.Value<string>("url");
+            bool isPrivate = repository.Value<bool>("private");
+            if (isPrivate)
+            {
+                repositoryUrl = repository.Value<string>("ssh_url");
```
davidebbo: Why use different logic in base and derived to get the ssh url? I'd suggest adding a `GetSSHUrl` helper in the base and use it from all 3 places.

Kudu.Services/ServiceHookHandlers/GithubHandler.cs
```diff
@@ -17,6 +36,25 @@ public override DeployAction TryParseDeploymentInfo(HttpRequestBase request, JOb
...
+            }
+            else if (!String.Equals(new Uri(repositoryUrl).Host, "github.com"))
```
davidebbo: `StringComparison.OrdinalIgnoreCase`. I'd expect a code analysis error without it. Try running build.cmd from command line (at repo root).

Kudu.Services/ServiceHookHandlers/GithubHandler.cs
```diff
@@ -17,6 +36,25 @@ public override DeployAction TryParseDeploymentInfo(HttpRequestBase request, JOb
...
+            {
+                // github enterprise user, we don't really need to check for null...since only test will invoke githubhandler
```
davidebbo: This comment is not so meaningful to me. I would instead describe what the code is doing so someone reading it knows why we're doing this, e.g.

    // If we find an SSH key and it's GitHub Enterprise and not github.com, use ssh. This is needed if
    // the repo is public but the GHE instance is not

Kudu.Services/ServiceHookHandlers/GithubHandler.cs
watashiSHUN: are you talking about writing 1 `GetSSHUrl` with implementation in the base class, or just a template method and have the children classes override them? either way its not going to be straight forward

[gitlab](https://github.com/projectkudu/kudu/blob/master/Kudu.Services/ServiceHookHandlers/GitlabHqHandler.cs#L82-L83),[codebase](https://github.com/projectkudu/kudu/blob/fd0062a0888bf54ad6f14f2d761e5947a57de983/Kudu.Services/ServiceHookHandlers/CodebaseHqHandler.cs#L38-L46) all have different logics to determine "private" repo ... actually they override `GetDeploymentInfo` 

so I can't have them override `GetSSHUrl` without refactor their `GetDeploymentInfo` function , I keep the old code because we also use `githubcompathandler` as one of the parser, I am not sure if there are cases that will match it...Moreover inside `githubhandler` I can check for "github.com" with confidence...I am thinking it's safer to use this white list kind of strategy (less assumption)

watashiSHUN:
tested with github, gitlab and codebase.

PR #2151, opened 2016-09-27
fix #2121 and improve job initializing logging

### LINKED ISSUE #2121: Don't hash WebJobs name in logs
There was never a great reason to do this, and it makes investigations harder.

suwatch:
The fix contains ..
- log jobName as is (previously hash value)
- log job change event (initializing and initialized) - to help debug the missing job changed notification.

Sample log below,
image

### DIFF
--- a/Kudu.Core/Jobs/JobsFileWatcher.cs
+++ b/Kudu.Core/Jobs/JobsFileWatcher.cs
@@ -35,6 +35,7 @@ namespace Kudu.Core.Jobs
         private readonly Action<string> _onJobChanged;
         private readonly string _filter;
         private readonly Func<bool, IEnumerable<string>> _listJobNames;
+        private readonly string _jobType;
 
         public JobsFileWatcher(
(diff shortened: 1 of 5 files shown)


Kudu.Core/Jobs/JobsFileWatcher.cs
```diff
@@ -83,9 +86,13 @@ private void OnMakeChanges(object state)
                 try
                 {
                     _onJobChanged(updatedJobName);
+
+                    _analytics.JobEvent(updatedJobName, "Job initialized successfully", _jobType, String.Empty);
                 }
                 catch (Exception ex)
                 {
+                    _analytics.JobEvent(updatedJobName, "Job initialized failed", _jobType, ex.ToString());
```
davidebbo: "Job initialization failed" if we care about grammar :)

suwatch: Fixed.

PR #2152, opened 2016-09-28
fix regression in tracer due to deployment background

suwatch:
Tracer created in ctor tied to a HTTP request.   We need to use a background tracer for background work.


davidebbo:
:shipit:

davidebbo:
I merged it in.

PR #2157, opened 2016-09-28
Edit webAPI to retrieve/create function/master keys

watashiSHUN:
upgrade to .net 4.6 in order to use the dataprotection nuget package
Azure/azure-webjobs-sdk-script#662

### DIFF
--- a/Kudu.Core/Functions/FunctionManager.cs
+++ b/Kudu.Core/Functions/FunctionManager.cs
@@ -227,9 +227,9 @@ namespace Kudu.Core.Functions
                     }
                     return keyOp.GenerateKeyObject(key, name);
                 }
-                catch (Exception)
+                catch (IOException)
                 {
-                    // fallback to read key files
+                    // fallback to read key files only when the file already existed, failed to get the handler
                 }
             }
 


Kudu.Core/packages.config
```diff
@@ -18,6 +33,22 @@
...
+  <package id="System.Linq" version="4.1.0" targetFramework="net46" />
+  <package id="System.Linq.Expressions" version="4.1.0" targetFramework="net46" />
+  <package id="System.Reflection" version="4.1.0" targetFramework="net46" />
+  <package id="System.Resources.ResourceManager" version="4.0.1" targetFramework="net46" />
+  <package id="System.Runtime" version="4.1.0" targetFramework="net46" />
+  <package id="System.Runtime.Extensions" version="4.1.0" targetFramework="net46" />
+  <package id="System.Runtime.InteropServices" version="4.1.0" targetFramework="net46" />
+  <package id="System.Text.Encodings.Web" version="4.0.0" targetFramework="net46" />
+  <package id="System.Threading" version="4.0.11" targetFramework="net46" />
+  <package id="System.Threading.Tasks" version="4.0.11" targetFramework="net46" />
```
davidebbo: We need to make sure that we don't regress the Kudu startup time, so let's verify that.

Kudu.Contracts/Functions/MasterKeyOperation.cs
```diff
@@ -0,0 +1,46 @@
...
+using Newtonsoft.Json.Linq;
+using System;
+
+namespace Kudu.Core.Functions
+{
+    public class MasterKeyOperation : IKeyOperation
+    {
+        public JObject GenerateKeyJson(string masterKey)
+        {
+            JObject hostJson = JObject.Parse($"{{\"masterKey\":{{\"name\":\"master\",\"value\":\"{masterKey}\",\"encrypted\": true }}}}");
```
fabiocav: When generating the host secrets JSON, let's make sure the `functionKeys` array is present here, even if empty. We may also want to check the previous behavior, but today, when the host generates the host level secret file, it creates a function secret as well, so we should probably do the same here.

watashiSHUN: there is no previous behavior in kudu since it did not return keys in "host.json". looking at the [CanSerialize()](https://github.com/fabiocav/azure-webjobs-sdk-script/blob/31cd7632375a65b290c08dd80b7a5a9fbdfb71e3/src/WebJobs.Script.WebHost/Security/ScriptSecretSerializerV1.cs#L30-L33) function, it seems that if "host.json" does not have a "functionKeys" JArray, it is considered as malformatted...but we don't check whether if the JArray is empty, do you want to create an empty "functionKeys" JArray or an encrypted key name "default"

watashiSHUN:
Do we have to worry about the concurrency issue during key creation?

Kudu.Contracts/Functions/MasterKeyOperation.cs
fabiocav: Let's create the list with a newly generated key named "default"

watashiSHUN: should we check for "functionkey" in version 0 as well for consistency? [ScriptSecretSerializerV0.CanSerialize()](https://github.com/fabiocav/azure-webjobs-sdk-script/blob/31cd7632375a65b290c08dd80b7a5a9fbdfb71e3/src/WebJobs.Script.WebHost/Security/ScriptSecretSerializerV0.cs#L81-L83)

Kudu.Core/packages.config
watashiSHUN: I tested against kudu verison 58.50929.2438

the cold start function key retrieval is respectively **7000ms : 11500ms,**
this change will adds roughly 4 sec

also the following assemblies were not loaded, I guess some other functionalities of `dataprotection` depends are them, but they are not being used here

```
 +  <package id="System.Collections.Concurrent" version="4.0.12" targetFramework="net46"/>
 +  <package id="System.Collections" version="4.0.11" targetFramework="net46" />
 +  <package id="System.ComponentModel" version="4.0.1" targetFramework="net46" />
 +  <package id="System.Diagnostics.Debug" version="4.0.11" targetFramework="net46" />
 +  <package id="System.Globalization" version="4.0.11" targetFramework="net46" />
 +  <package id="System.IO" version="4.1.0" targetFramework="net46" />
 +  <package id="System.Linq" version="4.1.0" targetFramework="net46" />
 +  <package id="System.Linq.Expressions" version="4.1.0" targetFramework="net46" />
 +  <package id="System.Reflection" version="4.1.0" targetFramework="net46" />
 +  <package id="System.Runtime.Extensions" version="4.1.0" targetFramework="net46" />      
 +  <package id="System.Threading" version="4.0.11" targetFramework="net46" />
 +  <package id="System.Threading.Tasks" version="4.0.11" targetFramework="net46" />
```

watashiSHUN:
with and without this change, `/api/settings` both return in 5.16s on cold start

davidebbo:
Yep, now that assembly loads is optimized, I there is no more perf concern.

Kudu.Contracts/Functions/MasterKeyJsonOps.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        public int GetKeyNumbers()
+        {
+            return 2;
+        }
+
+        public string GenerateKeyJson(Tuple<string,string>[] keyPair, out string unencryptedKey)
+        {
+            unencryptedKey = keyPair[0].Item1;
+            return JObject.Parse($"{{\"masterKey\":{{\"name\":\"master\",\"value\":\"{keyPair[0].Item2}\",\"encrypted\": true }},\"functionKeys\":[{{\"name\": \"default\",\"value\": \"{keyPair[1].Item2}\",\"encrypted\": true}}]}}").ToString(Formatting.Indented);
```
fabiocav: Same commtne about the string -> JObject -> string here

Kudu.Contracts/Functions/FunctionSecretsJsonOps.cs
```diff
@@ -0,0 +1,66 @@
...
+        public int GetKeyNumbers()
+        {
+            return 1;
+        }
+
+        // have the schema related info enclosed in this class
+        public string GenerateKeyJson(Tuple<string,string>[] keyPair, out string unencryptedKey)
+        {
+            unencryptedKey = keyPair[0].Item1;
+            return JObject.Parse($"{{\"keys\":[{{\"name\":\"default\",\"value\":\"{keyPair[0].Item2}\",\"encrypted\": true }}]}}").ToString(Formatting.Indented);
```
fabiocav: Should be able to simplify this so we don't go from string to JObject just to get a formatted string.

Kudu.Contracts/Functions/FunctionSecretsJsonOps.cs
```diff
@@ -0,0 +1,66 @@
...
+                // all parse issue ==> format exception
+                throw new FormatException("Invalid secrets file format.");
+            }
+            isEncrypted = false;
+            return null;
+        }
+
+        public FunctionSecrets GenerateKeyObject(string functionKey, string functionName)
+        {
+            return new FunctionSecrets { Key = functionKey, TriggerUrl = String.Format(@"https://{0}/api/{1}?code={2}", System.Environment.GetEnvironmentVariable("WEBSITE_HOSTNAME") ?? "localhost", functionName, functionKey) };
```
fabiocav: nit: break this line up so it's not so long

Kudu.Contracts/Functions/IKeyJsonOps.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Newtonsoft.Json.Linq;
+using System;
+
+namespace Kudu.Core.Functions
+{
+    public interface IKeyJsonOps<T>
+    {
+        int GetKeyNumbers();
```
fabiocav: This would probably be more appropriately named as something like `RequiredKeyCount` and as a property, instead of a method.

Kudu.Contracts/Functions/IKeyJsonOps.cs
```diff
@@ -0,0 +1,15 @@
...
+        string GenerateKeyJson(Tuple<string,string>[] keyPairs, out string unencryptedKey);
+        // return IsEncrypted
```
fabiocav: nit: remove comment

Kudu.Core/Infrastructure/SecurityUtility.cs
```diff
@@ -1,21 +1,43 @@
...
+            var protector = DataProtectionProvider.CreateAzureDataProtector().CreateProtector(_defaultProtectorStr);
+            for (int i = 0; i < number; i++)
+            {
+                string unencryptedKey = GenerateSecretString();
+                unencryptedToEncryptedKeyPair[i] = new Tuple<string, string>(unencryptedKey, protector.Protect(unencryptedKey));
+            }
+            return unencryptedToEncryptedKeyPair;
+        }
+
+        private const string _defaultProtectorStr = "function-secrets";
```
fabiocav: nit: Move this to the top of the class and name the const `DefaultProtectorPurpose`

Kudu.Core/Infrastructure/SecurityUtility.cs
```diff
@@ -1,21 +1,43 @@
...
+
+        public static string DecryptSecretString(string content)
+        {
+            return DataProtectionProvider.CreateAzureDataProtector().CreateProtector(_defaultProtectorStr).Unprotect(content);
```
fabiocav: nit: break this into multiple lines for readability?

ahmelsayed:
@watashiSHUN what's the status of this PR?

watashiSHUN:
@ahmelsayed most of the changes are done, but I think it is still under review

fabiocav:
@ahmelsayed looks like the nuget source is not set. We're pulling the data protection package from our myget feed. We can publish to nuget if that'd make things easier.

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -191,34 +199,71 @@ internal static JArray GetTriggers(IEnumerable<FunctionEnvelope> functions, ITra
...
+                    return keyOp.GenerateKeyObject(key, name);
+                }
+                catch (Exception)
+                {
+                    // fallback to read key files
+                }
             }
-            else
+
+            for (int timeOut = 5; timeOut >= 0; timeOut--)
```
davidebbo: The retry loop should only be around the reading of the file. This way, if other random things blow up (e.g. DecryptSecretString), it won't look like we're dealing with a file being held issue.

Kudu.Core/Infrastructure/SecurityUtility.cs
```diff
@@ -1,21 +1,44 @@
...
+            var protector = DataProtectionProvider.CreateAzureDataProtector().CreateProtector(DefaultProtectorPurpose);
+            for (int i = 0; i < number; i++)
+            {
+                string unencryptedKey = GenerateSecretString();
+                unencryptedToEncryptedKeyPair[i] = new Tuple<string, string>(unencryptedKey, protector.Protect(unencryptedKey));
+            }
+            return unencryptedToEncryptedKeyPair;
+        }
+
+        public static string DecryptSecretString(string content)
```
davidebbo: `private`

Kudu.Core/app.config
```diff
@@ -1,19 +1,23 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <runtime>
     <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
 			<dependentAssembly>
-				<assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30AD4FE6B2A6AEED" culture="neutral"/>
-				<bindingRedirect oldVersion="0.0.0.0-6.0.0.0" newVersion="6.0.0.0"/>
+				<assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30AD4FE6B2A6AEED" culture="neutral" />
+				<bindingRedirect oldVersion="0.0.0.0-6.0.0.0" newVersion="6.0.0.0" />
```
davidebbo: Restore formatting to avoid big diff. Same in a few other configs.

Kudu.Contracts/Functions/FunctionSecretsJsonOps.cs
```diff
@@ -0,0 +1,77 @@
...
+    {
+        public int RequireKeyCount()
+        {
+            return 1;
+        }
+
+        // have the schema related info enclosed in this class
+        public string GenerateKeyUglyJson(Tuple<string, string>[] keyPair, string functionRt, out string unencryptedKey)
+        {
+            unencryptedKey = keyPair[0].Item1;
```
watashiSHUN: Kudu api called to create master/function keys only work for the default format (ie we don't take any parameters), names are hard coded in the string

Kudu.Contracts/Functions/FunctionSecretsJsonOps.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        public string GetKeyInString(string json, out bool isEncrypted)
+        {
+            try
+            {
+                JObject hostJson = JObject.Parse(json);
+                if (hostJson["key"]?.Type == JTokenType.String)
+                {
+                    isEncrypted = false;
```
watashiSHUN: there are two assignment statement, but it is only set once (if old format, we default to not encrypted, else we read the `encrypted` field in the new key format)

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -191,34 +199,71 @@ internal static JArray GetTriggers(IEnumerable<FunctionEnvelope> functions, ITra
...
+                {
+                    using (var fileStream = FileSystemHelpers.OpenFile(keyPath, FileMode.CreateNew, FileAccess.Write, FileShare.None))
+                    // will fail if file exists, prevent reading prematurely
+                    // getting the lock early so no redundant work is being done
+                    {
+                        string jsonContent = keyOp.GenerateKeyUglyJson(SecurityUtility.GenerateSecretStringsKeyPair(keyOp.RequireKeyCount()), FunctionSiteExtensionVersion, out key);
+                        using (var sw = new StringWriter())
+                        using (var sr = new System.IO.StringReader(jsonContent))
+                        {
+                            new JsonTextWriter(sw) { Formatting = Formatting.Indented }.WriteToken(new JsonTextReader(sr));
```
watashiSHUN: from previous code review, we moved away from String->jsonobject->String
json

Kudu.Core/Infrastructure/SecurityUtility.cs
watashiSHUN: it is called [here](https://github.com/projectkudu/kudu/pull/2157/files/f43631a94f7ae08732bffc0c2efaf6acde95d98a#diff-770a930cac2c5747cde5d2d9940eeca2R245)

Kudu.Contracts/Functions/IKeyJsonOps.cs
```diff
@@ -0,0 +1,18 @@
...
+using System;
+
+namespace Kudu.Core.Functions
+{
+    public interface IKeyJsonOps<T>
+    {
+        int RequireKeyCount();
+        
+        // key generation is based on run time
+        string GenerateKeyUglyJson(Tuple<string,string>[] keyPairs, string functionRt, out string unencryptedKey);
```
watashiSHUN: renamed, it was called ulglyjson as opposed to prettyjson (notepad++ json after formatting), uglyjson return a json string in with no whitespace or newlines

Kudu.Contracts/Functions/FunctionSecretsJsonOps.cs
davidebbo: I just mean as a small reflector to have one fewer line. Understood that only one is executed.

Kudu.Core/Infrastructure/SecurityUtility.cs
davidebbo: My bad

Kudu.Core/app.config
```diff
@@ -1,19 +1,23 @@
-<?xml version="1.0" encoding="utf-8"?>
```
davidebbo: Why is this file showing as a full-fill diff? Maybe some unwanted formatting issue?

watashiSHUN:
Azure/azure-websites-security#2

ahmelsayed:
open a new pull request

watashiSHUN:
the exception change is checkin already #2232

Issue #2174, opened 2016-10-09
Post Deployment Hook should include slot information


motasem13hamad:
Using post deployment hook, is there any way to do the following because it seems that it does not differentiate between a deployment slot and instances and only carry the site name,

For example, 
If there is a website that has two Slots, And each slot has 3 instances, can we have something to notify us on which instance, slot that has new deployment.

```
{
  "id": "2bdc42572263361fef1a3334c562be57dfb06c27",
  "status": "success",
  "statusText": "",
  "authorEmail": "test@test.com",
  "author": "John Doe",
  "message": "Epic new feature!",
  "progress": "",
  "deployer": "",
  "receivedTime": "2015-06-11T09:58:46.9983824Z",
  "startTime": "2015-06-11T10:39:02.1322211Z",
  "endTime": "2015-06-11T10:39:07.555094Z",
  "lastSuccessEndTime": "2015-06-11T10:39:07.555094Z",
  "complete": true,
  **"siteName": "azure-dummy-site"**
}
```

davidebbo:
I can see how knowing the slot name could be interesting, but I don't think the specific instance could ever be relevant, since all instances share the same file system. Can you clarify why you would want the instance info?

For the slot, we don't really have the slot name, but we do have something related: the hostname. e.g. the hostname would be something like "mysite.azurewebsites.net" or "mysite-myslot.azurewebsites.net". So we could include that in the payload.

motasem13hamad:
Thank you,  i guess its related to speed of deployment to the websites instances,  but it should be no huge difference in deployment speed between instances in the same slot,
Is there an ETA for the slot name to be added to this api?

Thanks

Get Outlook for Android<https://aka.ms/ghei36>



On Mon, Oct 10, 2016 at 3:29 AM +0300, "David Ebbo" <notifications@github.com<mailto:notifications@github.com>> wrote:


I can see how knowing the slot name could be interesting, but I don't think the specific instance could ever be relevant, since all instances share the same file system. Can you clarify why you would want the instance info?

For the slot, we don't really have the slot name, but we do have something related: the hostname. e.g. the hostname would be something like "mysite.azurewebsites.net" or "mysite-myslot.azurewebsites.net". So we could include that in the payload.

-

davidebbo:
Sorry, I'm not following what you describe relating to speed of deployment. Can you clarify?

For the slot name, please see my previous comment about potentially using the hostname instead. Will they work in your scenario?

[...]

PR #2185, opened 2016-10-18
added hostName to webhook payload, feature #2174

watashiSHUN:
sample output looks like this:
capture123


davidebbo:
:shipit: 

And please update https://github.com/projectkudu/kudu/wiki/Web-hooks#hook-request-fired-by-the-kudu-service with new field

PR #2189, opened 2016-10-20
Make sure to delete any leftover artifacts by a deleted function with…

ahmelsayed:
… the same name as a function being created. Fixes projectkudu/AzureFunctionsPortal#581


davidebbo:
:shipit:

PR #2199, opened 2016-10-27
kudu errors out on function creation if the name is not valid, 

watashiSHUN:
Azure/azure-webjobs-sdk-script#832

the example in logic apps looks like this 
capture


davidebbo:
:shipit:

PR #2201, opened 2016-10-28
suppress thread aborted by kudu event

suwatch:
To reduce noises in Kudu log (file system as well as internal jarvis/kusto).   This helps us identify the actual unexpected errors more efficiently.


davidebbo:
Nice, that'll definitely help in some cases! :shipit:

PR #2207, opened 2016-11-01
fix #2205 is_in_place caused cyclic change in dotnet job

### LINKED ISSUE #2205: is_in_place and UpdateAppConfigAddTraceListeners caused cyclic of Change notification
For C# with *.exe.config, we are injecting appSettings, connectionStrings as well as traceListeners ([here](https://github.com/projectkudu/kudu/blob/master/Kudu.Core/Jobs/BaseJobRunner.cs#L377) and [here](https://github.com/projectkudu/kudu/blob/master/Kudu.Core/Jobs/BaseJobRunner.cs#L444)) to the job *.exe.config file.   If is_in_place is true, this causes an infinite cyclic of change notification, job startup and inject config.

We discovered this when cleaning up the Kudu error.   For internal reference, [site info] (https://observer/sites/id/13596425).   

The suggested fix would be to, after update the config, only save the file if content has changed.   This would cause the one extra change notification - trade off with simplicity.

suwatch:
Trivial per discussion on https://github.com/projectkudu/kudu/issues/2205

### DIFF
--- a/Kudu.Core/Jobs/BaseJobRunner.cs
+++ b/Kudu.Core/Jobs/BaseJobRunner.cs
@@ -135,7 +135,6 @@ namespace Kudu.Core.Jobs
             {
                 _inPlaceWorkingDirectory = JobBinariesPath;
                 SafeKillAllRunningJobInstances(logger);
-                UpdateAppConfigs(WorkingDirectory, _analytics);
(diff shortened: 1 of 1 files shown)


davidebbo:
:shipit:

PR #2210, opened 2016-11-02
fix #2208 auto swap on redeploy

### LINKED ISSUE #2208: Redeploy (PUT /api/deployment/{id}) does not perform autoswap
Issue was reported by customer.   We simply forgot to add codes to do it for [this scenario](https://github.com/projectkudu/kudu/blob/master/Kudu.Services/Deployment/DeploymentController.cs#L155).   

See [This](https://github.com/projectkudu/kudu/blob/master/Kudu.Console/Program.cs#L162-L171) for the correct implementation.

suwatch:


Kudu.Console/Program.cs
```diff
@@ -162,12 +162,15 @@ private static int Main(string[] args)
                     deploymentManager.DeployAsync(gitRepository, changeSet: null, deployer: deployer, clean: false)
                         .Wait();
 
-                    string branch = settingsManager.GetBranch();
-                    ChangeSet changeSet = gitRepository.GetChangeSet(branch);
-                    IDeploymentStatusFile statusFile = deploymentStatusManager.Open(changeSet.Id);
-                    if (statusFile != null && statusFile.Status == DeployStatus.Success)
+                    if (autoSwapHander.IsAutoSwapEnabled())
```
davidebbo: This part is just an optimization and not a functional change, right? Since `HandleAutoSwap` already checks if enabled.

Kudu.Services/Deployment/DeploymentController.cs
```diff
@@ -153,6 +157,22 @@ public void Delete(string id)
...
+                        // auto-swap
+                        if (_autoSwapHandler.IsAutoSwapEnabled())
+                        {
+                            if (changeSet == null)
+                            {
+                                var targetBranch = _settings.GetBranch();
+                                changeSet = repository.GetChangeSet(targetBranch);
+                            }
+
+                            IDeploymentStatusFile statusFile = _status.Open(changeSet.Id);
```
davidebbo: Might consider a `IDeploymentStatusManager.IsSuccessfulChanset` extension method, since same logic in 3 places. But not a big deal.

davidebbo:
Minor comment. :shipit:

Kudu.Console/Program.cs
suwatch: Correct.  Below open/read files.  We should avoid that if no autoswap.

```csharp
string branch = settingsManager.GetBranch();
ChangeSet changeSet = gitRepository.GetChangeSet(branch);
IDeploymentStatusFile statusFile = deploymentStatusManager.Open(changeSet.Id);
```

PR #2232, opened 2016-11-16
dotnet core preview2 preview3 support in kudu

watashiSHUN:
the detection is through `<TargetFramework>netcoreapp1.0</TargetFramework>` in "csproj"
here is the kuduscript changes: projectkudu/KuduScript#56

### DIFF
--- a/Kudu.Core/Deployment/DeploymentHelper.cs
+++ b/Kudu.Core/Deployment/DeploymentHelper.cs
@@ -10,7 +10,7 @@ namespace Kudu.Core.Deployment
 {
     public static class DeploymentHelper
     {
-        private static readonly string[] _projectFileExtensions = new[] { ".csproj", ".vbproj", ".fsproj" };
+        private static readonly string[] _projectFileExtensions = new[] { ".csproj", ".vbproj", ".fsproj", ".xproj" };
 
         public static readonly string[] ProjectFileLookup = _projectFileExtensions.Select(p => "*" + p).ToArray();
 
@@ -23,12 +23,6 @@ namespace Kudu.Core.Deployment
         public static bool IsProject(string path)
         {
             return _projectFileExtensions.Any(extension => path.EndsWith(extension, StringComparison.OrdinalIgnoreCase));
-        }
-
-        public static bool IsDeployableProject(string path)
-        {
-            return IsProject(path) &&
-                   (VsHelper.IsWap(path) || VsHelper.IsExecutableProject(path));
         }
 
         public static bool IsDefaultWebRootContent(string webroot)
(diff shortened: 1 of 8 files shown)


davidebbo:
So I just gave this a try on my test repo, and it didn't seem to work (deployed as plain site). The repo is: https://github.com/davidebbo-test/MvcCoreVS2017RC. Right now it has a custom script, so yank that and commit before trying. Note that it uses a class library. Maybe that's throwing it off?

watashiSHUN:
@davidebbo , how did you get your `.sln` file? when I created a dotnet core web app using VS 2017, here is my `projectTypeGuid` in `.sln` [FAE04EC0-301F-11D3-BF4B-00C04F79EFBC](https://github.com/watashiSHUN/PREVIEW3sln/blob/master/AspNetCoreNewToolingRC.sln#L8) compared to yours [9A19103F-16F7-4668-BE54-9A1E7A4F7556](https://github.com/davidebbo-test/MvcCoreVS2017RC/blob/master/MvcCoreVS2017RC.sln#L8).

Mine is identified as a [c# project](http://www.codeproject.com/Reference/720512/List-of-Visual-Studio-Project-Type-GUIDs) and I couldn't find yours...it seems to be related to some [xproj standard](https://github.com/333fred/roslyn-project-system/blob/7581e869512f86de7a051121d59f57892dea2386/src/Microsoft.VisualStudio.ProjectSystem.CSharp.VS/Packaging/CSharpProjectSystemPackage.cs#L32-L34)....my theory is that **9A19103F-16F7-4668-BE54-9A1E7A4F7556** is mean for dotnet (when they are developing preview2) projects, but it was not very well documented so in kudu...instead of using it we just checked for "xproj" file extension

if your `.sln` is created by VS2017, we need to talk to these people and ask about the which one is the **standard dotnet core web app projectTypeGuid**

davidebbo:
I just created it with VS 2017 RC (the official build). What's odd is that my **library** project actually has that `FAE04EC0-301F-11D3-BF4B-00C04F79EFBC` guid, but my web project has a different one.

davidebbo:
Oh, check [this commit](https://github.com/davidebbo-test/MvcCoreVS2017RC/commit/2a32ccb23d976a69dc2e1882bf8ee8c6bc553225) out, which is when I added the library. My web project had the `FAE04EC0` guid, but then after adding the library, the library got that guid and the web project got a new one. Weird stuff!

davidebbo:
I did another test, and oddly this time both the Web project and Library ended up with the `FAE04EC0` guid. Though even that would likely cause your code to fail since it couldn't differentiate between the two.

I think we can't avoid looking in the .csproj to detect which are actually web projects. And even that is non-trivial, as checking for `<TargetFramework>netcoreapp1.0</TargetFramework>` is not enough. e.g. if you create a plain core Console app, it has that as well. We might need to check for other things like presence of a `web.config`.

In the end, it's a heuristic, and no heuristic is perfect, but we can probably go a long way. And when the heuristic fail, user just needs to set `PROJECT` explicitely.

watashiSHUN:
tested against two preview2 projects
[multiple webprojects with solution](https://github.com/watashiSHUN/PREVIEW2web-websln)
[multiple webprojects without solution](https://github.com/watashiSHUN/PREVIEW2web-webnsln)

PR #2247, opened 2016-12-02
update test to latest commit (sync with private kudu)

watashiSHUN:
these two tests are updated with "global.json" in kudu private


davidebbo:
:shipit:

PR #2258, opened 2016-12-07
fix #2254 404 issue when get job history

### LINKED ISSUE #2254: 404 reading job status because status file was being written.
The issue was the race condition between reading/writing the status log.     I took the last couple samples of 404 from one site.   The common pattern was triggered jobs was started, ran for 20s-30s then stopped.   Status file was updated when started and stopped.  While stopping, there was a request to get status.    The read status tried to read the being-written status file and failed.   We have a retry loop but it was too short interval.   The improvement options are ...

- read and write status could conflict and should have longer try interval.
- invent some sort of file lock so they won't conflict in the first place.

Given the occurrence, I suggest we do just first option and see how it goes.

suwatch:
Two key fixes.

- more retries since lack of file lock (bigger fix).
- ensure unexpected exception go thru Kudu table.

### DIFF
--- a/Kudu.Core/Jobs/JobLogger.cs
+++ b/Kudu.Core/Jobs/JobLogger.cs
@@ -23,6 +23,8 @@ namespace Kudu.Core.Jobs
 
         protected ITraceFactory TraceFactory { get; private set; }
 
+        protected IAnalytics Analytics { get; private set; }
+
         protected string InstanceId { get; private set; }
 
         private string _statusFilePath;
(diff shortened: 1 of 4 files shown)


Kudu.Core/Jobs/JobLogger.cs
```diff
@@ -102,15 +105,16 @@ protected string GetStatusFilePath()
                     return null;
                 }
 
+                // since we don't have proper lock on file, we are more forgiving is retry (10 times 250 ms interval).
```
davidebbo: "we are more forgiving is retry" doesn't sound right :)

suwatch: fix typo, thank.

suwatch:
merged

PR #2271, opened 2016-12-15
change siteExtension src to nuget.org and add tag filter for search

watashiSHUN:
TODO, wait until nuget sever support searchTerm `PackageType:"azuresiteextension"`
TODO, get a more unique name for the tag, ie `azuresiteextension`

python script for migration can be found [here](https://github.com/watashiSHUN/nugetMigrate)


davidebbo:
So this is work in progress and not ready to go in, right?

Issue #2294, opened 2017-01-11
Output of .NET Core CLI is not displayed


fabiocav:
When running the "dotnet" (by simply typing `dotnet` in the diagnostics console, CMD or PowerShell), the output briefly flashes on the screen and disappears.

davidebbo:
They seem to do some weird CR/LF things that confuse Kudu console. You can even repro like this:

- Run: `dotnet > out.txt`
- Run: `type out.txt`

If you look at out.txt, instead of having `0x0d 0x0a` line ending sequences, it has doubled Carriage Returns before the Line Feed: `0x0d 0x0d 0x0a`. And for some reason that makes Kudu console behave funny.

@ahmelsayed I'm thinking it's something we could fix on Kudu side? We could report as dotnet issue, but the fact is that it appears to be harmless in a regular console, so in that sense it is our bug.

ahmelsayed:
we interpret alone `\r` as a line clear command, so when we get `\r\r\n` we interpret that as clear the line, then move to the line below. This was done because that's how curl outputs download progress indicator.
I think that was just nice to have, but I'm not sure why few tools output things in that format. So far I know of typescript compiler and now dotnet.exe

davidebbo:
Cool. Did you test both curl and dotnet, to make sure there is no noticeable regressions?

ahmelsayed:
yep, tested curl with 
`curl -o file.zip https://functionsbay.blob.core.windows.net/public/1.0.0.beta.9/Azure.Functions.Cli.zip`
and dotnet just by listing the help.

watashiSHUN:
it looks like the [fix](https://github.com/projectkudu/kudu/blob/42317ae0116b689be06b305d495267d0654b0707/Kudu.Services.Web/Content/Scripts/KuduExecV2.js#L228-L230) is checking `(current line ends with "\r" && last line equals "\r\n")`, if the condition is `true`, we will **not** skip current line
This will not work for `tsc --version` since it output like this:
![capture](https://cloud.githubusercontent.com/assets/6531400/21967306/0edbc42c-db39-11e6-9c0d-d5d9ba2da4e0.JPG)

I used another command line exe to test (`youtube-dl`...a little tool that I used to download video, of cause with the progress indication) since `curl` **does not hit the line being changed**: rather than outputing `"progress 10%\r" "progress 11%\r"` it is doing this `"progress 10%" "\r" "progress 11%" "\r"`

PR #2296, opened 2017-01-15
fix kudu console not displaying tsc --version

watashiSHUN:
#2294


ahmelsayed:
If you made sure curl, dotnet and any other standard command (like dir or something) didn't regress feel free to merge it in.

davidebbo:
@watashiSHUN is the issue you describe something that was always there, or something that got regressed with @ahmelsayed's recent change?

ahmelsayed:
@davidebbo It was always there. tsc output looked like dotnet output where it'll print the line then clears it, I thought it'd be fixed with my fix for dotnet, but I never tested it so maybe not.

PR #2304, opened 2017-01-20
fix index out of range in ProcessExtensions.EnvToDictionary

suwatch:
This is to address the exception below.   There are many potential index-out-of-length issues where we incremented `i` and used it without checking whether it was out of length.    I started fixing by adding checking here and there and the codes become hard to maintain and understand.   I decided to rewrite and simplify it.   I have done a backward compatibility comparison between old vs. new implementations against various processes to make sure no regression. 

```csharp
at Kudu.Core.Infrastructure.ProcessExtensions.EnvToDictionary(Byte[] env)
   at Kudu.Core.Infrastructure.ProcessExtensions.GetEnvironmentVariablesCore(IntPtr hProcess)
   at Kudu.Core.Infrastructure.ProcessExtensions.GetEnvironmentVariables(Process process)
   at Kudu.Services.Performance.ProcessController.SafeGetValue[T](Func`1 func, T defaultValue)
```

### DIFF
--- a/Kudu.Core/Infrastructure/ProcessExtensions.cs
+++ b/Kudu.Core/Infrastructure/ProcessExtensions.cs
@@ -549,37 +549,17 @@ namespace Kudu.Core.Infrastructure
                 }
             }
 
-            char[] environmentCharArray = Encoding.Unicode.GetChars(env, 0, len);
-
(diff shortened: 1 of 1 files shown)


ahmelsayed:
It does look much simpler this way. 
:shipit:

Issue #2278, opened 2016-12-19
nuget searches in SiteExtension tests fail with internal server error


watashiSHUN:
Here's the error message:
```
"Message": "An error has occurred.",
"ExceptionMessage": "Collection was modified after the enumerator was instantiated.",
"ExceptionType": "System.InvalidOperationException",
"StackTrace": "   at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)	
   at System.Collections.Generic.Stack`1.Enumerator.MoveNext()			
   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source)	 	 	
   at System.ComponentModel.Composition.Hosting.ImportEngine.InPrerequisiteLoop()	 	 	
   at System.ComponentModel.Composition.Hosting.ImportEngine.TrySatisfyImportsStateMachine(PartManager partManager, ComposablePart part)	 	 	
   at System.ComponentModel.Composition.Hosting.ImportEngine.TrySatisfyImports(PartManager partManager, ComposablePart part, Boolean shouldTrackImports)	 	 	
   at System.ComponentModel.Composition.Hosting.ImportEngine.SatisfyImports(ComposablePart part)	 	 	
   at System.ComponentModel.Composition.Hosting.CompositionServices.GetExportedValueFromComposedPart(ImportEngine engine, ComposablePart part, ExportDefinition definition)	 	 	
   at System.ComponentModel.Composition.Hosting.CatalogExportProvider.GetExportedValue(CatalogPart part, ExportDefinition export, Boolean isSharedPart)	 	 	
   at System.ComponentModel.Composition.Primitives.Export.get_Value()	 	 	
   at System.ComponentModel.Composition.ExportServices.GetCastedExportedValue[T](Export export)	 	 	
   at System.Lazy`1.CreateValue()	 	 	
   at System.Lazy`1.LazyInitValue()	 	 	
   at NuGet.Client.SourceRepository.<GetResourceAsync>d__8`1.MoveNext()	 
```

since we didn't introduce any new code, this is likely an issue with [GetResourceAsync()](https://github.com/projectkudu/kudu/blob/2ea953e532d0a44afe291dc4b9e370df9fc51686/Kudu.Core/SiteExtensions/FeedExtensions.cs#L250)

[...]

PR #2305, opened 2017-01-21
fix issue #2278

watashiSHUN:
Kudu.FunctionalTests.SiteExtensions.SiteExtensionApiFacts.SiteExtensionV2AndV3FeedTests
Kudu.FunctionalTests.SiteExtensions.SiteExtensionApiFacts.SiteExtensionGetArmTest
Kudu.FunctionalTests.SiteExtensions.SiteExtensionApiFacts.SiteExtensionBasicTests
....

which all calls [GetRemoteExtensions](https://github.com/projectkudu/kudu/blob/4593ba51254eaca21fe7f2274718a995239980c2/Kudu.FunctionalTests/SiteExtensions/SiteExtensionApiFacts.cs#L96) should no longer appear in teamcity test errors

### DIFF
--- a/Kudu.Core/SiteExtensions/FeedExtensions.cs
+++ b/Kudu.Core/SiteExtensions/FeedExtensions.cs
@@ -57,6 +57,23 @@ namespace Kudu.Core.SiteExtensions
 
             return latestPackage;
         }
+        
+        // TODO refactor last function, after kudu private test all passes
+        public static async Task<UIPackageMetadata> GetLatestPackageById(this UIMetadataResource metadataResource, string packageId, bool includePrerelease = true, bool includeUnlisted = false)
+        {
+            UIPackageMetadata latestPackage = null;
+            IEnumerable<UIPackageMetadata> packages = await metadataResource.GetMetadata(packageId, includePrerelease, includeUnlisted, token: CancellationToken.None);
+            foreach (var p in packages)
+            {
+                if (latestPackage == null ||
+                    latestPackage.Identity.Version < p.Identity.Version)
+                {
+                    latestPackage = p;
+                }
+            }
+
+            return latestPackage;
+        }
 
         /// <summary>
         /// <para>Query source repository for a package base on given package id and version</para>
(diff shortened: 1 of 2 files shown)


Kudu.Core/SiteExtensions/FeedExtensions.cs
```diff
@@ -57,6 +57,23 @@ public static class FeedExtensions
 
             return latestPackage;
         }
+        
+        // TODO refactor last function, after kudu private test all passes
```
davidebbo: Not sure I understand why we need this method that looks identical to previous. Can't the previous one just call this one?

Kudu.Core/SiteExtensions/FeedExtensions.cs
```diff
@@ -245,7 +262,7 @@ public static class FeedExtensions
             }
         }
 
-        private static async Task<T> GetResourceAndValidateAsync<T>(this SourceRepository srcRepo) where T : class, INuGetResource
+        public static async Task<T> GetResourceAndValidateAsync<T>(this SourceRepository srcRepo) where T : class, INuGetResource
```
davidebbo: `internal` is preferred here

Kudu.Core/SiteExtensions/SiteExtensionManager.cs
```diff
@@ -87,13 +87,16 @@ public SiteExtensionManager(IContinuousJobsManager continuousJobManager, ITrigge
                             .OrderByDescending(p => p.DownloadCount);
             }
 
-            using (tracer.Step("Convert search result to SiteExtensionInfos"))
+            using (tracer.Step("Convert search result to SiteExtensionInfos with max concurrent requests: {0}", System.Environment.ProcessorCount))
             {
+                // GetResourceAsync is not thread safe
```
davidebbo: Is the comment referring to `GetResourceAndValidateAsync` or a different method called `GetResourceAsync`?

watashiSHUN: public static async Task<T> GetResourceAndValidateAsync<T>(this SourceRepository srcRepo) where T : class, INuGetResource
        {
            var resource = await srcRepo.GetResourceAsync<T>();

GetResourceAsync is the nuget method being called in the end

PR #2308, opened 2017-01-24
Add routingTrigger when a site has routing enabled. 

safihamid:
This is needed for scale decisions.


davidebbo:
Looks good. Is current scale controller already able to handle this (or ignore gracefully)? Kudu will likely be deployed in next day or two, so need to make sure it doesn't cause problem with current scale controller bits.

safihamid:
Yes it will be handled by current scale controller. so we are fine.

davidebbo:
Rebased and merged.

Issue #2309, opened 2017-01-25
Linux: Kudu uses wrong path separator in $PATH


nickwalkmsft:
On startup, Kudu prepends a bunch of computed paths to the PATH environment variable. It uses ";" as the path separator, which works on Windows but is broken on Linux.

https://github.com/projectkudu/kudu/blob/719319a11cf4a75608914696482e7d20aff0b56d/Kudu.Services.Web/App_Start/NinjectServices.cs#L701-L723

Needs to detect Linux and use ":" instead.

Same for Env.cshtml:
https://github.com/projectkudu/kudu/blob/719319a11cf4a75608914696482e7d20aff0b56d/Kudu.Services.Web/Env.cshtml#L84

davidebbo:
Yep. I don't suppose there is fx support for this like we have `Path.PathSeparator` to get the right slashes?

nickwalkmsft:
Whoa, thanks for making me actually go look :)

https://msdn.microsoft.com/en-us/library/system.io.path.pathseparator(v=vs.110).aspx

davidebbo:
Ha, I was thinking `Path.DirectorySeparatorChar`! :)

nickwalkmsft:
Also need to look around elsewhere in the codebase where semicolons may be hardcoded in relation to $PATH.

nickwalkmsft:
Closing as fixed.

PR #2313, opened 2017-01-27
Switch to Path.PathSeparator instead of hard-coded ";"

nickwalkmsft:
Fixing #2309


davidebbo:
Looks straightforward enough! :shipit:

Issue #2301, opened 2017-01-19
Async deployment creates temp deployment without acquiring lock


nickwalkmsft:
Calling /deploy?isAsync=true to run a background deployment creates the temporary deployment before acquiring the deployment lock.

https://github.com/projectkudu/kudu/blob/ee38ddde4405be333c1f724ea6ea16031eac046f/Kudu.Services/ServiceHookHandlers/FetchHandler.cs#L150

If the status of the latest deployment is requested from Kudu between the time the temp deployment is created and the time the lock is acquired, the status-retrieval logic detects the presence of a temp deployment but does not perceive a deployment occurring. As a result, it marks the temp deployment as failed, and then returns it as the latest deployment. Later, when the deployment actually completes, the status will be reflected correctly.

The window during which this can occur is normally tiny, but I have intermittently reproduced this on a production Linux site when requesting a deployment via an update to the site's sourceControl resource in ARM, which starts polling the completion status immediately after making the deployment request. My guess is that the nature and resource load of the site were such that the task that runs the background deployment takes a long time to start, or the lock takes a long time to acquire, or both. When Kudu responds to the poll request with the failure status of the temp deployment, the admin API called by ARM permanently marks the operation as failed. **This makes source control updates via Resource Explorer or template deployments look like they failed, even if they eventually completed successfully.**

nickwalkmsft:
Reproduced by manipulating thread timing in debugger.

Looking at the timestamps of logs in a production Linux instance it's easy to see how it would be possible for the front end's status request to come after the temp deployment is completed but before the lock is taken, especially under load.

nickwalkmsft:
Closing as fixed.

PR #2315, opened 2017-01-31
Fix for #2301

nickwalkmsft:
Fix for #2301 

For background deployments, create the temp deployment only after acquiring the deployment lock. If the deployment was requested as isAsync, wait (with timeout) for the temp deployment to be created before returning to the caller.

### DIFF
--- a/Kudu.Services/ServiceHookHandlers/FetchHandler.cs
+++ b/Kudu.Services/ServiceHookHandlers/FetchHandler.cs
@@ -140,21 +140,15 @@ namespace Kudu.Services
                 {
                     using (_tracer.Step("Start deployment in the background"))
                     {
-                        ChangeSet tempChangeSet = null;
-                        IDisposable tempDeployment = null;
-                        if (isAsync)
-                        {
-                            // create temporary deployment before the actual deployment item started
-                            // this allows portal ui to readily display on-going deployment (not having to wait for fetch to complete).
-                            // in addition, it captures any failure that may occur before the actual deployment item started
-                            tempDeployment = _deploymentManager.CreateTemporaryDeployment(
-                                                            Resources.ReceivingChanges,
-                                                            out tempChangeSet,
-                                                            deployInfo.TargetChangeset,
-                                                            deployInfo.Deployer);
-                        }
-
-                        PerformBackgroundDeployment(deployInfo, _environment, _settings, _tracer.TraceLevel, context.Request.Url, tempDeployment, _autoSwapHandler, tempChangeSet);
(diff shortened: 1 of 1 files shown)


Kudu.Services/ServiceHookHandlers/FetchHandler.cs
```diff
@@ -422,7 +423,11 @@ public static void PerformBackgroundDeployment(DeploymentInfo deployInfo, IEnvir
                 {"method", "POST"}
             });
 
-            Task.Run(() =>
+            // For monitoring creation of temp deployment
+            var tcs = new TaskCompletionSource<object>();
```
suwatch: why not simply auto or manual reset event?

Kudu.Services/ServiceHookHandlers/FetchHandler.cs
```diff
@@ -476,6 +498,13 @@ public static void PerformBackgroundDeployment(DeploymentInfo deployInfo, IEnvir
                     backgroundTrace.Dispose();
                 }
             });
+
+            if (waitForTempDeploymentCreation)
+            {
+                // Wait until the temp deployment is created before returning
+                var timeout = TimeSpan.FromSeconds(5);
+                await Task.WhenAny(tcs.Task, Task.Delay(timeout));
```
suwatch: Task.Delay will have an orphan timer in happy.   It will be moot if using auto/manual reset event.   If timeout what to do? Nothing if this is simply best effort to reduce (but not eliminate) the race temp deployment issue.

Kudu.Services/ServiceHookHandlers/FetchHandler.cs
nickwalkmsft: TCS was a little "friendlier" for the situation. ARE/MRE require disposal, try/catch check for ObjectDisposedException when calling Set() (if the timeout fires first, we'll dispose it before Set() gets called), and you can await a TCS task directly.

Kudu.Services/ServiceHookHandlers/FetchHandler.cs
nickwalkmsft: After reviewing a few things, I am getting rid of the timeout. It was intended to be a safeguard against the response never being returned if something went really wrong, but now I see that the only caller that will trigger this codepath is the frontend/ARM, and its behavior heavily depends on the temp deployment being created before the response is returned.

Instead of falling back on a timeout, it will fall back on the completion of the deployment task itself, in case it blew up early or failed to run for some odd reason.

suwatch:
:shipit:

PR #2320, opened 2017-02-11
kudusync needs to use LF and mono

michimune:
Since kudusync is supposed to run on Linux, all line break characters need to be LF instead of CRLF.
Also kudusync.NET.exe cannot be directly executed without mono.

### DIFF
--- a/Kudu.Core/Scripts/kudusync
+++ b/Kudu.Core/Scripts/kudusync
@@ -1,4 +1,4 @@
 #!/bin/sh
-"$0.NET.exe" "$@"
+mono "$0.NET.exe" "$@"
 ret=$?
 exit $ret
--- a/Kudu.Core/Scripts/kudusync
+++ b/Kudu.Core/Scripts/kudusync
@@ -1,4 +1,10 @@
 #!/bin/sh
-mono "$0.NET.exe" "$@"
+
+# Use mono if it is available
+MONO=
+mono --version >/dev/null 2>/dev/null
+if [ $? -eq 0 ]; then MONO=mono; fi
+
+$MONO "$0.NET.exe" "$@"
 ret=$?
 exit $ret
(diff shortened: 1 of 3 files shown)


Kudu.Core/Scripts/kudusync
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
-"$0.NET.exe" "$@"
+mono "$0.NET.exe" "$@"
```
davidebbo: We need to make sure that this file is never used on Windows, or it will be broken.

nickwalkmsft: Gah I said the same thing but didn't commit my review :)

Windows or no, isn't this file already used _somewhere_ and working?

davidebbo: I guess it must be used when users generate a custom deployment script and choose bash for the type. We need to test this.

michimune: I see. What about checking OS type, then? If it's Unix then prepend mono at the line. Does this logic sound right?

davidebbo: Yes, I think so. Note that to make things more confusing, we have 2 versions of kudusync: kudusync.net and kudusync (in different ProjectKudu repos). In Azure, we use kudusync.net, but I don't quite recall when the other one is used.

michimune:
Added logic to check mono and use it if available.
For the multiple kudusync versions, I found that there was node version of kudusync. But this shell script is explicitly invoking kudusync.NET.exe and I am not changing it. So I don't think my change will affect the behavior.

Anyway I want to test it on Windows but I don't know how. Is there an easy way to test?

nickwalkmsft:
This has been superseded by a fix to the Kudu Docker image in appsvc. The script in question is only intended to be called as a shim when a user supplies a custom bash deployment script on Windows. @michimune, please close this PR at your convenience.

michimune:
Great! I will close this and open a new PR for the InvalidOperationException issue.

PR #2323, opened 2017-02-14
Common post deployment routine

suwatch:
The objective is to provide common post deployment helper for various deployment engine (kudu and msdeploy). 

  - Relocate sync function trigger and auto swap routine to PostDeploymentHelper class.  Future post deployment steps will be handled by this class.    As a result, OperationClient and AutoSwapHandler classes are therefore deprecated.

  - PostDeploymentHelper class will be invoked using reflection from msdeploy.  To avoid dependency requirements, the class will only use .NET framework assemblies.  For instance, we will use the good old JavajScriptSerializer instead of external NewtonSoft.Json.          

  - Tracing is based on .NET TraceListener (again from Framework).


davidebbo:
@suwatch isn't that already merged in? If so, please close, thanks!

PR #2333, opened 2017-02-22
OnePerRequestHttpModule -> SafeOnePerRequestHttpModule

nickwalkmsft:
Replacing OnePerRequestHttpModule with similar implementation that
does an extra null check to avoid exceptions on Linux/Mono.

### DIFF
--- a/Kudu.Services.Web/SafeOnePerRequestHttpModule.cs
+++ b/Kudu.Services.Web/SafeOnePerRequestHttpModule.cs
@@ -1,0 +1,34 @@
+﻿namespace Kudu.Services.Web
+{
+    using System.Web;
+    using Ninject.Activation.Caching;
+    using Ninject;
+
+    /// <summary>
+    /// Based on Ninject.Web.Common's OnePerRequestHttpModule
+    /// (https://github.com/ninject/Ninject.Web.Common/blob/14fc90d197f02a528ab81eb2a3f12dccdfacb642/src/Ninject.Web.Common/OnePerRequestHttpModule.cs),
+    /// but does a null check on HttpContext.Current before calling MapKernels. On Linux/Mono, HttpContext.Current is occasionally null
+    /// in the EndRequest stage of the pipeline due to a bug, and the call to MapKernels throws.
+    /// If HttpContext.Current is null, disposal of InRequestScope resources will still happen, it just won't be as aggressive.
+    /// </summary>
+    public sealed class SafeOnePerRequestHttpModule : GlobalKernelRegistration, IHttpModule
+    {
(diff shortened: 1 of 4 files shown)


Kudu.Services.Web/SafeOnePerRequestNinjectModule.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// WebCommonNinjectModule registers its own implementation.
+    /// Note that the Ninject implementation is still registered (automatically via assembly
+    /// reflection on Ninject.Web.*), but it won't do anything at runtime since we don't
+    /// register its HttpModule with the application.
+    /// </summary>
+    public class SafeOnePerRequestNinjectModule : GlobalKernelRegistrationModule<SafeOnePerRequestHttpModule>
+    {
+        public override void Load()
+        {
+            base.Load();
```
davidebbo: Why override and just call base? Isn't it the same as not overriding?

Kudu.Services.Web/SafeOnePerRequestHttpModule.cs
```diff
@@ -0,0 +1,34 @@
...
+    using System.Web;
+    using Ninject.Activation.Caching;
+    using Ninject;
+
+    /// <summary>
+    /// Based on Ninject.Web.Common's OnePerRequestHttpModule
+    /// (https://github.com/ninject/Ninject.Web.Common/blob/14fc90d197f02a528ab81eb2a3f12dccdfacb642/src/Ninject.Web.Common/OnePerRequestHttpModule.cs),
+    /// but does a null check on HttpContext.Current before calling MapKernels. On Linux/Mono, HttpContext.Current is occasionally null
+    /// in the EndRequest stage of the pipeline due to a bug, and the call to MapKernels throws.
+    /// If HttpContext.Current is null, disposal of InRequestScope resources will still happen, it just won't be as aggressive.
```
davidebbo: Maybe open a Ninject issue (or PR) as well?

Kudu.Services.Web/SafeOnePerRequestNinjectModule.cs
nickwalkmsft: Whoops, correct of course. Removed.

Kudu.Services.Web/SafeOnePerRequestHttpModule.cs
nickwalkmsft: I'll file an issue with them and see what the maintainers say. The null HttpContext is a very weird issue, I've only seen it occur on Mono with our InfoRefsController.

Issue #2334, opened 2017-02-24
Functions that use the precompiled DLL in non-root folder aren't returned by "api/functions" call


alrod:
I have the following structure:

```
FunctionName\
  - function.json
bin
  - assembly.dll
```

The function FunctionName doesn't show up at all, until I add an empty content file to the folder. Expected behavior: show only function.json if a script content file doesn't exist.

move from portal repo:
https://github.com/projectkudu/AzureFunctionsPortal/issues/925

watashiSHUN:
so what should the `script URL` look like in this case? in my simple fix, `api/functions` simply return `null` which will make portal display nothing (change it to the address of `assembly.dll` is not going to generate a pretty display either?)

```
[{
	"name": "dllFunTest",
	"function_app_id": null,
	"script_root_path_href": "https://shunfunctionapp.scm.azurewebsites.net/api/vfs/site/wwwroot/dllFunTest/",
	"script_href": null,
	"config_href": "https://shunfunctionapp.scm.azurewebsites.net/api/vfs/site/wwwroot/dllFunTest/function.json",
	"secrets_file_href": "https://shunfunctionapp.scm.azurewebsites.net/api/vfs/data/functions/secrets/dllFunTest.json",
	"href": "https://shunfunctionapp.scm.azurewebsites.net/api/functions/dllFunTest",
	"config": {
		"bindings": [{
			"type": "manualTrigger",
			"direction": "in",
			"name": "input"
		}],
		"disabled": false
	},
	"files": null,
	"test_data": ""
},
{
	"name": "ManualTriggerCSharp1",
	"function_app_id": null,
	"script_root_path_href": "https://shunfunctionapp.scm.azurewebsites.net/api/vfs/site/wwwroot/ManualTriggerCSharp1/",
	"script_href": "https://shunfunctionapp.scm.azurewebsites.net/api/vfs/site/wwwroot/ManualTriggerCSharp1/run.csx",
	"config_href": "https://shunfunctionapp.scm.azurewebsites.net/api/vfs/site/wwwroot/ManualTriggerCSharp1/function.json",
	"secrets_file_href": "https://shunfunctionapp.scm.azurewebsites.net/api/vfs/data/functions/secrets/ManualTriggerCSharp1.json",
	"href": "https://shunfunctionapp.scm.azurewebsites.net/api/functions/ManualTriggerCSharp1",
	"config": {
		"bindings": [{
			"type": "manualTrigger",
			"direction": "in",
[...]

PR #2338, opened 2017-03-01
address issue #2334

watashiSHUN:
address issue #2334 
"function.json"
```
{
    "scriptFile": "bin/Debug/PreCompiledFunctionSample.dll",
    "entryPoint": "PreCompiledFunctionSample.MyFunction.Run",
     ...
}
```
kudu/api/functions:
```
[{
        ...
	"script_href": "https://shunfunctionapp.scm.azurewebsites.net/api/vfs/site/wwwroot/dllTest/bin/Debug/PreCompiledFunctionSample.dll",
	"config_href": "https://shunfunctionapp.scm.azurewebsites.net/api/vfs/site/wwwroot/dllTest/function.json",
	"secrets_file_href": "https://shunfunctionapp.scm.azurewebsites.net/api/vfs/data/functions/secrets/dllTest.json",
	"href": "https://shunfunctionapp.scm.azurewebsites.net/api/functions/dllTest"
        ...
}]
```
`DeterminePrimaryScriptFile()` comes from this [commit](https://github.com/Azure/azure-webjobs-sdk-script/commit/71142175acd4a48a938c79951861492583576313).

TODO: refactor this function into a library

### DIFF
--- a/Kudu.Core/Functions/FunctionManager.cs
+++ b/Kudu.Core/Functions/FunctionManager.cs
@@ -106,7 +106,7 @@ namespace Kudu.Core.Functions
                     FileSystemHelpers
                     .GetDirectories(_environment.FunctionsPath)
                     .Select(d => TryGetFunctionConfigAsync(Path.GetFileName(d))));
-                    // TryGetFunctionConfigAsync checks the existence of function.config
+                    // TryGetFunctionConfigAsync checks the existence of function.json
             return configList.Where(c => c != null);
         }
 
@@ -261,12 +261,13 @@ namespace Kudu.Core.Functions
         private FunctionEnvelope CreateFunctionConfig(string configContent, string functionName)
         {
             var functionConfig = JObject.Parse(configContent);
+            var functionPath = GetFunctionPath(functionName);
 
             return new FunctionEnvelope
             {
                 Name = functionName,
-                ScriptRootPathHref = FilePathToVfsUri(GetFunctionPath(functionName), isDirectory: true),
-                ScriptHref = FilePathToVfsUri(DeterminePrimaryScriptFile(functionConfig, GetFunctionPath(functionName))),
+                ScriptRootPathHref = FilePathToVfsUri(functionPath, isDirectory: true),
+                ScriptHref = FilePathToVfsUri(DeterminePrimaryScriptFile(functionConfig, functionPath)),
                 ConfigHref = FilePathToVfsUri(GetFunctionConfigPath(functionName)),
                 SecretsFileHref = FilePathToVfsUri(GetFunctionSecretsFilePath(functionName)),
                 Href = GetFunctionHref(functionName),
@@ -303,7 +304,7 @@ namespace Kudu.Core.Functions
             else
             {
                 string[] functionFiles = FileSystemHelpers.GetFiles(scriptDirectory,"*.*", SearchOption.TopDirectoryOnly)
-                    .Where(p => Path.GetFileName(p).ToLowerInvariant() != "function.json")
+                    .Where(p => ! String.Equals(Path.GetFileName(p),"function.json", StringComparison.OrdinalIgnoreCase))
                     .ToArray();
 
                 if (functionFiles.Length == 0)
@@ -321,8 +322,8 @@ namespace Kudu.Core.Functions
                     // if there is a "run" file, that file is primary,
                     // for Node, any index.js file is primary
                     functionPrimary = functionFiles.FirstOrDefault(p =>
-                        Path.GetFileNameWithoutExtension(p).ToLowerInvariant() == "run" ||
-                        Path.GetFileName(p).ToLowerInvariant() == "index.js");
+                        String.Equals(Path.GetFileNameWithoutExtension(p),"run",StringComparison.OrdinalIgnoreCase) ||
+                        String.Equals(Path.GetFileName(p),"index.js",StringComparison.OrdinalIgnoreCase));
                 }
             }
 
--- a/Kudu.Core/Functions/FunctionManager.cs
+++ b/Kudu.Core/Functions/FunctionManager.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Configuration;
 using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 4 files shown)


Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -266,7 +266,7 @@ private FunctionEnvelope CreateFunctionConfig(string configContent, string funct
             {
                 Name = functionName,
                 ScriptRootPathHref = FilePathToVfsUri(GetFunctionPath(functionName), isDirectory: true),
-                ScriptHref = FilePathToVfsUri(GetFunctionScriptPath(functionName, functionConfig)),
+                ScriptHref = FilePathToVfsUri(DeterminePrimaryScriptFile(functionConfig, GetFunctionPath(functionName))),
```
davidebbo: nit: Use `functionPath` variable since it's used in two places.

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -275,35 +275,46 @@ private FunctionEnvelope CreateFunctionConfig(string configContent, string funct
...
             }
             else
             {
-                // First see if there is an explicit primary file indicated
-                // in config. If so use that.
-                string functionPrimary = null;
-                string scriptFileName = (string)functionConfig["scriptFile"];
-                if (!string.IsNullOrEmpty(scriptFileName))
+                string[] functionFiles = FileSystemHelpers.GetFiles(scriptDirectory,"*.*", SearchOption.TopDirectoryOnly)
+                    .Where(p => Path.GetFileName(p).ToLowerInvariant() != "function.json")
```
davidebbo: Use case insensitive compare instead of ToLower(). Same for couple calls below. Generally, always avoid using ToLower() for comparison purpose. Use `String.Equals(s1, s2, StringComparison.OrdinalIgnoreCase) `.

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -313,9 +324,15 @@ internal static string DeterminePrimaryScriptFile(JObject functionConfig, string
                         Path.GetFileNameWithoutExtension(p).ToLowerInvariant() == "run" ||
                         Path.GetFileName(p).ToLowerInvariant() == "index.js");
                 }
+            }
 
-                return functionPrimary;
+            if (string.IsNullOrEmpty(functionPrimary))
+            {
+                throw new ConfigurationErrorsException("Unable to determine the primary function script. Try renaming your entry point script to 'run' (or 'index' in the case of Node), " +
+                    "or alternatively you can specify the name of the entry point script explicitly by adding a 'scriptFile' property to your function metadata.");
```
davidebbo: You throw various exception with descriptive messages, but don't they all get ignored in `TryGetFunctionConfigAsync`? Is so, what's the point?

davidebbo: I wonder if the better behavior would be to just have a null `ScriptHref` if we can't figure out the primary script. This way, they still show up in the portal. Talk to @alrod to make sure Portal can handle that. I think generally we want to be fairly lenient so that user can still open function in portal and fix issue. If you skip it, it just disappears.

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -275,35 +275,46 @@ private FunctionEnvelope CreateFunctionConfig(string configContent, string funct
...
+            // First see if there is an explicit primary file indicated
+            // in config. If so use that.
+            string functionPrimary = null;
+            string scriptFile = (string)functionConfig["scriptFile"];
+
+            if (!string.IsNullOrEmpty(scriptFile))
             {
-                // if there is only a single file, that file is primary
-                return functionFiles[0];
+                string scriptPath = Path.Combine(scriptDirectory, scriptFile);
```
davidebbo: Note that in some cases, the script file will be outside the folder (e.g. `../otherfolder/foo.js`). e.g. this will happen in scenarios like @christopheranderson's funcpack.

watashiSHUN: var filePathRelative = @"..\emptyWeb\index.html";
            var folderPath = @"D:\test";
            var pathConcate = Path.Combine(folderPath, filePathRelative); 
            var pathFull = Path.GetFullPath(pathConcate); 
            Console.WriteLine($"{pathConcate} : {File.Exists(pathConcate)}"); // D:\test\..\emptyWeb\index.html : True
            Console.WriteLine($"{pathFull} : {File.Exists(pathFull)}"); // D:\emptyWeb\index.html : True

we are calling `Path.GetFullPath` on `functionPrimary` at the end of this function

davidebbo: My question is more whether the caller expects the condition. i.e. would the portal then just open this as the primary file for that function even if it's not in the folder? They may cause weird behavior with the file list. Please discuss with @alrod

Kudu.Core/Functions/FunctionManager.cs
watashiSHUN: if we return null, exception messages will be lost (which are useful information for users, whereas currently if we give an invalid url, here's the message: `"Error:We are unable to reach your function app (Not Found). Please try again later."`), maybe we can do one of the following:
1. return the error string in `ScriptHref`
2. return null, add an error field to `functionEnvelop`
3. write the exception message to a error file in function folder, and pass its uri

I personally prefer **3**, as it requires almost no change on portal, but I don't know if it can cause function runtime/(3rd party which calls `api/functions`) to behave weirdly?

Kudu.Core/Functions/FunctionManager.cs
watashiSHUN: portal has a file system viewer, which does not support going outside of function folder

watashiSHUN:
@davidebbo we didn't finish our conversation before you left, should I add unit tests like these? (which will cover edge cases you mentioned in code review, they are basically testing the same thing as the [runtime tests](https://github.com/Azure/azure-webjobs-sdk-script/blob/dev/test/WebJobs.Script.Tests/ScriptHostTests.cs)), I am not sure whether I should duplicate these tests in kudu? the key test in this pull request is probably something like : [`DeterminePrimaryScriptFile_RelativeSourceFileSpecified()`](https://github.com/Azure/azure-webjobs-sdk-script/blob/dev/test/WebJobs.Script.Tests/ScriptHostTests.cs#L234-L254)

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -275,47 +276,71 @@ private FunctionEnvelope CreateFunctionConfig(string configContent, string funct
...
+                    .ToArray();
+
+                if (functionFiles.Length == 0)
                 {
-                    functionPrimary = functionFiles.FirstOrDefault(p =>
-                        string.Compare(Path.GetFileName(p), scriptFileName, StringComparison.OrdinalIgnoreCase) == 0);
+                    TraceAndThrowError(new ConfigurationErrorsException("No function script files present."));
+                }
+
+                if (functionFiles.Length == 1)
```
ahmelsayed: else if?

davidebbo: No point, it throws

PR #2347, opened 2017-03-06
Fix DebugConsole monospace font for Macs

snobu:
Added **Monaco** to DebugConsole font list. Courier makes the console look like it's 1995 on a Mac (which depending on the nostalgia factor could be regarded as a feature)..

Before (Courier New):
screen shot 2017-03-06 at 11 09 31 am

After (Monaco):
screen shot 2017-03-06 at 11 07 24 am

### DIFF
--- a/Kudu.Services.Web/Content/Styles/FileBrowser.css
+++ b/Kudu.Services.Web/Content/Styles/FileBrowser.css
@@ -86,7 +86,7 @@ div.upload-unzip-show {
 }
 
 div.zip-upload-text {
-    font-family: Consolas, 'Lucida Console', 'Courier New', monospace;
+    font-family: Consolas, Monaco, 'Lucida Console', 'Courier New', monospace;
     font-size: large;
     position: relative;
     text-align: center;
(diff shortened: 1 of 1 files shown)


davidebbo:
I'm not much of a font person. @lindydonna what do you think? :)

snobu:
The main idea here being that everything on a Mac uses Monaco and Courier looks a little alien there..   C 👽 urier

davidebbo:
To tell you how clueless I am about fonts, I didn't even know there was one called Monaco. Interestingly, it's not an option in Word/Outlook. Maybe it's a Mac only thing?

davidebbo:
Ok, so on Windows, your change makes no difference at all. So I assume it only affects Mac. Can you confirm? I'm fine getting that in.

snobu:
Correct. No Monaco on Windows, it doesn't affect a thing. Monaco is a Mac thing only, like kind of what Consolas is to Windows. High-DPIs very well.. If that's a verb...

lindydonna:
@snobu I like the change!  Monaco is my favorite monospace font on the Mac.

PR #2349, opened 2017-03-06
Preserve original order of PATH items

toddbrownmcis:
Updated Env.cshtml to preserve the original order of PATH items.  The original order of PATH items is significant, ordering them alphabetically is wrong.


davidebbo:
Fair enough.

davidebbo:
Thanks for the contrib!

Issue #2342, opened 2017-03-03
Function listsecret API creates secrets even when function does not exist


davidebbo:
e.g. doing a `POST /api/functions/bogus/listsecrets` creates and returns a secret. Instead, it should return a 404 if the Function folder does not exist.

jvano:
Hi

If the problem persists and is related to running it on Azure App Service, please open a support incident in Azure:
https://learn.microsoft.com/en-us/azure/azure-portal/supportability/how-to-create-azure-support-request

This way we can better track and assist you on this case

Thanks,

Joaquin Vano
Azure App Service

PR #2359, opened 2017-03-13
check function existance before creating function keys, #2342

watashiSHUN:
if the function folder is not found, it returns a 404:
```
https://azurekudufunctionapp.scm.azurewebsites.net/api/functions/ManualTriggerJS4/listsecrets

{
  "Message": "An error has occurred.",
  "ExceptionMessage": "Function (ManualTriggerJS4) does not exist",
  "ExceptionType": "System.IO.FileNotFoundException",
  "StackTrace":  ...
}
```
to be safer, we call also read the `function.json` to [make sure the function is valid](https://github.com/projectkudu/kudu/blob/master/Kudu.Core/Functions/FunctionManager.cs#L116-L120).

### DIFF
--- a/Kudu.Core/Functions/FunctionManager.cs
+++ b/Kudu.Core/Functions/FunctionManager.cs
@@ -195,7 +195,7 @@ namespace Kudu.Core.Functions
         public async Task<FunctionSecrets> GetFunctionSecretsAsync(string functionName)
         {
             // check to see if the function folder exists
-            GetFunctionPath(functionName);
+            GetFuncPathAndCheckExistance(functionName);
             return await GetKeyObjectFromFile<FunctionSecrets>(functionName, new FunctionSecretsJsonOps());
         }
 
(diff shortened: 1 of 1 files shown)


watashiSHUN:
since this is the only **GetXYZPath()** that might throw an exception (compared to [these](https://github.com/projectkudu/kudu/blob/master/Kudu.Core/Functions/FunctionManager.cs#L410-L420)) and because `listsecret()` depends on this behavior, I renamed the function in case future code change forget about this

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -376,7 +378,7 @@ private string GetFunctionPath(string name)
 
         private string GetFunctionConfigPath(string name)
         {
-            return Path.Combine(GetFunctionPath(name), Constants.FunctionsConfigFile);
+            return Path.Combine(GetFuncPathAndCheckExistance(name), Constants.FunctionsConfigFile);
```
davidebbo: It's spelled 'exist**e**nce' :)

davidebbo:
Other than misspelling 'misspelled', looks good! :)

PR #2362, opened 2017-03-15
Fix proper flowing requestid and jwt

suwatch:


suwatch:
For deployment inproc (thru /deploy), we flow requestid and jwt thru IEnvironment.    Except in one git push (postreceive), we will have to flow via env variables.

davidebbo:
This was merged earlier.

Issue #2340, opened 2017-03-03
Kudu REST call failure should be logged to azure logging table


suwatch:
For any REST call failure, we should log to Kusto (azure logging table).  It should contain.

  - REST api path (w/o query string)
  - Http Status
  - Exception or error message
  - RequestId for correlation - this might already be there.
  - Optionally Latency
  - Optionallu UserAgent, ClientPrincipal (these are incoming headers)

jvano:
Hi

If the problem persists and is related to running it on Azure App Service, please open a support incident in Azure:
https://learn.microsoft.com/en-us/azure/azure-portal/supportability/how-to-create-azure-support-request

This way we can better track and assist you on this case

Thanks,

Joaquin Vano
Azure App Service

PR #2368, opened 2017-03-20
add ETWTracer to kudu

watashiSHUN:
capture

first line is the test output of :
image
second line:
image
third line:
image

fix issue #2340, but not yet tested in geneva


watashiSHUN:
@suwatch why in [traceModule](https://github.com/projectkudu/kudu/blob/master/Kudu.Services.Web/Tracing/TraceModule.cs#L222) we use `Guid.NewGuid()`, but in [ninject](https://github.com/projectkudu/kudu/blob/master/Kudu.Core/Environment.cs#L138) we use `Guid.Empty` for missing requestId?

PR #2374, opened 2017-03-21
Update detailed tracing agent GUID

pharring:
In a future Antares build (targeting end of April 2017) , we will be updating the ServiceProfilerAgent.dll which is used to assist with profiling. In that update, we will remove support for (what we call) the V1V2 circular monitor agent. V3 will be the only one supported in Antares.

Therefore, the ```DetailedTracingAgentGuid``` constant needs to be updated to the V3 agent GUID:
https://github.com/projectkudu/kudu/blob/master/Kudu.Services/Diagnostics/ProfileManager.cs#L22

This change can be made now. V3 has been available since November 2016, so it is already in all Antares stamps.

Just updating the GUID is enough. There are no other changes required.


davidebbo:
@safihamid / @puneet-gupta, could you test with a private Kudu build that things still work on Azure with this change? If so, I'll merge.

puneet-gupta:
I just saw this...will test tomorrow and confirm...thanks

puneet-gupta:
@davidebbo  verified that this GUID works. tested with a private built of KUDU

PR #2377, opened 2017-03-23
remove TODO comment on list-secrets

watashiSHUN:
this `TODO` was created at [this commit back in March 2016](https://github.com/projectkudu/kudu/commit/f9947cc441240dcb266698e2539ca8a30b5480fa) and could be removed by [this commit in May 2016](https://github.com/projectkudu/kudu/commit/0bfe30f857fb61a67a31ef17e57a15bac4443845)

btw, why are we using `POST` instead of `PUT` for "list-secrets"?


davidebbo:
POST is the standard ARM pattern for what would normally be a GET, but cannot be because it's sensitive and we don't want 'readers' to access it.

Issue #2380, opened 2017-03-24
Linux: Custom Kuduscript-generated scripts fail


nickwalkmsft:
Custom bash deployment scripts and associated .deployment files generated by a developer, checked into a repo and pushed to a Linux App Service result in deployment failure with confusing "cannot execute binary file" error.

Root cause is explicit inclusion of "bash" in the .deployment file's command field, but fix is not to remove it. This is required on Windows and shouldn't cause a failure. Fix is in handling of deployment script commands within Kudu.

PR #2381, opened 2017-03-24
Fix for #2380

nickwalkmsft:
- Creates a bash "starter script" for use in ExternalCommandFacotry.BuildExternalCommandExecutable on Linux, to parallel the Windows implementation. Original implementation prefixed all commands with "/bin/bash", which really only makes sense if command is a bash script.

### DIFF
--- a/Kudu.Core/Scripts/starter.sh
+++ b/Kudu.Core/Scripts/starter.sh
@@ -1,0 +1,2 @@
+#!/bin/bash
+exec "$@"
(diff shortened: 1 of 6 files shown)


Kudu.Core/Scripts/starter.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
```
am11: This complains when bash is not the default shell (e.g Ubuntu 16 uses dash, Alpine Linux uses ash, NetBSD has csh etc.) 
Please consider using more portable shebang: `#!/usr/bin/env bash` instead of `#!/bin/bash`.

PR #2384, opened 2017-03-25
Fixing the CLR rundown events logic in IIS and threadtime profiling

puneet-gupta:
The current diagsession file collected is missing the CLR rundown events which prohibits us to see all the function names properly. Creating this pull request after understanding the right sequence of commands to capture the traces to avoid this issue.


pharring:
Looks good to me.

PR #2386, opened 2017-03-27
migrate from masterkey to admintoken for portal->function runtime authentication

watashiSHUN:
address Azure/azure-webjobs-sdk-script#1334

1. add a new api `https://testshun.scm.azurewebsites.net/api/functions/admin/token`
2. returned payload: 
```
{
  "nbf": 1490650554,
  "exp": 1490650674,
  "iat": 1490650554,
  "iss": "https://testShun.scm.azurewebsites.net",
  "aud": "https://testShun.azurewebsites.net/azurefunctions"
}
```
**[Note1]** this api will fail when testing on dev machine (non azure environment)
**[Note2]** `<package id="Microsoft.Azure.WebSites.DataProtection" version="0.1.77-alpha" targetFramework="net46" />` comes from **azure app service feed**, not yet in **nuget.org**

looks like there's inconsisitency in VS xml formatting: [this commit](https://github.com/projectkudu/kudu/commit/121eef5fe9f1d987bc0662eee72ca166cfdcb5c0#diff-a57a16f860710345042779a3debf81ad) has both `123/>` and `123 />` style when it comes to self-closing tag.

I suggest we move all to `123 />` with an extra space

### DIFF
--- a/Kudu.Core/Infrastructure/SecurityUtility.cs
+++ b/Kudu.Core/Infrastructure/SecurityUtility.cs
@@ -39,6 +39,13 @@ namespace Kudu.Core.Infrastructure
             return protector.Unprotect(content);
         }
 
+        public static string GenerateFunctionToken()
+        {
+            string siteName = ServerConfiguration.GetApplicationName();
+            string issuer = $"https://{siteName}.scm.azurewebsites.net";
+            string audience = $"https://{siteName}.azurewebsites.net/azurefunctions";
+            return JwtGenerator.GenerateToken(issuer, audience, expires: DateTime.UtcNow.AddMinutes(2));
+        }
     }
 }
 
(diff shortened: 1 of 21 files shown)


Kudu.Core/Infrastructure/SecurityUtility.cs
```diff
@@ -39,6 +39,13 @@ public static string DecryptSecretString(string content)
             return protector.Unprotect(content);
         }
 
+        public static string GenerateFunctionToken()
+        {
+            string siteName = ServerConfiguration.GetApplicationName();
+            string issuer = $"https://{siteName}.scm.azurewebsites.net";
```
davidebbo: Note that on other deployments, the host name is not azurewebsites.net. It's probably not a big deal, since I don't think this is used as an actual URL. Does it even need to be a url? /cc @fabiocav

Issue #2392, opened 2017-03-30
WebHooks Manager does not validate the URL passed in via the Web Hook.


FinVamp1:
If there is no URL in the JSON for the WebHook then a URI Exception can bring down the process when the WebHook is triggered.

E.G.

RegisterServices() at C:\Kudu Files\Private\src\master\Kudu.Services.Web\App_Start\NinjectServices.cs:194

System.UriFormatException: Invalid URI: The URI is empty.

   at System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind)

   at System.Uri..ctor(String uriString)

   at Kudu.Core.Hooks.WebHooksManager.CreateHttpClient(WebHook webHook)

   at Kudu.Core.Hooks.WebHooksManager.<PublishToHookAsync>d__17.MoveNext()

--- End of stack trace from previous location where exception was thrown ---

   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)

   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)

   at Kudu.Core.Hooks.WebHooksManager.<PublishToHooksAsync>d__19.MoveNext()

--- End of stack trace from previous location where exception was thrown ---

   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)

   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)

   at Kudu.Core.Hooks.WebHooksManager.<>c__DisplayClass14_0.<<PublishEventAsync>b__0>d.MoveNext()

--- End of stack trace from previous location where exception was thrown ---

   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)

   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)

   at Kudu.Contracts.Infrastructure.LockExtensions.<LockOperationAsync>d__4.MoveNext()

--- End of stack trace from previous location where exception was thrown ---

   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)

   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)

   at Kudu.Core.Hooks.WebHooksManager.<PublishEventAsync>d__14.MoveNext()

[...]

PR #2393, opened 2017-03-30
This is a simple change to check if there is a valid Address before t…

FinVamp1:
…rying to run the Web Hook.


davidebbo:
So it seems that it doesn't really check if the URL is valid, but only looks for empty string. Does it cause en exception if the string is empty? Maybe there is a more general way of handling bad URLs, which presumably throw when trying to new the Uri() object.

FinVamp1:
It does throw an exception #2392  has this detail. 

RegisterServices() at C:\Kudu Files\Private\src\master\Kudu.Services.Web\App_Start\NinjectServices.cs:194

System.UriFormatException: Invalid URI: The URI is empty.

at System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind)

at System.Uri..ctor(String uriString)

at Kudu.Core.Hooks.WebHooksManager.CreateHttpClient(WebHook webHook)

at Kudu.Core.Hooks.WebHooksManager.d__17.MoveNext()

--- End of stack trace from previous location where exception was thrown ---

at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)

at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)

at Kudu.Core.Hooks.WebHooksManager.d__19.MoveNext()

I opened two issues. 1) Consider validation in the UI before creating the local JSON and 2) Have some protection here to see if it is valid as well.  

Will have more time soon to see if I can fix it differently in either scenario.

davidebbo:
Ok, but with your fix, it no longer throw then (API returns 200), and instead just logs the error? From an API standpoint, it seems wrong to have the user pass an invalid payload (empty url), and not have that result in an API error.

If our UI makes it too easy to forget the URL, I'd rather just invest in fixing that client side, and leave the API alone.

Does that make sense?

FinVamp1:
That makes sense. The only issue is that this exception does shut down Kudu it seems. I will do some more testing on that front.

davidebbo:
Really, it causes the Kudu process to crash? That's not good, and is surprising to me.

davidebbo:
I wasn't able to repro this. If I go to WebHooks UI and click Add Url without entering a URL, the Add happens (which is a bit bogus), but it doesn't crash. Or maybe it crashes later when the hook fires?

PR #2394, opened 2017-03-31
Jamchri docker ci

Hamster-Huey:
Add an http route for DockerHub and ACR webhook receiving endpoint to kudu. Linux only. 

Please let me know what I need to do to make this endpoint "compliant"/production-worthy. 

All it does is receive a payload and update a timestamp file in the site directory.

### DIFF
--- a/Kudu.Services.Web/App_Start/NinjectServices.cs
+++ b/Kudu.Services.Web/App_Start/NinjectServices.cs
@@ -531,6 +531,9 @@ namespace Kudu.Services.Web.App_Start
             routes.MapHttpRoute("get-masterkey", "api/functions/admin/masterkey", new { controller = "Function", action = "GetMasterKey" }, new { verb = new HttpMethodConstraint("GET") });
             routes.MapHttpRoute("delete-function", "api/functions/{name}", new { controller = "Function", action = "Delete" }, new { verb = new HttpMethodConstraint("DELETE") });
 
+            // Docker Hook Endpoint
+            routes.MapHttpRoute("docker", "docker/hook", new { controller = "Docker", action = "ReceiveHook" }, new { verb = new HttpMethodConstraint("POST") });
+
             // catch all unregistered url to properly handle not found
             // this is to work arounf the issue in TraceModule where we see double OnBeginRequest call
             // for the same request (404 and then 200 statusCode).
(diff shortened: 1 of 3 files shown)


Kudu.Services.Web/App_Start/NinjectServices.cs
```diff
@@ -531,6 +531,9 @@ public static void RegisterRoutes(IKernel kernel, RouteCollection routes)
             routes.MapHttpRoute("get-masterkey", "api/functions/admin/masterkey", new { controller = "Function", action = "GetMasterKey" }, new { verb = new HttpMethodConstraint("GET") });
             routes.MapHttpRoute("delete-function", "api/functions/{name}", new { controller = "Function", action = "Delete" }, new { verb = new HttpMethodConstraint("DELETE") });
 
+            // Docker Hook Endpoint
+            routes.MapHttpRoute("docker", "docker/hook", new { controller = "Docker", action = "ReceiveHook" }, new { verb = new HttpMethodConstraint("POST") });
```
davidebbo: I'll let @nickwalkmsft review, but should the route only be added on Linux?

Kudu.Services/Docker/DockerController.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.IO;
+using System.Net;
+using System.Net.Http;
+using System.Web.Http;
+
+namespace Kudu.Services.Docker
+{
+    public class DockerController : ApiController
+    {
+        const string TimestampDirectory = "/home/config";
```
nickwalkmsft: Does the functionality require that the timestamp file is in the mounted site directory? If so we'll want to keep a close eye on this, as a feature that may not be too far off is allowing the site drive to be mounted somewhere besides /home.

Kudu.Services/Docker/DockerController.cs
```diff
@@ -0,0 +1,46 @@
...
+                return Request.CreateResponse(HttpStatusCode.NotFound);
+            }
+
+            var tracer = _traceFactory.GetTracer();
+            using (tracer.Step("Docker.SetDockerTimestamp"))
+            {
+                string timestampPath = Path.Combine(TimestampDirectory, "dockerTimestamp.txt");
+                try
+                {
+                    File.WriteAllText(timestampPath, DateTime.UtcNow.ToString());
```
nickwalkmsft: Can't remember - will this create the directory if it doesn't exist?

Kudu.Services.Web/App_Start/NinjectServices.cs
Hamster-Huey: For now I only see this route being useful in the Linux ecosystem as it is purely for containers/docker images.

nickwalkmsft: That's what I'm saying - can you completely skip the route registration here if we're on Windows?

Hamster-Huey: Ah i see now. your comment didn't pop up until i refreshed this page. adding the check here makes more sense for sure.

PR #2400, opened 2017-04-06
change match string for npm error message (using GUID)

watashiSHUN:
the public test failed because somehow `MadeUpKuduPackage` is added to `npm cache` on kudu-ci machine. Adding a `GUID` to the `madeUpPackageName` should prevent it from matching any `npm cache` (if accident happened again)

I removed the old match string: `"One or more of the selected node/npm paths do not exist."` since this error message is returned by [kudu](https://github.com/projectkudu/kudu/blob/2ea953e532d0a44afe291dc4b9e370df9fc51686/Kudu.Core/Scripts/selectNodeVersion.js#L288), not by `npm install`


davidebbo:
:shipit:

PR #2402, opened 2017-04-06
Put funcpack on path and generalize npm tool handling

davidebbo:
@nickwalkmsft adding you to review as this affects the Linux code path as well, and I didn't test it.

### DIFF
--- a/Kudu.Core/Infrastructure/PathUtils/PathUtilityBase.cs
+++ b/Kudu.Core/Infrastructure/PathUtils/PathUtilityBase.cs
@@ -8,6 +8,14 @@ namespace Kudu.Core.Infrastructure
 {
     public abstract class PathUtilityBase
     {
+        readonly string[] _npmTools = new string[]
+        {
+            "bower",
+            "grunt",
+            "gulp",
+            "funcpack",
+        };
+
         // Return a list of folders that need to be on the %PATH%
         public virtual List<string> GetPathFolders(IEnvironment environment)
         {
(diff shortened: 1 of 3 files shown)


Kudu.Core/Infrastructure/PathUtils/PathUtilityBase.cs
```diff
@@ -8,6 +8,14 @@ namespace Kudu.Core.Infrastructure
 {
     public abstract class PathUtilityBase
     {
+        readonly string[] _npmTools = new string[]
```
ahmelsayed: could be static, right?

Kudu.Core/Infrastructure/PathUtils/PathUtilityBase.cs
```diff
@@ -26,12 +34,8 @@ public virtual List<string> GetPathFolders(IEnvironment environment)
             toolsPaths.AddRange(ResolveNodeNpmPaths());
             toolsPaths.Add(ResolveNpmGlobalPrefix());
 
-            toolsPaths.AddRange(new[]
-            {
-                ResolveBowerPath(),
-                ResolveGruntPath(),
-                ResolveGulpPath()
-            }.Where(p => !String.IsNullOrEmpty(p)).Select(Path.GetDirectoryName));
+            toolsPaths.AddRange(_npmTools.Select(name => ResolveNpmToolsPath(name))
```
ahmelsayed: could be simplified to `_npmTools.Select(ResolveNpmToolsPath)`

ahmelsayed:
so I'm just curious, are these tools also installed on linux workers? they are installed through rapid update for Windows workers, but I'm not sure what happens for linux.

nickwalkmsft:
Funny, I had the same thought but moved to an offline conversation...

These tools don't actually exist at that path on Linux, but I haven't seen any problems with the default node deployment scripts running. Are these explicitly put on the path for support in custom deployment scripts, or do node deployments on Windows not work at all without them on the path?

Kudu.Core/Infrastructure/PathUtils/PathUtilityBase.cs
davidebbo: Yep, I'll change.

Kudu.Core/Infrastructure/PathUtils/PathUtilityBase.cs
davidebbo: Indeed!

davidebbo:
I actually don't quite remember why we needed bower/grunt/gulp. I think it was related to building dotnet core, but I'm not sure if that applies to all three.

ahmelsayed:
yeah, the default VS template for ASP.NET Core was expecting these tools to be on the path. That might have changed since then, but I'm pretty sure others have taken dependency on these tools in their deployment scripts since then

PR #2404, opened 2017-04-07
Avoid two redundant directoryExits()

watashiSHUN:
[`GetFunctionConfigPath`](https://github.com/projectkudu/kudu/blob/master/Kudu.Core/Functions/FunctionManager.cs#L253) calls [`GetFuncPathAndCheckExistence`](https://github.com/projectkudu/kudu/blob/master/Kudu.Core/Functions/FunctionManager.cs#L256) and immediately after
[`CreateFunctionConfig`](https://github.com/projectkudu/kudu/blob/master/Kudu.Core/Functions/FunctionManager.cs#L269-L277) calls [`GetFuncPathAndCheckExistence`](https://github.com/projectkudu/kudu/blob/master/Kudu.Core/Functions/FunctionManager.cs#L381-L384) two more times, which I believe are wasted file system access

### DIFF
--- a/Kudu.Core/Functions/FunctionManager.cs
+++ b/Kudu.Core/Functions/FunctionManager.cs
@@ -250,10 +250,11 @@ namespace Kudu.Core.Functions
         {
             try
             {
-                var path = GetFunctionConfigPath(name);
-                if (FileSystemHelpers.FileExists(path))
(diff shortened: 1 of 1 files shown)


davidebbo:
It feels like a micro optimization. Is the benefit greater than the risk of regression? :)

PR #2268, opened 2016-12-13
Ruby Site Deployment 

Hamster-Huey:
Add support for detecting ruby sites and running a ruby-specific deployment scenario.

PR #2405, opened 2017-04-10
PHP Site Deployment

Benjiiim:
Now that kudu container on Azure App Service on Linux has PHP and composer support (cf https://github.com/Azure-App-Service/kudu/pull/15), it seems a great time for kudu to be able to detect php apps.
PHP deployement logic has been submitted on https://github.com/projectkudu/KuduScript/pull/69
This PR was inspired from the Ruby support at #2268

### DIFF
--- a/Kudu.Core/Deployment/Generator/SiteBuilderFactory.cs
+++ b/Kudu.Core/Deployment/Generator/SiteBuilderFactory.cs
@@ -159,6 +159,10 @@ namespace Kudu.Core.Deployment.Generator
             {
                 return new RubySiteBuilder(_environment, perDeploymentSettings, _propertyProvider, repositoryRoot, projectPath);
             }
+            else if (IsPHPSite(sourceProjectPath))
+            {
+                return new PHPSiteBuilder(_environment, perDeploymentSettings, _propertyProvider, repositoryRoot, projectPath);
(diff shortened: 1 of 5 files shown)


nickwalkmsft:
KuduScript changes have been merged, if you can modify updateNodeModules, we should be able to get this in as well.

Benjiiim:
updateNodeModules updated to get the last version of KuduScript which include the PHP deployment script.

Kudu.Core/Deployment/Generator/SiteBuilderFactory.cs
```diff
@@ -159,6 +159,10 @@ private ISiteBuilder ResolveNonAspProject(string repositoryRoot, string projectP
             {
                 return new RubySiteBuilder(_environment, perDeploymentSettings, _propertyProvider, repositoryRoot, projectPath);
             }
+            else if (IsPHPSite(sourceProjectPath))
```
davidebbo: I think we should only do this for Linux, or we might break existing PHP sites on Windows.

davidebbo: Oh, I see that the enabler class checks for this, so we're good.

PR #2409, opened 2017-04-14
rename applicationHost.xdt to scmApplicationHost.xdt

watashiSHUN:
1. kudu itself will come with `scmApplicationHost.xdt` instead of `applicationHost.xdt`
2. for siteExtensions that do not have their own xdts, the default generated by kudu will also be a `scmApplicationHost.xdt` since kudu's template has the environment var: `"%XDT_SCMSITENAME%"` instead of `%XDT_SITENAME%` (tested with [repoNoXdt](https://github.com/watashiSHUN/siteExtensionTest) and [repoNewFormatXdt](https://github.com/watashiSHUN/siteExtensionTest/tree/webrootwithscmxdt))
3. code path with the `SiteExtensionInfo.SiteExtensionType.WebRoot == type` is not full tested, since we don't really hit it (in kudu we check [if the package is of type WebRoot](https://github.com/projectkudu/kudu/blob/master/Kudu.Core/SiteExtensions/SiteExtensionManager.cs#L255-L260) before [the package is downloaded](https://github.com/projectkudu/kudu/blob/master/Kudu.Core/SiteExtensions/SiteExtensionManager.cs#L282) could be a bug...but leaving it like this for now)


suwatch:
Merged https://github.com/projectkudu/kudu/commit/218518c663000a81fb6f053eb46e3f0abcff0f2e

PR #2412, opened 2017-04-20
Honor scripts.start even if there is no web.config

### LINKED ISSUE #2411: Should modify iisnode.yml when it is under a folder and pointed by package.json/scripts/start
The location of `iisnode.yml` is based on the location of the Node.js entrypoint, but not `web.config`.

Currently, Kudu modify the `iisnode.yml` in the root, instead of locating it at the same level of the Node.js entrypoint.

### Repro steps.
1. Write a web server entrypoint named `lib/app.js` and it simply spill out `process.env.node_env`
2. Modify `package.json` and set `scripts.start` to `node lib/app.js`
3. Write a `lib/iisnode.yml` with content `node_env: under-lib`
4. Write another `iisnode.yml` with content `node_env: root`
5. Have a usual `web.config` that run the project with `iisnode`, for example, copy from https://gist.github.com/davidebbo/2571104
5. Deploy thru Project Kudu (in my case, it's Azure Web App)

#### Expected
1. `lib/iisnode.yml` is appended with pointer to Node.js of the latest version (`7.7.4` when I deployed to Azure)
2. When browse to the site, it say `under-lib`

#### Actual
1. `iisnode.yml` is modified instead
2. When browse to the site, it say `root`

### Project structures.
```
/iisnode.yml (this file should be ignored)
/package.json (this file has scripts.start pointing to /lib/app.js)
/lib/iisnode.yml (this file should be modified)
/lib/app.js (entrypoint and response with process.env.node_env)
```

### The log/error given by the failure.
The wrong `iisnode.yml` is modified.

### Mention any other details that might be useful.
I verified the behavior of locating `iisnode.yml` in an environment with IIS and iisnode, no Kudu and not on Azure.

You can take my [Docker image](https://hub.docker.com/r/compulim/iisnode) that have `iisnode` preinstalled and try it out yourself easily.

davidebbo:
This should fix #2411.

### DIFF
--- a/Kudu.Core/Scripts/selectNodeVersion.js
+++ b/Kudu.Core/Scripts/selectNodeVersion.js
@@ -146,50 +146,48 @@ var nodeStartFilePath = (function createIisNodeWebConfigIfNeeded() {
         webConfigWwwRootPath = path.join(wwwroot, 'web.config'),
         nodeStartFilePath = null;
 
-    if (!existsSync(webConfigRepoPath)) {
-        // Check for {"scripts": {"start": < startupCommand > } } exists
-        if (typeof json === 'object' && typeof json.scripts === 'object' && typeof json.scripts.start === 'string') {
-            var startupCommand = json.scripts.start;
-            var defaultNode = "node ";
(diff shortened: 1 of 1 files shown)


davidebbo:
Closing old PR.

PR #2414, opened 2017-04-25
add case for functionapp with package.json under root folder

watashiSHUN:
use funcpack for functionApp CI deployment

tested for 
1. `package.json` under root [[repro](https://github.com/watashiSHUN/FunctionAppWithBigNpm)]
2. `package.json` under each function folder [[repro](https://github.com/watashiSHUN/FunctionAppWithBigNpm/tree/twopackage.json)]
3. combine 1 and 2 [[repro](https://github.com/watashiSHUN/FunctionAppWithBigNpm/tree/oneatrootoneatsubfolder)]

kuduscript changes see [here](https://github.com/projectkudu/KuduScript/commit/197392188376c695ecefd81d3d077553fc8375e3?diff=unified)

### DIFF
--- a/Kudu.Core/Deployment/Generator/SiteBuilderFactory.cs
+++ b/Kudu.Core/Deployment/Generator/SiteBuilderFactory.cs
@@ -139,7 +139,12 @@ namespace Kudu.Core.Deployment.Generator
         private ISiteBuilder ResolveNonAspProject(string repositoryRoot, string projectPath, IDeploymentSettingsManager perDeploymentSettings)
         {
             string sourceProjectPath = projectPath ?? repositoryRoot;
-            if (IsNodeSite(sourceProjectPath))
+            if (IsFunctionApp(sourceProjectPath))
+            {
+                return new FunctionAppBuilder(_environment, perDeploymentSettings, _propertyProvider, repositoryRoot, projectPath);
+            }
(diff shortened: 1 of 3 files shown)


davidebbo:
Added comment to https://github.com/projectkudu/KuduScript/commit/197392188376c695ecefd81d3d077553fc8375e3

Kudu.Core/Deployment/Generator/SiteBuilderFactory.cs
```diff
@@ -139,7 +139,12 @@ public ISiteBuilder CreateBuilder(ITracer tracer, ILogger logger, IDeploymentSet
         private ISiteBuilder ResolveNonAspProject(string repositoryRoot, string projectPath, IDeploymentSettingsManager perDeploymentSettings)
         {
             string sourceProjectPath = projectPath ?? repositoryRoot;
-            if (IsNodeSite(sourceProjectPath))
+            if (IsFunctionApp(sourceProjectPath))
```
davidebbo: To make sure we don't hijack any existing Node apps that are not meant to be Functions, can you change the `IsFunctionApp` implementation to check if the `FUNCTIONS_EXTENSION_VERSION` environment variable is defined (value doesn't matter)? This is the best way to guarantee it's a function app, without needing to make guesses based on files. Note that we may allow pushing functions apps later that don't have a `host.json`.

PR #2416, opened 2017-05-01
disable in place deployment for function app

watashiSHUN:
add a simple patch to the existing function`CopyDirectoryRecursive` to exclude directories/subdirectories with name "node_modules". This will cover functionApp's use case where there's a `package.json` under `wwwroot` or under each `wwwroot/functionFolder`

note to self:
1. add `SCM_USE_FUNCPACK` to kudu doc as a deployment flag
2. add functional tests for git clone functionApp

### DIFF
--- a/Kudu.Core/Infrastructure/FileSystemHelpers.cs
+++ b/Kudu.Core/Infrastructure/FileSystemHelpers.cs
@@ -205,7 +205,7 @@ namespace Kudu.Core.Infrastructure
         public static void CopyDirectoryRecursive(string sourceDirPath, string destinationDirPath, bool overwrite = true, HashSet<string> ignoreDirs = null)
         {
             // Get the subdirectories for the specified directory.
-            var sourceDir = new DirectoryInfo(sourceDirPath);
+            var sourceDir = DirectoryInfoFromDirectoryName(sourceDirPath);
 
             if (!sourceDir.Exists)
             {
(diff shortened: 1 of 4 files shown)


Kudu.Services/GitServer/InfoRefsController.cs
```diff
@@ -185,11 +187,29 @@ public void InitialCommitIfNecessary()
...
 
-                // do initial commit
-                repository = _repositoryFactory.EnsureRepository(RepositoryType.Git);
+                // if it is a functionApp
+                // git clone will copy repo from wwwroot/ to repository/
+                // then treate it like its not in place deployment
+                if (!string.IsNullOrEmpty(System.Environment.GetEnvironmentVariable(Constants.FunctionRunTimeVersion)))
+                {
+                    // since git clone doesn't happen very often, create this dynamically instead of static
+                    var ignoreDir = new HashSet<string>();
```
ahmelsayed: maybe use a `HashSet<string>(StringComparer.OrdinalIgnoreCase)` is OS is Windows?

Kudu.Core/Infrastructure/FileSystemHelpers.cs
```diff
@@ -201,7 +201,8 @@ public static void MoveFile(string sourceFileName, string destFileName)
         }
 
         // From MSDN: http://msdn.microsoft.com/en-us/library/bb762914.aspx
-        public static void CopyDirectoryRecursive(string sourceDirPath, string destinationDirPath, bool overwrite = true)
+        // <param name="ignoreDir">string of directory/subdirectory names to be excluded from the copy</param>
+        public static void CopyDirectoryRecursive(string sourceDirPath, string destinationDirPath, bool overwrite = true, HashSet<string> ignoreDir = null)
```
ahmelsayed: nit: ignoreDir**s**?

watashiSHUN:
added unit test

jvano:
Hi

Thanks,

Joaquin Vano
Azure App Service

PR #2423, opened 2017-05-04
selectNodeVersion.js revisions for Linux

nickwalkmsft:
(Looking at adding tests, just getting this out now for an early look)

selectNodeVersion.js is called by kuduscript during node deployments to select node and npm versions used during npm package restore. It contains a lot of logic that isn't relevant on Linux, and is fairly convoluted.

This change bifurcates the majority of the script based on platform, adding a Linux-specific branch with simpler/clearer logic, and fixing problems we're seeing on Linux due to different environment and inputs. It intends to keep Windows behavior completely unchanged.

It looks like github's diff view has a really hard time aligning the two versions; you may want to view in another app.

### DIFF
--- a/Kudu.Core/Infrastructure/PathUtils/PathLinuxUtility.cs
+++ b/Kudu.Core/Infrastructure/PathUtils/PathLinuxUtility.cs
@@ -28,7 +28,12 @@ namespace Kudu.Core.Infrastructure
 
         internal override string ResolveNpmJsPath()
         {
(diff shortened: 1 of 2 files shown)


nickwalkmsft:
After sleeping on it, I will probably be cancelling https://github.com/projectkudu/KuduScript/pull/75 and updating this PR with a couple extra compensating changes, to help ensure continuity of as-is behavior on Windows.

PR #2426, opened 2017-05-05
fix typo in GetUrlFromApplicationHost()

### LINKED ISSUE #2422: Unable to browse site extensions from kudu and portal.azure.com
Hi 

It seems that a recent update have broken the functionality of the siteextensions UI in the kudu portal. 
image
Normally there's a play button that takes you to the homepage of the installed site extension, that is missing. This is on a web app with x64 bit enabled (but also reproducable on 32bit) hosted in North Europe region. 

I think it might also have a related side effect in the portal.azure.com UI, as the browse button is not clickable. 
image

Btw. the site extensions are still browsable, by manually typing in the full url, it is just the UI to browse to them that is not working.

watashiSHUN:
fix #2422
before:
before
now:
now


suwatch:
:shipit:

Issue #2428, opened 2017-05-08
Kudu synctrigger logic doesn't support having 'enabled' set to an environment variable


davidebbo:
It's valid for functions.json to have:

```
{
  "disabled": "MyEnvVariable",
  "bindings": [ ...  ]
}
```

The runtime supports it, but Kudu does not, causing the triggers to get ignored.

And not that `MyEnvVariable` above is correct. It should not be `%MyEnvVariable%`

davidebbo:
Issue is on https://github.com/projectkudu/kudu/blob/5a81fb0659a240c3834d627d23d7b92eb3ff1c28/Kudu.Core/Helpers/PostDeploymentHelper.cs#L278, where it assumes the value is a bool.

suwatch:
This is not a regression, correct?  It is just that we never support it in Kudu and simply need to support (sync up) with runtime.

watashiSHUN:
we want to duplicate [this function code](https://github.com/Azure/azure-webjobs-sdk-script/blob/a9bafba78a3a8092bfd61a8c7093200dae867efb/src/WebJobs.Script/Host/ScriptHost.cs#L1476-L1498) in kudu?

suwatch:
we want to simply do Environment.Expand..   The reason is this code is shared with msdeploy and we need to make sure not pulling any more library than this specific one.

davidebbo:
@suwatch correct, it's not a regression. Just something that had not come up.

Agreed, let's keep it simple, but note that `Environment.Expand` is not correct, since that requires %, and we just have it straight. So it should just be:

- Environment.GetEnvironmentVariable() is not null, use that value
- Else, convert to bool as we do now

watashiSHUN:
in azure function, we consider [`"1" == true`](https://github.com/Azure/azure-webjobs-sdk-script/blob/a9bafba78a3a8092bfd61a8c7093200dae867efb/src/WebJobs.Script/Host/ScriptHost.cs#L1489-L1490) same applies here?

davidebbo:
May as well throw that in!

ahawes-clarity:
Is this fix in production yet?

davidebbo:
Yes, it is.

ahawes-clarity:
I can confirm on my end. Thanks

ahawes-clarity:
[...]

PR #2430, opened 2017-05-09
disabled json field can have environment variable as value

watashiSHUN:
fix issue #2428

random spaces are deleted since my VS auto format my entire file on save

NOTE to self, add unit test

### DIFF
--- a/Kudu.Core/Helpers/PostDeploymentHelper.cs
+++ b/Kudu.Core/Helpers/PostDeploymentHelper.cs
@@ -33,7 +33,7 @@ namespace Kudu.Core.Helpers
         // for mocking or override behavior
         public static Func<HttpClient> HttpClientFactory
         {
-            get {  return _httpClientFactory ?? new Func<HttpClient>(() => new HttpClient()); }
+            get { return _httpClientFactory ?? new Func<HttpClient>(() => new HttpClient()); }
             set { _httpClientFactory = value; }
         }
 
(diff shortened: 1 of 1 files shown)


Kudu.Core/Helpers/PostDeploymentHelper.cs
```diff
@@ -275,11 +275,26 @@ private static void VerifyEnvironments()
...
+                    string stringValue = value.ToString();
+                    bool disabled;
+                    // if "disabled" is not a boolean, we try to expend it as an environment variable
+                    if (!Boolean.TryParse(stringValue, out disabled))
+                    {
+                        string expandValue = System.Environment.GetEnvironmentVariable(stringValue);
+                        // null/"" -> false, "1"/"true" -> true
+                        disabled = string.IsNullOrEmpty(expandValue) ? false :
+                            string.Equals(expandValue, "1", StringComparison.OrdinalIgnoreCase) ? true :
+                            string.Equals(expandValue, "true", StringComparison.OrdinalIgnoreCase);
```
davidebbo: A bit simpler to just write:

    disabled = !string.Equals(expandValue, "1", StringComparison.OrdinalIgnoreCase) && !string.Equals(expandValue, "true", StringComparison.OrdinalIgnoreCase)

watashiSHUN: I changed it to `||`, since "true" means **disabled = true**, not **false**

PR #2437, opened 2017-05-16
fix job traces spaming logfiles folder

suwatch:
This was observed when investigating issues with webjobs.  We traces to logfiles folder when starting trigger and continuous job executables.  This could generate a flood of traces (if those come up/down multiple time such as continuous job that stops frequently).    The side effect is the more important traces will be purged out.   The fix is to migrate those traces to ETW instead.  User can still view jog.log file to see if any issue.


davidebbo:
That's already in.

PR #2439, opened 2017-05-17
use nuget4.3 as default

watashiSHUN:
change does following things
1. add download link to ~[nuget4.1](https://dist.nuget.org/win-x86-commandline/v4.1.0/nuget.exe)~ [nuget4.3](https://dist.nuget.org/win-x86-commandline/v4.3.0-preview1/nuget.exe)
2. ~replace msbuild14 in `PATH` with `D:\Program Files (x86)\MSBuild\15.0\Bin`~
3. ~replace msbuild14 as value of `MSBUILD_PATH` with `D:\Program Files (x86)\MSBuild\15.0\Bin\MSBuild.exe`~
4. ~at startup, copy msbuild installation information to `%ProgramData%\Microsoft\VisualStudio\Packages\_Instances`~

### DIFF
--- a/Build/NuGet.targets
+++ b/Build/NuGet.targets
@@ -62,10 +62,10 @@
                try {
                    OutputFilename = Path.GetFullPath(OutputFilename);
 
-                   Log.LogMessage("Downloading latest version of NuGet.exe...");
                    WebClient webClient = new WebClient();
-                   webClient.DownloadFile("https://dist.nuget.org/win-x86-commandline/v3.5.0/NuGet.exe", OutputFilename);
-
+				   Log.LogMessage("Downloading NuGet.exe version4.3...");
+				   webClient.DownloadFile("https://dist.nuget.org/win-x86-commandline/v4.3.0-preview1/nuget.exe", OutputFilename);
+				   
                    return true;
                }
                catch (Exception ex) {
--- a/Build/NuGet.targets
+++ b/Build/NuGet.targets
@@ -62,10 +62,10 @@
                try {
                    OutputFilename = Path.GetFullPath(OutputFilename);
 
+                   Log.LogMessage("Downloading NuGet.exe version 4.3...");
                    WebClient webClient = new WebClient();
-				   Log.LogMessage("Downloading NuGet.exe version4.3...");
-				   webClient.DownloadFile("https://dist.nuget.org/win-x86-commandline/v4.3.0-preview1/nuget.exe", OutputFilename);
-				   
+                   webClient.DownloadFile("https://dist.nuget.org/win-x86-commandline/v4.3.0-preview1/nuget.exe", OutputFilename);
+
                    return true;
                }
                catch (Exception ex) {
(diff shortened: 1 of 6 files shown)


Kudu.Services.Web/App_Start/NinjectServices.cs
```diff
@@ -76,6 +76,14 @@ public static void Start()
             HttpApplication.RegisterModule(typeof(NinjectHttpModule));
             HttpApplication.RegisterModule(typeof(TraceModule));
             _bootstrapper.Initialize(CreateKernel);
+
+            // if ProgramData already have the msbuild15/state.json,do nothing
+            // else copy state.json for msbuild15 discovery purpose
+            string dest = $"{VfsSpecialFolders.ProgramDataPath}\\Microsoft\\VisualStudio\\Packages\\_Instances";
```
nickwalkmsft: Don't run this on Linux. I don't think it'll break anything, but better to be explicit.

watashiSHUN:
@davidebbo, I suggest we yank the related tests? I was thinking of testing the error but since we are removing them at some time in the future anyway, might as well...

davidebbo:
Yep, let's yank those tests.

Build/NuGet.targets
```diff
@@ -62,10 +62,10 @@
                try {
                    OutputFilename = Path.GetFullPath(OutputFilename);
 
-                   Log.LogMessage("Downloading latest version of NuGet.exe...");
                    WebClient webClient = new WebClient();
-                   webClient.DownloadFile("https://dist.nuget.org/win-x86-commandline/v3.5.0/NuGet.exe", OutputFilename);
-
+				   Log.LogMessage("Downloading NuGet.exe version4.3...");
```
davidebbo: Looks like your editor is producing tabs.

davidebbo: Also, missing space in 'version4.3'.

Kudu.Core/Deployment/Generator/AspNetCoreBuilder.cs
```diff
@@ -21,6 +22,10 @@ public AspNetCoreBuilder(IEnvironment environment, IDeploymentSettingsManager se
             else if (_projectPath.EndsWith(".xproj", StringComparison.OrdinalIgnoreCase))
             {
                 _version = "xproj";
+                // if it's xproj, throw invalidOperationException
+                throw new InvalidOperationException(String.Format(CultureInfo.CurrentCulture,
+                                                             Resources.Error_ProjectNotDeployable,
+                                                             projectPath));
             }
             else
             {
```
davidebbo: Shouldn't project.json case throw as well?

Kudu.Core/Deployment/Generator/AspNetCoreBuilder.cs
```diff
@@ -21,6 +22,10 @@ public AspNetCoreBuilder(IEnvironment environment, IDeploymentSettingsManager se
             else if (_projectPath.EndsWith(".xproj", StringComparison.OrdinalIgnoreCase))
             {
                 _version = "xproj";
```
davidebbo: No need to set `_version` if it throws.

PR #2444, opened 2017-05-23
Add API to download function app archive that can add local.settings.…

ahmelsayed:
…json and csproj file to the archive.

This is needed for https://github.com/Azure/azure-functions-ux/issues/985

### DIFF
--- a/Kudu.Core/Functions/FunctionManager.cs
+++ b/Kudu.Core/Functions/FunctionManager.cs
@@ -1,8 +1,11 @@
 ﻿using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Configuration;
 using System.IO;
+using System.IO.Compression;
 using System.Linq;
+using System.Text;
 using System.Threading.Tasks;
 using Kudu.Contracts.Functions;
 using Kudu.Contracts.Tracing;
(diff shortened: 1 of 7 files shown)


Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -422,5 +425,113 @@ private string GetFunctionSecretsFilePath(string functionName)
...
+        /// It can also include local.settings.json and .csproj files for a full Visual Studio project.
+        /// sln file is not included since it changes between VS versions and VS can auto-generate it from the csproj.
+        /// All existing functions are added as content with "Always" copy to output.
+        /// </summary>
+        /// <param name="zip">the <see cref="ZipArchive"/> to be populated with function app content.</param>
+        /// <param name="includeAppSettings">Optional: indicates whether to add local.settings.json or not to the archive. Default is false.</param>
+        /// <param name="includeCsproj">Optional: indicates whether to add a .csproj to the archive. Default is false.</param>
+        /// <param name="projectName">Optional: the name for *.csproj file if <paramref name="includeCsproj"/> is true. Default is appName.</param>
+        /// <returns></returns>
+        public IList<ZipArchiveEntry> CreateArchive(ZipArchive zip, bool includeAppSettings = false, bool includeCsproj = false, string projectName = null)
```
davidebbo: Not sure I understand this code. This returns a `IList<ZipArchiveEntry>`, but I don't see any code actually using that.

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -422,5 +425,113 @@ private string GetFunctionSecretsFilePath(string functionName)
...
+        /// it also contain 'Always' copy entries for all the files that are in wwwroot.
+        /// </summary>
+        /// <param name="zip"><see cref="ZipArchive"/> to add csproj file to.</param>
+        /// <param name="files"><see cref="IEnumerable{ZipArchiveEntry}"/> of entries in the zip file to include in the csproj.</param>
+        /// <param name="projectName">the {projectName}.csproj</param>
+        private static ZipArchiveEntry AddCsprojFile(ZipArchive zip, IEnumerable<ZipArchiveEntry> files, string projectName)
+        {
+            const string microsoftAzureWebJobsVersion = "2.1.0-beta1";
+            const string microsoftAzureWebJobsExtensionsHttpVersion = "1.0.0-beta1";
+            const string microsoftNETSdkFunctionsVersion = "1.0.0-alpha4";
```
ahmelsayed: that's true. I guess an alternative would be to query nuget.org for the latest. 

VS would work fine, it just might be using an outdated build task.

Kudu.Core/Functions/FunctionManager.cs
ahmelsayed: yes, it's not used. I was validating it in a unit test as I wasn't able to mock ZipArchive, I'll push the test shortly

Kudu.Services/Functions/FunctionController.cs
```diff
@@ -179,6 +181,21 @@ public HttpResponseMessage Delete(string name)
             }
         }
 
+        [HttpGet]
+        public HttpResponseMessage DownloadFunctions(bool includeCsproj = true, bool includeAppSettings = false)
+        {
+            var tracer = _traceFactory.GetTracer();
+            using (tracer.Step($"{nameof(FunctionController)}.{nameof(DownloadFunctions)}({includeCsproj}, {includeAppSettings})"))
```
ahmelsayed: right. The current proposal is on https://github.com/Azure/azure-functions-ux/issues/985. it looks like this
image

PR #2446, opened 2017-05-24
Roundtrip WebHooks of all types when firing one hook. Fixes #2445

### LINKED ISSUE #2445: Firing a WebHook causes hooks of a different type to be deleted
Steps:

- Create both a PostDeployment and a TriggeredJobFinished webhook
- Trigger a deployment so the hook gets fired

Results: as a side effect, the WebHook is deleted.

davidebbo:


suwatch:
:shipit:

PR #2455, opened 2017-06-01
Add new WEBJOBS_DISABLE_SCHEDULE flag

davidebbo:
Was straightforward enough and seems like a valid need when dealing with slots. The cron schedule is disabled, but the job can still be triggered manually for testing.


suwatch:
:shipit:

xt0rted:
I needed to turn off our triggered jobs for a few hours but `WEBJOBS_STOPPED` was disabling the appinsights job too. Using `WEBJOBS_DISABLE_SCHEDULE` let me turn mine off while leaving appinsights on. Thanks for adding this option in.

PR #2467, opened 2017-06-07
Skip ssl verification in Azure test env

suwatch:
In Azure test env, there exists registry to indicate whether to skip ssl verification.  We will use this as default if none is defined (appSettings).   This simplifies testing on Azure test env.   In addition, it is also consistent with Web Deploy and Scaling in Azure Functions Runtime.


suwatch:
Merged

Issue #2442, opened 2017-05-22
Should include traces and its details to ETW even requests from browsers


suwatch:
For starter, we could beginning doing it for non GET to reduce the noises.

PR #2468, opened 2017-06-07
Always enable ETW for request tracer

suwatch:
Fix for https://github.com/projectkudu/kudu/issues/2442.   For requests from browser, we currently disable file tracer.   This change is to always write to ETW.


suwatch:
merged

PR #2470, opened 2017-06-07
Polling FileSystemWatcher for log streaming on Linux

nickwalkmsft:


Kudu.Services/Diagnostics/LogStreamManager.cs
```diff
@@ -142,13 +143,14 @@ private void Initialize(string path)
 
             if (_watcher == null)
             {
-                FileSystemWatcher watcher = new FileSystemWatcher(path);
+                IFileSystemWatcher watcher = OSDetector.IsOnWindows() 
+                    ? (IFileSystemWatcher)new FileSystemWatcherWrapper(path, true)
```
davidebbo: Include flag name on call for readability (i.e. `includeSubdirectories: true`)

Kudu.Core/Infrastructure/FileSystemWatcherWrapper.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        public void Stop()
+        {
+            this.inner.EnableRaisingEvents = false;
+        }
+
+        public string Path
+        {
+            get => this.inner.Path;
```
davidebbo: I think our kudu-ci machine is not yet set up for C# 6 so that'll probably break the build.

Kudu.Core/Infrastructure/FileSystemWatcherWrapper.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        public event FileSystemEventHandler Deleted
+        {
+            add
+            {
+                lock (eventLock)
```
davidebbo: Why is this lock necessary, given that previous code was directly acting on the underlying FileSystemWatcher?

Kudu.Core/Infrastructure/NaiveFileSystemWatcher.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        private async Task Poll()
+        {
+            while (!cts.IsCancellationRequested)
+            {
+                var handler = Changed;
+                if (handler != null)
+                {
+                    var filepaths = Directory.EnumerateFiles(Path, "*.*", SearchOption.AllDirectories)
+                        .Where(p => logFileExtensions.Any(ext => ext == System.IO.Path.GetExtension(p)));
```
ahmelsayed: do you care about case sensitivity for extensions?

Kudu.Core/Infrastructure/NaiveFileSystemWatcher.cs
```diff
@@ -0,0 +1,81 @@
...
+
+                    foreach (var filepath in filepaths)
+                    {
+                        var dir = System.IO.Path.GetDirectoryName(filepath);
+                        var name = System.IO.Path.GetFileName(filepath);
```
ahmelsayed: nit: you already have a using for System.IO

Kudu.Services/Diagnostics/LogStreamManager.cs
ahmelsayed: is the type cast needed?

Kudu.Core/Infrastructure/FileSystemWatcherWrapper.cs
ahmelsayed: C# 7

davidebbo: Right, 7.

Kudu.Core/Infrastructure/FileSystemWatcherWrapper.cs
nickwalkmsft: Locking in add/remove is a best practice in general, but you're right, it's not needed here.

Kudu.Core/Infrastructure/NaiveFileSystemWatcher.cs
nickwalkmsft: Collides with Path property

Kudu.Services/Diagnostics/LogStreamManager.cs
nickwalkmsft: It is

Kudu.Core/Infrastructure/NaiveFileSystemWatcher.cs
ahmelsayed: got it

Kudu.Services/Diagnostics/LogStreamManager.cs
davidebbo: I understand why, but it should be smarter! :)

PR #2473, opened 2017-06-09
Throttle FunctionController calls

davidebbo:
Not as fancy as doing caching, but simpler.

### DIFF
--- a/Kudu.Services/Filters/RequestThrottleActionFilterAttribute.cs
+++ b/Kudu.Services/Filters/RequestThrottleActionFilterAttribute.cs
@@ -1,0 +1,36 @@
+﻿using System;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Threading;
+using System.Web.Http.Controllers;
+using System.Web.Http.Filters;
+
+namespace Kudu.Services.Filters
+{
+    public sealed class RequestThrottleActionFilterAttribute : ActionFilterAttribute
+    {
+        int _concurrency = 0;
+
+        public int MaxConcurrency { get; set; } = 5;
+
+        public override void OnActionExecuting(HttpActionContext actionContext)
+        {
+            // If we've already reached the concurrency limit, return 429 (too many requests)
+            // and ask to retry after 5 seconds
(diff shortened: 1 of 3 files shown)


davidebbo:
Note that we want want to solve this at a different level like geo. Only having this here for reference in case we go down that route.

Kudu.Services/Filters/RequestThrottleActionFilterAttribute.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Web.Http.Controllers;
+using System.Web.Http.Filters;
+
+namespace Kudu.Services.Filters
+{
+    public sealed class RequestThrottleActionFilterAttribute : ActionFilterAttribute
+    {
+        int _concurrency = 0;
+
+        public int MaxConcurrency { get; set; } = 5;
```
suwatch: Wonder if this should be appSettings with upper limit at, say, 20 pending requests.  I understand bad usages can abuse it (they can anyway by deploying private kudu).  On the flip side, it gives us a quick knob to mitigate if regression.     We can ship with appSettings and then look at log if normal usage will ever get to this and make it const.

Kudu.Services/Filters/RequestThrottleActionFilterAttribute.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public override void OnActionExecuting(HttpActionContext actionContext)
+        {
+            // If we've already reached the concurrency limit, return 429 (too many requests)
+            // and ask to retry after 5 seconds
+            if (_concurrency >= MaxConcurrency)
+            {
+                actionContext.Response = new HttpResponseMessage((HttpStatusCode)429);
```
suwatch: In the future, we will need to inspect how many sites fall into this.   I don't know if 429  from IISLog is obvious enough.   We might consider ETW tracing `once` when we get it here and reset that state when it is back to normal.

```csharp
TraceServices.CurrentRequestTracer?.Trace(...);
```

davidebbo:
Let's not do this since ARM level throttling feels like the better solution.

PR #2492, opened 2017-06-23
remove redundant ajax call when select in file explorer

watashiSHUN:
current behavior when you click in file explorer:
capture
new behavior:
image

@ahmelsayed 
while I am working on the front end routing...
I would suggest that we move the logic to **notify command line from explorer** out of the [selected observable.subscribe()](https://github.com/projectkudu/kudu/blob/master/Kudu.Services.Web/Content/Scripts/FileBrowser.js#L519-L528) (asking for approval :))

for file browser functions, we can notify command line explicitly, for example
`notify command line; selected(new value)`

and when command line wants to notify the file explorer about directory change, we just do
`select(new value)`

so we don't need the [flag variable](https://github.com/projectkudu/kudu/blob/master/Kudu.Services.Web/Content/Scripts/FileBrowser.js#L502) and the [extra promise](https://github.com/projectkudu/kudu/blob/master/Kudu.Services.Web/Content/Scripts/FileBrowser.js#L333)...and make the code readable

### DIFF
--- a/Kudu.Services.Web/Content/Scripts/FileBrowser.js
+++ b/Kudu.Services.Web/Content/Scripts/FileBrowser.js
@@ -6,7 +6,7 @@ var statusbar = {
             try {
                 filename = viewModel.fileEdit.peek().name();
             }
-            catch(e) {
+            catch (e) {
                 filename = 'Can not get filename. See console for details.';
                 if (typeof console == 'object') {
                     console.error('Can not get filename: %s', e);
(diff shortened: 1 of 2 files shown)


davidebbo:
Not familiar with that part of the code, so I'll let others review this one.

Note: you can go to https://github.com/projectkudu/kudu/pull/2492/files?w=1 to get a much smaller diff that ignores white space changes.

ahmelsayed:
@watashiSHUN I don't have any objections. It'll be great if you can making that code more readable even if just a little bit. the communication between the console and the fileBrowser is messy

Kudu.Services.Web/Content/Scripts/FileBrowser.js
```diff
@@ -368,8 +360,7 @@ $.connection.hub.start().done(function () {
             viewModel.editText('');
             statusbar.fetchingContents();
             viewModel.fileEdit(this);
-            if(this.mime === "text/xml")
-            {
+            if (this.mime === "text/xml") {
                 Vfs.getContent(this)
                    .done(function (data) {
                        viewModel.editText(vkbeautify.xml(data));
```
snobu: Do we still need to vkbeautify XML? I believe the Ace editor is already doing that if the extension is .xml (loads the XML mode). I'll test and confirm. vkbeautify was historically used here just to get a readable XML in the (old) editor, right?

snobu: Nope, i was wrong. Setting XML mode in Ace does not beautify the content, only enables coloring. Let's just leave that in.

davidebbo:
@watashiSHUN go ahead and get this in if you think it's all done.

watashiSHUN:
06e0777b8d9ec2353f1b4fcc1efd89ad45bc2d4e

PR #2500, opened 2017-07-10
Change `Value` property to `Values` in generated local.settings.json

anthonychu:
Reference: https://docs.microsoft.com/en-us/azure/azure-functions/functions-run-local#local-settings-file


anthonychu:
Can someone please take a look at this PR? Thanks. cc @davidebbo

Issue #2501, opened 2017-07-11
Upload folder containing more than 100 files via KUDU will lost files


MSLuyi:
When uploading folder containing more than 100 files, only got 100 files copied. 
However, uploading the files directly into a folder, works fine.

watashiSHUN:
if you [unzip](https://github.com/projectkudu/kudu/wiki/Kudu-console#upload-and-expand-zip-file) the folder, it works?

MSLuyi:
Yes, this works. Thanks.

But if upload an unzipped file, it will only copy 100 files ?

watashiSHUN:
for uploading a folder containing large amount of files, I'd suggest you submit a zip folder as it's a lot faster, I need to investigate about what happened in the regular folder case

ahmelsayed:
looks like we're using [`readEntries`](https://github.com/projectkudu/kudu/blob/master/Kudu.Services.Web/Content/Scripts/FileBrowser.js#L838) incorrectly. it returns 100 files at a time and you need to keep calling it until you get an empty array for files.

suwatch:
@ahmelsayed  so it is a bug which we can decide whether or not to fix it, correct?

MSLuyi:
Seems there is also limitation uploading a zip file, it can only copy 299 files.

MSLuyi:
Double verified from FileZilla, there are actually more than 299 files (actually 360 files) in the folder and it had been successfully uploaded, but the KUDU just displayed 299 files.

davidebbo:
@MSLuyi that is unrelated to this issue. There is no limit for the number of files in a zip file, and what you're seeing is just the *display* limit, which is by design.

MSLuyi:
@davidebbo Thanks for the confirmation. My bad for the confusion.

jvano:
Hi

If the problem persists and is related to running it on Azure App Service, please open a support incident in Azure:
https://learn.microsoft.com/en-us/azure/azure-portal/supportability/how-to-create-azure-support-request

This way we can better track and assist you on this case

Thanks,

Joaquin Vano
Azure App Service

PR #2502, opened 2017-07-11
recursively call readEntries

watashiSHUN:
fix issue #2501
click [here](https://github.com/projectkudu/kudu/pull/2502/files?w=1) to see changes without couting whitspaces

### DIFF
--- a/Kudu.Services.Web/Content/Scripts/FileBrowser.js
+++ b/Kudu.Services.Web/Content/Scripts/FileBrowser.js
@@ -1,67 +1,67 @@
 // Custom status bar for Ace (aka Project Wunderbar)
 var statusbar = {
     showFilename:
-        function () {
-            var filename;
-            try {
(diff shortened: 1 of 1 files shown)


watashiSHUN:
I think we can check to see if the array size if < 100, we can avoid a second empty check...
but looking at the [documentation](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemDirectoryReader/readEntries), I am not sure if the upper bound is always gona stay at 100:

> If there are no files left, or you've already called readEntries() on this FileSystemDirectoryReader, the array is empty.

watashiSHUN:
https://github.com/projectkudu/kudu/commit/8b9ad1d0df7671273cb90fe84ec2ccac826683ed
fairly simple change, I have tested it
merged since I hit this bug myself today
 when pasting private function runtime in kudu....(238 filesin a folder)

PR #2525, opened 2017-08-17
relax project GUID check for .sln

watashiSHUN:
the functionApp created using VS2017 15.3 changed the project GUID to C# 
I decide to relax the GUID check, even if it returns a `SolutionProjectType.Unknown` 
we still let it go through our logic, if we still ended up finding a project to deploy
we continue with build and let build error out

### DIFF
--- a/Kudu.Core/Deployment/Generator/SiteBuilderFactory.cs
+++ b/Kudu.Core/Deployment/Generator/SiteBuilderFactory.cs
@@ -87,6 +87,7 @@ namespace Kudu.Core.Deployment.Generator
             // figure out with some heuristic, which one to deploy.
 
             // TODO: Pick only 1 and throw if there's more than one
+            // shunTODO need to implement this
             VsSolutionProject project = solution.Projects.Where(p => p.IsWap || p.IsWebSite || p.IsAspNetCore || p.IsFunctionApp).FirstOrDefault();
 
             if (project == null)
@@ -105,6 +106,8 @@ namespace Kudu.Core.Deployment.Generator
 
                 logger.Log(Resources.Log_NoDeployableProjects, solution.Path);
 
+                // we have a solution file, but no deployable project
+                // shunTODO how often do we run into this
                 return ResolveNonAspProject(repositoryRoot, null, settings);
             }
 
@@ -149,8 +152,7 @@ namespace Kudu.Core.Deployment.Generator
         private ISiteBuilder ResolveNonAspProject(string repositoryRoot, string projectPath, IDeploymentSettingsManager perDeploymentSettings)
         {
             string sourceProjectPath = projectPath ?? repositoryRoot;
-            // "FUNCTIONS_EXTENSION_VERSION" environment variable implies a functionApp
-            if (FunctionAppHelper.LooksLikeFunctionApp())
+            if (IsFunctionApp(sourceProjectPath))
             {
                 return new FunctionBasicBuilder(_environment, perDeploymentSettings, _propertyProvider, repositoryRoot, projectPath);
             }
@@ -176,6 +178,12 @@ namespace Kudu.Core.Deployment.Generator
             }
 
             return new BasicBuilder(_environment, perDeploymentSettings, _propertyProvider, repositoryRoot, projectPath);
+        }
+
+        private static bool IsFunctionApp(string projectPath)
+        {
+            // projectPath ==> project folder
+            return FunctionAppHelper.LooksLikeFunctionApp(projectPath);
         }
 
         private static bool IsGoSite(string projectPath)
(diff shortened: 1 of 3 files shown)


Kudu.Core/Deployment/Generator/SiteBuilderFactory.cs
```diff
@@ -149,8 +152,7 @@ public ISiteBuilder CreateBuilder(ITracer tracer, ILogger logger, IDeploymentSet
         private ISiteBuilder ResolveNonAspProject(string repositoryRoot, string projectPath, IDeploymentSettingsManager perDeploymentSettings)
         {
             string sourceProjectPath = projectPath ?? repositoryRoot;
-            // "FUNCTIONS_EXTENSION_VERSION" environment variable implies a functionApp
-            if (FunctionAppHelper.LooksLikeFunctionApp())
+            if (IsFunctionApp(sourceProjectPath))
```
davidebbo: Why move this into an additional level of helper if all it does is call `FunctionAppHelper.LooksLikeFunctionApp`?

Kudu.Core/Infrastructure/FunctionAppHelper.cs
```diff
@@ -1,10 +1,16 @@
-﻿namespace Kudu.Core.Infrastructure
+﻿using System.IO;
+
+namespace Kudu.Core.Infrastructure
 {
     internal static class FunctionAppHelper
     {
-        public static bool LooksLikeFunctionApp()
+        const string hostJson = "host.json";
+        public static bool LooksLikeFunctionApp(string projectFolder)
```
davidebbo: We have two kind of function apps: VS precompiled and script based. Does `LooksLikeFunctionApp` refer to both, or just one of the two? If only one, then method name should be more explicit.

Kudu.Core/Infrastructure/VsSolutionProject.cs
```diff
@@ -190,7 +179,7 @@ private void EnsureProperties()
         private enum SolutionProjectType
         {
             Unknown,
-            KnownToBeMSBuildFormat,
+            KnownToBeMSBuildFormat,  // KnownToBeMSBuildFormat: C#, VB, and VJ# projects
```
davidebbo: So it seems we never check for this one anymore?

Kudu.Core/Deployment/Generator/SiteBuilderFactory.cs
watashiSHUN: it used to be like this, I changed in the last commit, because it doesn't take a `projectPath` as argument like its siblings https://github.com/projectkudu/kudu/blob/5ab5fbcf795a24340b7b80ce694ad5dc66a907d1/Kudu.Core/Deployment/Generator/SiteBuilderFactory.cs#L196-L199

but since we don't throw error when there's multiple deployable project
I think we need to check for "host.json" so the logic is more strict 
which bring it back to the old pattern, now it stand side by side as others:
https://github.com/projectkudu/kudu/blob/5da2e47f725799ed432c91cd644f7ba48d89be51/Kudu.Core/Deployment/Generator/SiteBuilderFactory.cs#L189-L192

Kudu.Core/Infrastructure/FunctionAppHelper.cs
watashiSHUN: both, since we are checking for "host.json" and environment variable, not `.sln` `csproj` or `index.js`

watashiSHUN:
https://github.com/projectkudu/kudu/blob/5da2e47f725799ed432c91cd644f7ba48d89be51/Kudu.Core/Deployment/Generator/SiteBuilderFactory.cs#L303-L339

when there's a project file provided, we deploy the project using this order
with a solution, we can do the same thing
if the solution has multiple projects, we will deploy the first found (higher up in the `.sln`)

this requires a rewrite, probably not going to be in this pull request
instead of doing: https://github.com/projectkudu/kudu/blob/5da2e47f725799ed432c91cd644f7ba48d89be51/Kudu.Core/Deployment/Generator/SiteBuilderFactory.cs#L114-L150

we just repeatedly call `DetermineProject` for each project found in .sln
so we can avoid to have duplicate logics

Kudu.Core/Infrastructure/VsSolutionProject.cs
```diff
@@ -161,22 +161,11 @@ private void EnsureProperties()
...
-                }
-                else if (FunctionAppHelper.LooksLikeFunctionApp())
-                {
-                    _isFunctionApp = true;
-                }
+
+                _isAspNetCore = AspNetCoreHelper.IsDotnetCoreFromProjectFile(_absolutePath, _projectTypeGuids);
+                _isWap = VsHelper.IsWap(_projectTypeGuids);
+                _isExecutable = VsHelper.IsExecutableProject(_absolutePath);
+                _isFunctionApp = FunctionAppHelper.LooksLikeFunctionApp(Path.GetDirectoryName(_absolutePath));
```
watashiSHUN: 1.`isAspNetCore` check if the csproj has no Guid
2.`isWap` check if the  csproj contains a particular Guid
3.`isExecutable` check if the csproj has outputtype == exe
4.`isFunctionApp` check if the project has host.json and environment variable `FUNCTION_EXTENSION_VERSION`

1,4 or 1,3 can be true at the same time, but neither of them have our support in Azure

PR #2526, opened 2017-08-18
Docker logs API

nickwalkmsft:
Dedicated endpoints for retrieving current Docker logs on Linux. This has been requested for GA.

**This is an early review** - mostly feeling out if anyone has a huge hole to shoot in this. There will be more changes and potentially functionality before this is complete.

Open to suggestions for the route URLs, the controller that this logic should go in, etc.

### DIFF
--- a/Kudu.Services/Diagnostics/DiagnosticsController.cs
+++ b/Kudu.Services/Diagnostics/DiagnosticsController.cs
@@ -13,11 +13,18 @@ using Kudu.Core.Settings;
 using Kudu.Core.Tracing;
 using Kudu.Services.Diagnostics;
 using Kudu.Services.Infrastructure;
+using Newtonsoft.Json.Linq;
+using System.Linq;
+using System.Text.RegularExpressions;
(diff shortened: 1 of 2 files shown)


Kudu.Services/Diagnostics/DiagnosticsController.cs
```diff
@@ -66,6 +73,89 @@ public HttpResponseMessage GetRecentLogs(int top = 100)
...
+        {
+            //TODO get from environment
+            var path = "/home/LogFiles";
+
+            var nonRolledDockerLogFilenames =
+                FileSystemHelpers.GetFiles(path, "*", SearchOption.TopDirectoryOnly)
+                .Where(f => NONROLLED_DOCKER_LOG_FILENAME_REGEX.IsMatch(Path.GetFileName(f)))
+                .ToArray();
+
+            // Find the latest date stamp and filter out those that don't have it
```
davidebbo: Example file name in comment

PR #2527, opened 2017-08-21
Mitigation for Process API errors on Linux

nickwalkmsft:
Effectively disabling Kudu's process API on Linux, as current implementation doesn't work.
Returning 400's as suggested by Suwat.
Will revisit in the future.

### DIFF
--- a/Kudu.Services/Diagnostics/LinuxProcessController.cs
+++ b/Kudu.Services/Diagnostics/LinuxProcessController.cs
@@ -11,77 +11,76 @@ namespace Kudu.Services.Performance
 
     public class LinuxProcessController : ApiController
     {
-        private const HttpStatusCode RETURNCODE = HttpStatusCode.BadRequest;
         private const string ERRORMSG = "Not supported on Linux";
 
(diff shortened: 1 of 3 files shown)


Kudu.Services/Diagnostics/LinuxProcessController.cs
```diff
@@ -0,0 +1,87 @@
...
+
+namespace Kudu.Services.Performance
+{
+    // This is a placeholder for future process API functionality on Linux,
+    // the implementation of which will differ from Windows enought that it warrants
+    // a separate controller class. For now this returns 400s for all routes.
+
+    public class LinuxProcessController : ApiController
+    {
+        private const HttpStatusCode RETURNCODE = HttpStatusCode.BadRequest;
```
davidebbo: nit: This constant has limited value IMO. Code is more readable if you directly use `HttpStatusCode.BadRequest` everywhere.

nickwalkmsft: Thanks! Will fix before commit.

PR #2529, opened 2017-08-23
Linux - Support for container restart trigger

nickwalkmsft:
Linux App Service now supports a generic file-based communication mechanism
for requesting a restart of the app container. This was previously implemented
in a limited fashion only for the "Docker CI" webhook. This generalizes the
implementation and also requests a restart after a git deployment by default,
in addition to receipt of a Docker webhook request.

### DIFF
--- a/Kudu.Contracts/Settings/DeploymentSettingsExtension.cs
+++ b/Kudu.Contracts/Settings/DeploymentSettingsExtension.cs
@@ -218,5 +218,18 @@ namespace Kudu.Contracts.Settings
         {
             return settings.GetValue(SettingsKeys.TouchWebConfigAfterDeployment) != "0";
         }
+
+        public static bool IsDockerCiEnabled(this IDeploymentSettingsManager settings)
+        {
+            string value = settings.GetValue(SettingsKeys.DockerCiEnabled);
+            return StringUtils.IsTrueLike(value);
+        }
+
+        public static bool RestartAppContainerOnGitDeploy(this IDeploymentSettingsManager settings)
+        {
+            // Default is true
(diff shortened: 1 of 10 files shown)


ahmelsayed:
Looks good to me :shipit:

Kudu.Core/Resources.resx
```diff
@@ -1,228 +1,231 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
davidebbo: Strange that it's showing a full file diff. Maybe some line ending issue?

Kudu.Contracts/Settings/DeploymentSettingsExtension.cs
```diff
@@ -218,5 +218,18 @@ public static bool TouchWebConfigAfterDeployment(this IDeploymentSettingsManager
...
+        {
+            string value = settings.GetValue(SettingsKeys.DockerCiEnabled);
+            return StringUtils.IsTrueLike(value);
+        }
+
+        public static bool RestartAppContainerOnGitDeploy(this IDeploymentSettingsManager settings)
+        {
+            // Default is true
+            string value = settings.GetValue(SettingsKeys.LinuxRestartAppContainerAfterDeployment) ?? "true";
+            return StringUtils.IsTrueLike(value);
```
davidebbo: nit: slightly ugly to hard code "true" and then parse it. `return value == null || StringUtils.IsTrueLike(value);` is a bit cleaner.

nickwalkmsft: Ah thank you, I thought this smelled funny too. I think I wrote this when the coffee machine was broken.

PR #2531, opened 2017-08-25
use node template for njsproj and python template for pyproj

watashiSHUN:
using `KnownToBeMSBuildFormat` will exclude preview projects (function app for example)
but removing the check, will include `pyproj`/`njsproj` (which are still VS webproj, but are built differently in Azure)

now we simply check for file extension, and it happened that there is already one function that does it
although I renamed it
now https://github.com/projectkudu/kudu/blob/5da2e47f725799ed432c91cd644f7ba48d89be51/Kudu.Core/Deployment/Generator/SiteBuilderFactory.cs#L303
and 
https://github.com/projectkudu/kudu/blob/5da2e47f725799ed432c91cd644f7ba48d89be51/Kudu.Core/Infrastructure/VsSolutionProject.cs#L166

2 places where we check for webproj are both guarded by `IsMsBuildProject`

### DIFF
--- a/Kudu.Core/Deployment/DeploymentHelper.cs
+++ b/Kudu.Core/Deployment/DeploymentHelper.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
-using System.IO.Abstractions;
 using System.Linq;
 using Kudu.Core.Infrastructure;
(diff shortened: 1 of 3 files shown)


watashiSHUN:
rebased and added test: 41c3f177e4355e7b7efd5ee2077b208d0fc38c3f

PR #2539, opened 2017-08-31
Support Dropbox v2 api

suwatch:
This is to support synchng using OAuth 2.0 and Dropbox v2 APIs.

### DIFF
--- a/Kudu.Services/ServiceHookHandlers/DropboxHandler.cs
+++ b/Kudu.Services/ServiceHookHandlers/DropboxHandler.cs
@@ -13,6 +13,7 @@ namespace Kudu.Services.ServiceHookHandlers
 {
     public class DropboxHandler : IServiceHookHandler
     {
+        private const string DropboxVersionKey = "dropbox_version";
         private const string DropboxTokenKey = "dropbox_token";
         private const string DropboxPathKey = "dropbox_path";
         private readonly DropboxHelper _dropBoxHelper;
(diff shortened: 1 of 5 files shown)


Kudu.Services/ServiceHookHandlers/DropboxHandler.cs
```diff
@@ -37,7 +38,7 @@ public DeployAction TryParseDeploymentInfo(HttpRequestBase request, JObject payl
                 dropboxInfo = DropboxInfo.CreateV1Info(payload, GetRepositoryType());
                 message = String.Format(CultureInfo.CurrentUICulture, Resources.Dropbox_SynchronizingNChanges, dropboxInfo.DeployInfo.Deltas.Count);
             }
-            else if (String.Equals(payload.Value<string>("scmType"), "DropboxV2", StringComparison.OrdinalIgnoreCase))
+            else if (String.Equals(payload.Value<string>(DropboxVersionKey), "2", StringComparison.OrdinalIgnoreCase))
```
suwatch: Since we will not (don't need to) introduce a new ScmType, I explicitly using dropbox_version in the payload to signify the different version.

nickwalkmsft:
Can you summarize a bit?

This looks mostly like modifications as opposed to additions; is any functionality being lost, or are we currently compatible with the v2 API but just not using it to the fullest? If it's the latter, what does this add?

Kudu.Services/ServiceHookHandlers/DropboxHandler.cs
nickwalkmsft: Is no one using an scmType of `DropboxV2`?

suwatch: Correct.   BTW, this was done long time ago (but not completely correct) to anticipate the Dropbox V2.   Changing ScmType is unnecessary and may involve more codes than we need.

suwatch:
A couple years back when Dropbox V2 was available, we added V2 support in Kudu in anticipation of Antares change.   However, the way we implemented was not entirely correct.   

Dropbox V2 introduced the new OAuth 2.0 support as well as a set of new V2 REST apis.   Our implementation only adopted the OAuth 2.0 but still using the V1 REST apis.   Note that Dropbox V1 api supports both OAuth 1.0 and 2.0 tokens while Dropbox V2 only supports OAuth 2.0.

My change is to embrace both OAuth 2.0 and V2 REST api - so it looks more like a modification than a full addition.

davidebbo:
@suwatch isn't this already in?

PR #2544, opened 2017-08-31
Linux: Fix links in VFS API

nickwalkmsft:
On Linux, absolute URLs to Kudu resources cannot be based on
request.RequestUri, as it's altered by the way the request is handed off within
the worker. This fixes the absolute URL links generated by the VFS API.

No changes on Windows; exact same calls are made.


nickwalkmsft:
Also fixes redirects, like when you browse to `/api/vfs` without a trailing slash and it redirects you to `/api/vfs/`

nickwalkmsft:
Can anyone take a look at this real quick? :eyes:

suwatch:
Looks good.

PR #2545, opened 2017-09-04
Update JobsManager to watch App_Data folder 

northtyphoon:
If we set up FileSystemWatcher to watch App_Data\jobs\continuous, it will hold the handler to App_Data\jobs\continuous. When we try to delete App_Data\jobs, it may block the deletion. It usually happens when WebDeploy tries to clean up App_Data folder.
The change sets up FileSystemWatcher to watch App_Data folder, it hold the handler to App_Data folder and still includes the sub directory changes.

@davidebbo @pharring @iusafaro

### DIFF
--- a/Kudu.Core/Jobs/JobsFileWatcher.cs
+++ b/Kudu.Core/Jobs/JobsFileWatcher.cs
@@ -15,8 +15,8 @@ namespace Kudu.Core.Jobs
     {
         private readonly ITraceFactory _traceFactory;
         private readonly IAnalytics _analytics;
-        private const int TimeoutUntilMakingChanges = 5*1000;
-        private const int CheckForWatcherTimeout = 30*1000;
+        private const int TimeoutUntilMakingChanges = 5 * 1000;
+        private const int CheckForWatcherTimeout = 30 * 1000;
 
         private HashSet<string> _updatedJobs = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
@@ -28,14 +28,16 @@ namespace Kudu.Core.Jobs
         private int _pendingOnError;
 
         private bool _makingChanges;
-        private readonly string _watchedDirectoryPath;
+        private readonly string _watchedJobOrParentDirectoryPath;
+        private readonly string _jobDirectoryPath;
         private readonly Action<string> _onJobChanged;
         private readonly string _filter;
         private readonly Func<bool, IEnumerable<string>> _listJobNames;
         private readonly string _jobType;
 
         public JobsFileWatcher(
-            string watchedDirectoryPath,
+            string watchedJobOrParentDirectoryPath,
+            string jobDirectoryPath,
             Action<string> onJobChanged,
             string filter,
             Func<bool, IEnumerable<string>> listJobNames,
@@ -45,7 +47,8 @@ namespace Kudu.Core.Jobs
         {
             _traceFactory = traceFactory;
             _analytics = analytics;
-            _watchedDirectoryPath = watchedDirectoryPath;
+            _watchedJobOrParentDirectoryPath = watchedJobOrParentDirectoryPath;
+            _jobDirectoryPath = jobDirectoryPath;
             _onJobChanged = onJobChanged;
             _filter = filter;
             _listJobNames = listJobNames;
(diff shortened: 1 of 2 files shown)


Kudu.Core/Jobs/JobsManagerBase.cs
```diff
@@ -84,7 +84,19 @@ protected JobsManagerBase(ITraceFactory traceFactory, IEnvironment environment,
 
             JobsBinariesPath = Path.Combine(Environment.JobsBinariesPath, jobsTypePath);
             JobsDataPath = Path.Combine(Environment.JobsDataPath, jobsTypePath);
-            JobsWatcher = new JobsFileWatcher(JobsBinariesPath, OnJobChanged, null, ListJobNames, traceFactory, analytics, jobsTypePath);
+
+            var appDataPath = Path.Combine(Environment.WebRootPath, Constants.AppDataPath);
+
+            if (JobsBinariesPath.StartsWith(appDataPath, StringComparison.OrdinalIgnoreCase))
```
davidebbo: When would it ever not be under App_Data, since jobs are always there? e.g. `wwwroot\jobs\continuous` and `wwwroot\jobs\triggered`. It would be best to avoid two different code paths.

davidebbo:
One mitigation is that we don't do the FileSystemWatcher at all if there are no WebJobs. So any high notification situation is limited to sites that have both WebJobs and App_Data that they actively write to.

suwatch:
@davidebbo  We don't do FileSystemWatcher if there is no jobs folders initially.  Instead, we have a timer to check if jobs folders (wwwroot/app_data/jobs/triggered or wwwroot/app_data/jobs/continuous) exists before starting FileWatcher.   Vice versa, when the folder (or any of its parent folders) is deleted, we get an OnError event and stop watcher accordingly.

@northtyphoon The issue is the race between deleting parent folder and an OnError event stopping the watcher.  If one were to retry deleting the parent folder again, it would succeed.

The concern for the fix is (as @davidebbo pointed out) duplicated watchers on app_data folder.  Both job managers will get noises from the others; though, the code should filter them properly.   In addition, any unknown regression that I can't yet think of.

Given the scenario is limited to clean up (anything else?), I wonder patching the web deploy to retry on deleting failure would be a proportional solution for the issue (less invasive than changing how Kudu listens to file changes).

northtyphoon:
@suwatch For your comment, "The issue is the race between deleting parent folder and an OnError event stopping the watcher." It made me think OnError event handler code should immediately dispose FileSystemWatcher to release the file handler instead of queuing a task.
BTW: Since the change watches App_Data, OnError event is only raised when deleting App_Data folder.

@EricSten-MSFT, what do you think of "patching web deploy to retry on deleting failure"?

Kudu.Core/Jobs/JobsManagerBase.cs
northtyphoon: I didn't know jobs folder should always under App_Data. I will merge the code.

Kudu.Core/Jobs/JobsFileWatcher.cs
```diff
@@ -108,7 +111,7 @@ private void StartWatcher(object state)
                 lock (_lockObject)
                 {
                     // Check if there is a directory we can listen on
-                    if (!FileSystemHelpers.DirectoryExists(_watchedDirectoryPath))
+                    if (!FileSystemHelpers.DirectoryExists(_jobDirectoryPath))
                     {
                         // If not check again in 30 seconds
                         _startFileWatcherTimer.Change(CheckForWatcherTimeout, Timeout.Infinite);
```
EricSten: Unrelated question: the comment says 30 seconds, but the timeout is Timeout.Infinite.  And you're doing the wait while holding _lockObject.  Doesn't this seem dangerous?

suwatch:
> It made me think OnError event handler code should immediately dispose FileSystemWatcher to release the file handler instead of queuing a task.

@northtyphoon Looking at the hostory, the queued task was done [intentionally](https://github.com/projectkudu/kudu/commit/840e8d6bac544787f7a5d42c01be755879db2403#diff-5e9b71e77ff5ace69b61dd5d4187777dR143).   Besides, the event is likely async and we will still have a race.

Kudu.Core/Jobs/JobsFileWatcher.cs
suwatch: The 2nd argument represents [period](https://msdn.microsoft.com/en-us/library/317hx6fa(v=vs.110).aspx) of the timer.    Infinite meant no period (just do one time timer per due time).   Make sense?

EricSten: Ah, so the 30-second timeout is wired into the _startFileWatcherTimer object.  Got it.  Thx!

northtyphoon:
We decided to hold the change and ask users to exclude the profiler web jobs from web deploy clean up task.

Issue #2551, opened 2017-09-07
Deployment to Function Apps should never be treated as anything but function apps


davidebbo:
Today, if host.json is missing, we can end up treating it as something else and do the wrong thing. But Function Apps can only run one thing, so we should tweak the detection accordingly.

PR #2552, opened 2017-09-07
remove checks for host.json when deploying a functionApp

watashiSHUN:
fix https://github.com/projectkudu/kudu/issues/2551
updated the [test repo](https://github.com/KuduApps/FunctionAppWithNodePackages/commit/728c31fd560c541e21978d7ed863e97add8690a7) to match this change (not included in this commit since it's a private only test)


watashiSHUN:
rebased  af2029b

Issue #2547, opened 2017-09-06
NullReferenceException in EnsureRequestIdHandlerAttribute


suwatch:
Encountered while investigating API Plane Health.

```json
{"Message":"An error has occurred.","ExceptionMessage":"Object reference not set to an instance of an object","ExceptionType":"System.NullReferenceException","StackTrace":"
  at Kudu.Services.EnsureRequestIdHandlerAttribute.OnActionExecuted (System.Web.Http.Filters.HttpActionExecutedContext actionExecutedContext) [0x00047] in <e144cc92a33944beb179bd9981d2fa89>:0 \n
  at System.Web.Http.Filters.ActionFilterAttribute.OnActionExecutedAsync (System.Web.Http.Filters.HttpActionExecutedContext actionExecutedContext, System.Threading.CancellationToken cancellationToken) [0x00000] in <f99f496cb0d249c1a945c1fcabce1695>:0 \n
--- End of stack trace from previous location where exception was thrown ---\n
  at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () [0x0000c] in <4dc8ec68b0964e099af86e50301f5f3c>:0 \n
  at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (System.Threading.Tasks.Task task) [0x0003e] in <4dc8ec68b0964e099af86e50301f5f3c>:0 \n
  at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Threading.Tasks.Task task) [0x00028] in <4dc8ec68b0964e099af86e50301f5f3c>:0 \n
```

nickwalkmsft:
@davidebbo shared this same stack trace with me a few days ago, he was able to generate it on a Linux site in a specific scenario; I'll let him share more details. I haven't looked into it in detail yet. Do you happen to know if the offending sites were Linux?

davidebbo:
In my case, I was calling `/api/functions` on Linux site and hit this, but no idea what caused it. @suwatch, I assume you saw this on Windows? Let's get to the bottom of it as it blocked me from making progress on my Functions/Linux experiments.

suwatch:
Just about to add that it is only happening on Linux workers.  Try Kusto below for details.   They are from /siteextensions api.

```
AntaresAdminGeoEvents
| where PreciseTimeStamp >= datetime(2017-09-01)
| where PreciseTimeStamp <= datetime(2017-09-06)
[...]

PR #2556, opened 2017-09-08
Fix for #2547 occasional null HttpContext on Mono

nickwalkmsft:


suwatch:
:shipit:

Kudu.Services/EnsureRequestIdHandlerAttribute.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Threading;
```
suwatch: what is this for?

PR #2568, opened 2017-09-18
Check NuGet.org for SiteExtensions

EricSten-MSFT:


Kudu.Core/SiteExtensions/SiteExtensionManager.cs
```diff
@@ -110,21 +118,35 @@ public SiteExtensionManager(IContinuousJobsManager continuousJobManager, ITrigge
...
             UIPackageMetadata package = null;
 
             if (string.IsNullOrWhiteSpace(version))
             {
                 using (tracer.Step("Version is null, search latest package by id: {0}", id))
                 {
-                    package = await remoteRepo.GetLatestPackageByIdFromSrcRepo(id);
+                    foreach (SourceRepository remoteRepo in remoteRepos)
+                    {
+                        package = await remoteRepo.GetLatestPackageByIdFromSrcRepo(id);
```
davidebbo: Not sure this is quite correct, as this will return any package from nuget.org, even if it doesn't have our `AzureSiteExtension` tag.

Kudu.Core/SiteExtensions/SiteExtensionManager.cs
```diff
@@ -110,21 +118,35 @@ public SiteExtensionManager(IContinuousJobsManager continuousJobManager, ITrigge
...
                 }
             }
             else
             {
                 using (tracer.Step("Search package by id: {0} and version: {1}", id, version))
                 {
-                    package = await remoteRepo.GetPackageByIdentity(id, version);
+                    foreach (SourceRepository remoteRepo in remoteRepos)
+                    {
+                        package = await remoteRepo.GetPackageByIdentity(id, version);
```
davidebbo: Same as previous comment.

Kudu.Core/SiteExtensions/SiteExtensionManager.cs
```diff
@@ -264,23 +287,40 @@ public SiteExtensionManager(IContinuousJobsManager continuousJobManager, ITrigge
                     {
                         using (tracer.Step("Version is null, search latest package by id: {0}, will not search for unlisted package.", id))
                         {
-                            repoPackage = await remoteRepo.GetLatestPackageByIdFromSrcRepo(id);
+                            foreach (SourceRepository rr in remoteRepos)
+                            {
+                                repoPackage = await rr.GetLatestPackageByIdFromSrcRepo(id);
```
davidebbo: Same `AzureSiteExtension` issue

Kudu.Core/SiteExtensions/SiteExtensionManager.cs
```diff
@@ -264,23 +287,40 @@ public SiteExtensionManager(IContinuousJobsManager continuousJobManager, ITrigge
...
                         }
                     }
                     else
                     {
                         using (tracer.Step("Search package by id: {0} and version: {1}, will also search for unlisted package.", id, version))
                         {
-                            repoPackage = await remoteRepo.GetPackageByIdentity(id, version);
+                            foreach (SourceRepository rr in remoteRepos)
+                            {
+                                repoPackage = await rr.GetPackageByIdentity(id, version);
```
davidebbo: Same `AzureSiteExtension` issue

davidebbo:
A good way to see the problem is to try to install the `NewRelic.Azure.WebSites` package. It exists in both gallery, but the one on NuGet is really unrelated, and doesn't have the tag. But we still end up trying to install it.

Kudu.Core/SiteExtensions/SiteExtensionManager.cs
```diff
@@ -75,17 +76,24 @@ public SiteExtensionManager(IContinuousJobsManager continuousJobManager, ITrigge
...
             using (tracer.Step("Search site extensions by filter: {0}", filter))
             {
-                packages = (await remoteRepo.Search(string.IsNullOrWhiteSpace(filter) ? string.Empty : filter, filterOptions: filterOptions))
-                            .OrderByDescending(p => p.DownloadCount);
+                foreach (SourceRepository remoteRepo in remoteRepos)
+                {
+                    var foundUIPackages = await remoteRepo.Search(string.IsNullOrWhiteSpace(filter) ? string.Empty : filter, filterOptions: filterOptions);
+                    if (null != foundUIPackages)
+                    {
+                        packages = packages.Concat(foundUIPackages);
```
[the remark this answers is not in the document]
EricSten-MSFT: I thought the mission was to find the ones on the NuGet.org feed first, and add the ones from siteextensions.org second.  I didn't know there was a requirement to eliminate duplicates.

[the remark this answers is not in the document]
davidebbo: I suppose we could tell people that once they transition to nuget.org, they should hide all the packages from siteextensions.net.

Kudu.Core/SiteExtensions/SiteExtensionManager.cs
```diff
@@ -75,17 +76,24 @@ public SiteExtensionManager(IContinuousJobsManager continuousJobManager, ITrigge
...
+                    }
+                }
+                packages.OrderByDescending(p => p.DownloadCount);
```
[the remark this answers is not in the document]
EricSten-MSFT: Originally I had:
```packages.Concat(await remoteRepo.Search(string.IsNullOrWhiteSpace(filter) ? string.Empty : filter, filterOptions: filterOptions));```
but it turns out that .Concat doesn't actually concat to packages; it simply returns the concatination of packages + 1st arg.  This was the cause of the last [bug fix](https://github.com/projectkudu/kudu/pull/2568/commits/0891d0e63b546d6f4061568a9ac4c01673c61100).
From the description of [foo.Concat(bar) ](https://msdn.microsoft.com/en-us/library/bb302894(v=vs.110).aspx), I foolishly thought that it would concatenate bar to foo in-place. I was wrong; foo is not modified by .Concat().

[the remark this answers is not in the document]
Comment (davidebbo) [reply Kudu.Core/SiteExtensions/SiteExtensionManager.cs]: Noe of the Linq methods ever have side effects. It's all about building new enumerations over existing ones.

File: Kudu.Core/SiteExtensions/SiteExtensionManager.cs (hunk 1 above)
Comment (EricSten-MSFT) [reply Kudu.Core/SiteExtensions/SiteExtensionManager.cs]: Ah, you're right.  Only [.Search()](https://github.com/projectkudu/kudu/blob/150dab5ff20fe525e7eb64f9629a64664adb7cb3/Kudu.Core/SiteExtensions/FeedExtensions.cs#L39-L59) has the special logic to automagically add the AzureSiteExtension magic tag.  I'll have to figure out the best place to put this magic to avoid duplicating code everywhere.

Comment (EricSten-MSFT) [reply Kudu.Core/SiteExtensions/SiteExtensionManager.cs]: Side Question: is "AzureSiteExtension" case sensitive?  And do we need to worry about collisions?
 e.g. "DefinitelyNotAnAzureSiteExtension" vs. "AzureSiteExtension"

Comment (davidebbo) [reply Kudu.Core/SiteExtensions/SiteExtensionManager.cs]: Not sure about the casing, but I highly doubt that it is sensitive (you'll have to try it on NuGet.org with the one package we have there).

No, I'm not worried about those collisions. The tag is unique enough that it won't happen accidentally.

File: Kudu.Core/SiteExtensions/FeedExtensions.cs [hunk 7]
```diff
@@ -52,9 +52,9 @@ public static class FeedExtensions
                 }
                 else
                 {
-                    // User provided simple string: treat it as a title. This is not ideal behavior, but
+                    // User provided simple string: treat it as a id. This is not ideal behavior, but
                     // is the best we can do based on how nuget.org works
-                    searchTerm = $"tags:AzureSiteExtension title:\"{searchTerm}\"";
+                    searchTerm = $"tags:AzureSiteExtension id:\"{searchTerm}\"";
```
EricSten-MSFT: I'm switching this to ```id:```, since the functional tests pass a package Id instead of a title in places.  If this is too much of a breaking change, I can change the functional tests to pass package Titles instead.

EricSten-MSFT:
@davidebbo Re: ```NewRelic.Azure.WebSites```, I added a functional test per your comment, above.  See: ```SiteExtensionNugetOrgTest()``` in [SiteExtensionApiFacts.cs](https://github.com/projectkudu/kudu/pull/2568/files#diff-e3c07df96e60c5667429a37746621975)

Kudu.Core/SiteExtensions/FeedExtensions.cs
davidebbo: @EricSten-MSFT you can assume that no one is using the nuget.org branch of the current code, so don't worry about breaking changes

Kudu.Core/SiteExtensions/SiteExtensionManager.cs
```diff
@@ -801,7 +856,18 @@ private void CleanupSiteExtensionJobs(string siteExtensionName)
 
         private async Task<SiteExtensionInfo> CheckRemotePackageLatestVersion(SiteExtensionInfo info, UIMetadataResource metadataResource)
         {
-            UIPackageMetadata localPackage = await metadataResource.GetLatestPackageByIdFromMetaRes(info.Id);
+            bool isNuGetPackage = false;
+
+            if (!string.IsNullOrEmpty(info.FeedUrl))
+            {
+                isNuGetPackage = System.Text.RegularExpressions.Regex.IsMatch(
```
davidebbo: Why not use the helper that does that check to avoid duplication?

davidebbo:
Actually, I think I can get GitHub to do the Squash & Merge.

Kudu.Core/SiteExtensions/SiteExtensionManager.cs
EricSten-MSFT: Because I was too lazy to refactor IsNuGetRepo to take a string instead of a SourceRepository.  I'll fix it.

Kudu.Core/SiteExtensions/SiteExtensionManager.cs
```diff
@@ -571,23 +610,30 @@ private static void EnsureInstallationEnviroment(string installationDir, ITracer
...
             {
-                // case 3
-                UIPackageMetadata remotePackage = await remoteRepo.GetLatestPackageByIdFromSrcRepo(id);
-                if (remotePackage != null)
+                // case 1 and 2
+                if (!string.IsNullOrWhiteSpace(version)
+                    && version.Equals(localPackageVersion, StringComparison.OrdinalIgnoreCase)
+                    && rr.PackageSource.Source.Equals(localPackageFeedUrl, StringComparison.OrdinalIgnoreCase))
+                {
+                    isInstalled = true;
```
davidebbo: This code is actually incorrect when the version is blank, as it prevents upgrading from the first feed to the second feed if you already have the latest from the first (but second has newer one).

davidebbo: Actually, maybe the issue is really in `GetLatestPackageByIdFromMetaRes`. When everything in the old feed is unlisted, it shouldn't return anything. But it's still returning the latest. Not entirely sure why.

EricSten: @davidebbo : Fixed.  Submitted a new [pull request](https://github.com/projectkudu/kudu/pull/2568) with the fix.

Kudu.Core/SiteExtensions/FeedExtensions.cs
davidebbo: @EricSten-MSFT as I'm trying to finally make this transition happen, this change from `title` to `id` seems problematic, because it cause search to only work for exac match. e.g. compare:

- Search by 'counter' successfully finds filecounter package when using `title:counter`: https://www.nuget.org/packages?q=tags%3AAzureSiteExtension+title%3Acounter
- The same doesn't find it when using `id:counter`: https://www.nuget.org/packages?q=tags%3AAzureSiteExtension+id%3Acounter

If the only issue here was a functional test, I think we should revert back to title. Do you know what test was failing?

davidebbo: Ok, I moved back to title and updated the tests. 59d028358398e636611ad46e9323846abde1f535

PR #2577, opened 2017-09-30
Return 409 on functions secret api if the storageType is set to blob.

### LINKED ISSUE #2515: Functions key API needs to return an error if slots is enabled
See https://github.com/Azure/azure-webjobs-sdk-script/issues/1752 for context.

If `AzureWebJobsSecretStorageType=Blob` is set, kudu returns invalid keys. Until its capable of returning the right keys, it should throw an error.

ahmelsayed:
Closes #2515

/cc @paulbatum


davidebbo:
@ahmelsayed is this still waiting for CR?

PR #2578, opened 2017-09-30
Reduce size of test_data for ARM requests to 8MB/2

### LINKED ISSUE #2576: Fix logic to reduce test data size on /api/functions/<name> endpoint
The payload for this API can include a test data file provided by the user that might be large. There is logic to limit this size to stay under the ARM 8mb limit but it doesn't work right, we are still seeing this in ARM logs:

```
System.Net.Http.HttpRequestException: Cannot write more bytes to the buffer than the configured maximum buffer size: 8388608.
   at System.Net.Http.HttpContent.LimitMemoryStream.CheckSize(Int32 countToAdd)
   at System.Net.Http.HttpContent.LimitMemoryStream.Write(Byte[] buffer, Int32 offset, Int32 count)
   at System.Net.Http.StreamToStreamCopy.TryStartWriteSync(Int32 bytesRead)
   at System.Net.Http.StreamToStreamCopy.StartRead()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
   at Microsoft.WindowsAzure.ResourceStack.Common.EventSources.CommonEventSourceExtensions.<TraceHttpOutgoingRequest>d__37.MoveNext() in E:\bt\845424\repo\src\common\instrumentation\EventSources\CommonEventSourceExtensions.cs:line 1098
```

Probably because other data in the payload is still pushing us over. Perhaps just halve the limit to make sure we get under without issue.

ahmelsayed:
Closes #2576

/cc @paulbatum

### DIFF
--- a/Kudu.Contracts/Functions/FunctionTestData.cs
+++ b/Kudu.Contracts/Functions/FunctionTestData.cs
@@ -2,9 +2,9 @@
 {
     public class FunctionTestData
     {
-        // test shows test_data of size 8310000 bytes still delivers as an ARM package
-        // whereas test_data of size 8388608 bytes fails
-        public const long PackageMaxSizeInBytes = 8300000;
+        // ARM has a limit of 8 MB -> 8388608 bytes
+        // divid by 2 to limit the over all size of test data to half of arm requirement to be safe.
+        public const long PackageMaxSizeInBytes = 8388608 / 2;
 
         public long BytesLeftInPackage { get; set; } = PackageMaxSizeInBytes;
 


davidebbo:
@ahmelsayed is this change in?

PR #2584, opened 2017-10-05
Fix build

jincod:
use `System.Web.Http` from nuget package

### DIFF
--- a/Kudu.Core/Kudu.Core.csproj
+++ b/Kudu.Core/Kudu.Core.csproj
@@ -194,7 +194,9 @@
       <Private>True</Private>
     </Reference>
     <Reference Include="System.Web.Extensions" />
-    <Reference Include="System.Web.Http" />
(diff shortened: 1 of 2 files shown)


jincod:
Hi @davidebbo 

I can't reproduce this issue again (broken build). But I have many warnings:
```
C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\Microsoft.Common.CurrentV
ersion.targets(1987,5): warning MSB3245: Could not resolve this reference. Could not locate the assembly 
"System.Web.Http". Check to make sure the assembly exists on disk. If this reference is required by your 
code, you may get compilation errors. [..\kudu\Kudu.Core\Kudu.Core.csproj]
```

Is it ok?

davidebbo:
Turns out this was a bogus reference, and I just yanked it altogether. No more warnings! Thanks for reporting.

PR #2585, opened 2017-10-05
Fix site extension version check across multiple feeds

EricSten-MSFT:
Tested that leaving an empty version will allow finding the most recent version of a site extension across multiple feeds.


davidebbo:
Looks like you're still based on the old tree, which is why it shows all the old commit. You need to rebase your work on the latest tree to reduce changes of conflict.

Kudu.Core/SiteExtensions/SiteExtensionManager.cs
```diff
@@ -622,19 +625,28 @@ private static void EnsureInstallationEnviroment(string installationDir, ITracer
...
                 {
                     // case 3
                     UIPackageMetadata remotePackage = await rr.GetLatestPackageByIdFromSrcRepo(id);
                     if (remotePackage != null)
                     {
-                        isInstalled = remotePackage.Identity.Version.ToNormalizedString().Equals(localPackageVersion, StringComparison.OrdinalIgnoreCase) && string.Equals(localPackageInstallationArgs, installationArgs);
+                        tracer.Trace("Found version: {0} on feed: {1}",
+                                     remotePackage.Identity.Version.ToNormalizedString(),
+                                     rr.PackageSource.Source);
+                        if (lastFoundVer < remotePackage.Identity.Version)
```
davidebbo: Is this check ok if `lastFoundVer` is null?

Kudu.Services/Kudu.Services.csproj
```diff
@@ -11,11 +11,6 @@
     <DocumentationFile>bin\Debug\Kudu.Services.XML</DocumentationFile>
     <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
     <TargetFrameworkProfile />
-  </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
-    <NoWarn>1591</NoWarn>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Release|AnyCPU'">
```
davidebbo: That looks like a random diff that probably should not be included.

Kudu.SiteManagement/Kudu.SiteManagement.csproj
```diff
@@ -10,10 +10,10 @@
     <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
     <TargetFrameworkProfile />
   </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+  <PropertyGroup Condition=" '$(Configuration)' == 'Debug' ">
     <DefineConstants>TRACE;DEBUG;CODE_ANALYSIS;SITEMANAGEMENT</DefineConstants>
   </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+  <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
```
davidebbo: This file should not be part of it either.

Kudu.Services/Kudu.Services.csproj
EricSten-MSFT: Uh, how do I pull the two files out of the change list?

davidebbo: Several ways, but one is to just push another commit that restores them. It all gets squashed in the end anyway.

PR #2588, opened 2017-10-06
streamline csproj files

EricSten-MSFT:
The $(Platform) is redundant.  Plus, the difference between $(Platform) for 'AnyCPU' vs. 'Any CPU' causes problems when building using msbuild.exe vs. Visual Studio.


davidebbo:
Merged.

PR #2591, opened 2017-10-11
Fix offset calculation issue in logstream with non-English characters

davidebbo:


Kudu.Services/Diagnostics/LogStreamManager.cs
```diff
@@ -406,7 +406,8 @@ private IEnumerable<string> GetChanges(FileSystemEventArgs e)
                             changes.Add(line);
                         }
 
-                        offset += line.Length;
+                        byte[] bytes = reader.CurrentEncoding.GetBytes(line);
+                        offset += bytes.Length;
```
suwatch: Thank for a quick one!   A alternative (simpler) fix is https://github.com/projectkudu/kudu/commit/0fe7fbe12ce4786aaefa233da6b2c4272578b5f4.  No need to worry about encoding or any calculation.   I did a couple of verification and they looks good.

davidebbo: Ah yes, that's much simpler! Let me test it a bit.

davidebbo: Ok, seems to work great. I'll close this PR.

Issue #2546, opened 2017-09-05
Drag-and-drop upload to Debug Console doesn't work from Edge


ggailey777:
1. In an **_Edge browser_**, open the Debug Console for a Function App, in this case: <https://glengastorageevents.scm.azurewebsites.net/DebugConsole>
2. Drag a site extension .zip file to the top of the console and drop it in the **Drag here to upload and unzip area**:
    ![Drag here to upload and unzip area](https://user-images.githubusercontent.com/1388263/30051798-2dc4e318-91d8-11e7-8048-2728a946accd.png)
**Expected results:** Knife spins and file upload % is shown, which completes with 100%. This is how it works on IE.
**Actual results:** Knife spins and nothing else happens. Percent is not shown and upload never succeeds.

davidebbo:
Indeed, I reproed this. Not sure if it ever worked.

nukec:
doesn't work for chrome nor firefox

PR #2593, opened 2017-10-11
nmallick Pull Request

nmallick1:
Suggesting fix for https://github.com/projectkudu/kudu/issues/2546
Edge is throwing the following error as dataTransfer property is null.

SCRIPT5007: Unable to get property 'items' of undefined or null reference
filebrowser.js (802,9)

The OR statement results in an error and the function terminates, hence the file upload fails. Checking for null and assuming zip to allow the function to continue. 

Tested the code change in Edge.


davidebbo:
@nmallick1 will re-one a new one for this.

PR #2597, opened 2017-10-13
Fix for bug #2546. 

nmallick1:
Checking to see if dataTransfer is null before calling a method on it. The dataTransfer object is null for IE and Edge hence the issue.


davidebbo:
Tried it and it works great!

Only issue, which I think was there before, is that sometimes the zip area image doesn't go away after it's done. But that's separate.

Thanks for the contrib!

PR #2604, opened 2017-10-24
Fix TrackPendingOperation

EricSten-MSFT:


Kudu.Core/Helpers/PostDeploymentHelper.cs
```diff
@@ -433,7 +433,7 @@ public static void RunPostDeploymentScripts(TraceListener tracer)
 
                 var cancelation = new CancellationTokenSource();
                 var delay = Task.Delay(DefaultUpdateMarkerIntervalMS, cancelation.Token);
-                var completed = await Task.WhenAny(delay, task);
+                var completed = await Task.WhenAny( new Task[] { task, delay } );
```
davidebbo: Isn't that equivalent? What's the rationale behind the change?

EricSten-MSFT: It's not, apparently.  WhenAny has overrides, and the (delay, task) was matching one of the overrides of (Task[], Int32).  Forcing to an explicit array fixes things.

davidebbo: I don't understand how that's possible, since that override takes an int and we are not passing any. We also use that same overload that takes 2 tasks in other places.

Maybe something else is at play and that is not really the issue you're after?

EricSten-MSFT: Yes, you're right.  Please reject this PR.  I'm debugging it, and it only works under the debugger.

PR #2616, opened 2017-11-09
Add support for sending triggered webjobs logs to App Logs

davidebbo:
Some big customers are asking for it, and it has come up before. Making it optional to make sure it doesn't surprise any existing customers.


suwatch:
:shipit:

PR #2638, opened 2017-12-01
Function Apps do not scale if there are only proxies in the function …

safihamid:
Function Apps do not scale if there are only proxies in the function app.

This is a recent regression when we moved away from SiteExtension. 

The issue described here: https://github.com/Azure/azure-webjobs-sdk-script/issues/2177


suwatch:
Look good.  I will merge it.

PR #2640, opened 2017-12-03
Adding UI for Zip Deploy and a link to this UI under Tools menu

nmallick1:
Adding a UI for Zip Deploy and a link to this UI under Tools menu.


davidebbo:
Ok, so two things:

- I found that your change does work for me on Azure, but doesn't work locally for some reason. So the issue was not related to the merge. Anyway, we can live with this for now since this is experimental
- You had done a merge and not a rebase. I did a rebase + squash instead, and pushed the change. So it'll be in the next deployment!

davidebbo:
@nmallick1 your new UI is now available in Production. Thanks for contributing! :)

@nickwalkmsft give it a try!

nickwalkmsft:
Love it, it's great to see this feature come to life in the browser.

This won't be lit up in Linux apps for a while, as we don't have the advanced console and visual file navigator working yet, but it'll get there.

Thanks for this contribution @nmallick1!

Issue #2630, opened 2017-11-21
Deployment status after synchronous zipdeploy


nikolaia:
In regards to #2562 and https://github.com/projectkudu/kudu/wiki/Deploying-from-a-zip-file

Loving the Zipdeploy feature as I can write my deployment scripts using FAKE (`.deployment -> deploy.cmd -> deploy.fsx` - all present in the .zip artifact) and can write a smart pipeline for what needs to be done for my WebApps (since kudusync works offline I can test the scripts 100% on my own machine, so awesome!):
```
"KuduStageFiles"
    ==> "MigrateDatabase"
    ==> "Deploy"
    ==> "SmokeTests"
```

However I'm having some trouble getting VSTS to fail when the deployment fails in Kudu (I'm running the call synchronous from an Azure CLI step in VSTS).

This script can be run from my own machine, and is in source controll with the deployment script (Double awesome). It is the script called by VSTS:

```powershell
Param(
    [Parameter(Mandatory=$True,Position=1)]
    [string]$appName
  )

$token = (ConvertFrom-Json -InputObject ([string](Invoke-Expression -Command:"az account get-access-token")))."accessToken"
$bearerToken = "Bearer $token"
$apiUrl = "https://$appName.scm.azurewebsites.net/api/zipdeploy"

$artifactDir = split-path -parent $MyInvocation.MyCommand.Definition
$zip = Get-ChildItem -Path $artifactDir -Filter *.zip | Select-Object -First 1
$zipPath = "$artifactDir/$($zip.Name)"

Write-Host "Calling $apiUrl with $zipPath using token '$bearerToken'"
Invoke-RestMethod -Uri $apiUrl -Headers @{Authorization=$bearerToken} -Method POST -InFile $zipPath -ContentType "multipart/form-data"
```

Seems like I'm just getting 200 OK back from the zipdeploy endpoint, even if the deployment script failed and the deployment is not active. Do I need to poll to check the status of my deployment? Would be nice to get in the body if the deployment is active or not after running synchronous.

davidebbo:
I wonder if this is the same root cause as #2272, which relates to error propagation in PowerShell scripts. You can try the workaround given by the users there to see if that works here.

nikolaia:
[...]

PR #2641, opened 2017-12-04
Give better feedback on synchronous zipdeploy

nikolaia:
As mentioned in #2630 it would be nice to know if the deployment succeeded or failed when running a synchronous zipdeploy

### DIFF
--- a/Kudu.Core/Deployment/DeploymentManager.cs
+++ b/Kudu.Core/Deployment/DeploymentManager.cs
@@ -151,7 +151,7 @@ namespace Kudu.Core.Deployment
             }
         }
 
-        public async Task DeployAsync(
+        public async Task<DeployResult> DeployAsync(
             IRepository repository,
             ChangeSet changeSet,
             string deployer,
@@ -275,6 +275,8 @@ namespace Kudu.Core.Deployment
                 {
                     throw new DeploymentFailedException(exception);
                 }
+
+                return GetResult(id);
             }
         }
 
--- a/Kudu.Core/Deployment/FetchDeploymentManager.cs
+++ b/Kudu.Core/Deployment/FetchDeploymentManager.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Threading.Tasks;
 using Kudu.Contracts;
 using Kudu.Contracts.Infrastructure;
@@ -72,7 +73,7 @@ namespace Kudu.Core.Deployment
             // push/clone endpoint and zip deployment.
             if (!(_settings.IsScmEnabled() || deployInfo.AllowDeploymentWhileScmDisabled))
             {
-                return FetchDeploymentRequestResult.ForbiddenScmDisabled;
+                return new FetchDeploymentRequestResult(FetchDeploymentRequestResultStatus.ForbiddenScmDisabled);
             }
 
             // for CI payload, we will return Accepted and do the task in the BG
@@ -93,8 +94,8 @@ namespace Kudu.Core.Deployment
                         waitForTempDeploymentCreation);
 
                     return successfullyRequested
-                    ? FetchDeploymentRequestResult.RunningAynschronously
-                    : FetchDeploymentRequestResult.ConflictDeploymentInProgress;
+                    ? new FetchDeploymentRequestResult(FetchDeploymentRequestResultStatus.RunningAynschronously)
+                    : new FetchDeploymentRequestResult(FetchDeploymentRequestResultStatus.ConflictDeploymentInProgress);
                 }
             }
 
@@ -105,11 +106,13 @@ namespace Kudu.Core.Deployment
                 {
                     if (PostDeploymentHelper.IsAutoSwapOngoing())
                     {
-                        return FetchDeploymentRequestResult.ConflictAutoSwapOngoing;
+                        return new FetchDeploymentRequestResult(FetchDeploymentRequestResultStatus.ConflictAutoSwapOngoing);
                     }
 
-                    await PerformDeployment(deployInfo);
-                    return FetchDeploymentRequestResult.RanSynchronously;
+                    var result = await PerformDeployment(deployInfo);
+                    return result != null && result.Status != DeployStatus.Failed 
+                        ? new FetchDeploymentRequestResult(FetchDeploymentRequestResultStatus.RanSynchronously)
+                        : new FetchDeploymentRequestResult(FetchDeploymentRequestResultStatus.RanSynchronouslyFailed, result?.StatusText);
                 }, "Performing continuous deployment", TimeSpan.Zero);
             }
             catch (LockOperationException)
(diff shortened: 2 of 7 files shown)


davidebbo:
@nickwalkmsft can you review this change to zip push deploy?

Kudu.Services/Deployment/PushDeploymentController.cs
```diff
@@ -109,6 +109,11 @@ public class PushDeploymentController : ApiController
                         break;
                     case FetchDeploymentRequestResult.RanSynchronously:
                         response.StatusCode = HttpStatusCode.OK;
+                        response.Content = new StringContent("Succeeded");
+                        break;
+                    case FetchDeploymentRequestResult.RanSynchronouslyFailed:
+                        response.StatusCode = HttpStatusCode.OK;
```
nickwalkmsft: Seems like this should be BadRequest or something similar?

nickwalkmsft: Along with error text?

Kudu.Services/ServiceHookHandlers/FetchHandler.cs
```diff
@@ -118,6 +118,7 @@ public class FetchHandler : HttpTaskAsyncHandler
                         context.ApplicationInstance.CompleteRequest();
                         break;
                     case FetchDeploymentRequestResult.RanSynchronously:
+                    case FetchDeploymentRequestResult.RanSynchronouslyFailed:
```
nickwalkmsft: This is actually the reason I left the feature in question out of Zipdeploy - because synchronous fetch deployment didn't do it. I didn't just want to add it and change the expected behavior, and it made more sense (to me, at the time) to keep the behavior the same between fetch and zip/push, so I did. @davidebbo what's your take on this?

davidebbo: So is the /api/zip behavior today that if something goes wrong on the server, the request doesn't fail? If so, that doesn't seem good, and I wouldn't make push zip behave the same way (ideally we'd fix both).

Kudu.Services/Deployment/PushDeploymentController.cs
nikolaia: I changed `FetchDeploymentRequestResult` to include a StatusText and renamed the enum. Don't know if you approve of the pattern, let me know. Changed to BadRequest, agree it's better.

Kudu.Services/ServiceHookHandlers/FetchHandler.cs
nikolaia: I changed fetch to act the same way as push on RanSynchronously, but I'm not 100% sure of the consequences here so feedback is appreciated. Tests seems to run fine at least 💯

nickwalkmsft: Not /api/zip, but /deploy (fetch deployment), which zipdeploy has the most in common with (a zip deployment just pushes the zip to the server and then does a "fetch" deployment with a custom fetch action that unzips the local file and sets the deployment source to the extract directory).

In DeploymentManager.Build(), the call to builder.Build() (which does most of the actual work of the deployment) is wrapped in a try/catch that doesn't rethrow. See https://github.com/projectkudu/kudu/blob/master/Kudu.Core/Deployment/DeploymentManager.cs#L651-L675. Since nothing further up the call stack checks the status, you get a 200 back. It's always been this way.

davidebbo: Ah I see. I guess the reason is that in those scenarios, the http call normally comes from GitHub/BitBucket, and the error is less relevant. Instead, the deployment status shows it. But for Zip, I can see how not having an error becomes more quirky.

nickwalkmsft:
Okey doke - overall I think this is a good idea and would like to merge it. Could you make a couple of revisions? I am adding a couple comments now.

Thanks for the contrib!

Kudu.Services/ServiceHookHandlers/FetchHandler.cs
```diff
@@ -97,27 +97,35 @@ public class FetchHandler : HttpTaskAsyncHandler
...
                         context.ApplicationInstance.CompleteRequest();
                         return;
-                    case FetchDeploymentRequestResult.ConflictDeploymentInProgress:
+                    case FetchDeploymentRequestResultStatus.ConflictDeploymentInProgress:
                         context.Response.StatusCode = (int)HttpStatusCode.Conflict;
                         context.Response.Write(Resources.Error_DeploymentInProgress);
                         context.ApplicationInstance.CompleteRequest();
                         break;
-                    case FetchDeploymentRequestResult.RanSynchronously:
+                    case FetchDeploymentRequestResultStatus.RanSynchronously:
```
nickwalkmsft: This new behavior is good for PushDeploymentController since it's new and likely to be used interactively, but here for FetchHandler, please put this bit back how it was (have RanSynchronously fall through to the default case), and have RanSynchronouslyFailed fall through to the default case as well. I'm hesitant to change the behavior here, since this is generally called by webhooks from many different CI services and I don't want to break expectations related to how they react to the response. A short comment to acknowledge the RanSynchronouslyFailed case and explaining that we're avoiding a behavior change would be good also.

Kudu.Services/Deployment/PushDeploymentController.cs
```diff
@@ -94,23 +94,27 @@ public class PushDeploymentController : ApiController
...
                         // Shouldn't happen here, as we disallow deferral for this use case
                         response.StatusCode = HttpStatusCode.Accepted;
                         break;
-                    case FetchDeploymentRequestResult.RanSynchronously:
+                    case FetchDeploymentRequestResultStatus.RanSynchronously:
                         response.StatusCode = HttpStatusCode.OK;
                         break;
-                    case FetchDeploymentRequestResult.ConflictDeploymentInProgress:
+                    case FetchDeploymentRequestResultStatus.RanSynchronouslyFailed:
+                        response.StatusCode = HttpStatusCode.BadRequest;
```
nickwalkmsft: I misspoke on this earlier; it should probably be a 500 instead of a 400.

Kudu.Core/Deployment/FetchDeploymentManager.cs
```diff
@@ -105,11 +106,14 @@ public class FetchDeploymentManager : IFetchDeploymentManager
                 {
                     if (PostDeploymentHelper.IsAutoSwapOngoing())
                     {
-                        return FetchDeploymentRequestResult.ConflictAutoSwapOngoing;
+                        return new FetchDeploymentRequestResult(FetchDeploymentRequestResultStatus.ConflictAutoSwapOngoing);
                     }
 
                     await PerformDeployment(deployInfo);
-                    return FetchDeploymentRequestResult.RanSynchronously;
+                    var result = _deploymentManager.GetResults().FirstOrDefault();
```
nickwalkmsft: GetResults() is not a good idea here (it hits the filesystem to enumerate all the existing results, does locking stuff, and has the side effect of purging old deployment records, none of which I want to incur here) but unfortunately there's no other option the way things are factored now.

DeploymentManager.DeployAsync() currently doesn't return anything. Given the functionality it would make sense to return the DeployResult, and in turn that could be returned up through PerformDeployment() and checked here. Please give that a go; if it's not looking great let me know and I'll take a swing at it or something similar.

Kudu.Core/Deployment/DeploymentManager.cs
```diff
@@ -274,6 +274,8 @@ public void Delete(string id)
                 {
                     throw new DeploymentFailedException(exception);
                 }
+
+                return GetResult(id);
```
nikolaia: Is this safe?

Kudu.Services/Deployment/PushDeploymentController.cs
nikolaia: Fixed

Kudu.Core/Deployment/FetchDeploymentManager.cs
nikolaia: Now using `GetResult(id)` to return the DeployResult from `DeployAsync`? Does that solve the lockingissues?

Kudu.Services/ServiceHookHandlers/FetchHandler.cs
nikolaia: Reverted to default case

Kudu.Core/Deployment/DeploymentManager.cs
nickwalkmsft: Sorry for the delay, I'll get to this soon...

nikolaia: Bump! Don't wanna nag, but thought I'd remind you about this in case it's been forgotten ⏳

nickwalkmsft: Thanks @nikolaia, have been working on other priorities but will get to this as soon as I can, it's still on my radar.

lizter: Any updates here? :)

honorew:
when isAsync=true,could return the deploy id ?

nikolaia:
I'm gonna close this. Ended up just using the Azure CLI to deploy. (Example: https://github.com/nikolaia/weather-in-harstad/blob/master/provision.fsx#L82)

PR #2667, opened 2018-01-18
fix triggered webjob not started during r/o filesystem

suwatch:


Kudu.Core/Jobs/TriggeredJobRunLogger.cs
```diff
@@ -29,7 +29,7 @@ private TriggeredJobRunLogger(string jobName, string id, IEnvironment environmen
             _id = id;
 
             _historyPath = Path.Combine(Environment.JobsDataPath, Constants.TriggeredPath, jobName, _id);
-            FileSystemHelpers.EnsureDirectory(_historyPath);
+            OperationManager.SafeExecute(() => FileSystemHelpers.EnsureDirectory(_historyPath));
```
mathewc: There's one more call to FileSystemHelpers in this file - should that one also be fixed?

PR #2674, opened 2018-01-24
Support WAR file based deployments

shrishrirang:
This commit introduces a new API (`/api/wardeploy`) for publishing WAR files to Java Web Apps to address the deployment issues with scaled out Java apps.

### Background
Publishing a WAR file to a Java web app is a common scenario. The WAR file is published to the wwwroot/webapps directory. Tomcat monitors the wwwroot/webapps directory for the presence of new *.war files (or changes to existing ones). The moment it detects either of that, it unzips the war file into a new directory that matches the name of the war file and loads the unzipped web application from that directory (WAR files are just zip files with a war extension).

In a scaled out web app, multiple Tomcat instances (one on each worker) attempt to unzip the WAR file causing multiple issues. Here are some common problems: When a change to the WAR file is detected, multiple Tomcat instances attempt to lock and unzip it, causing locking problems. Another issues is Tomcat attempting to reload a webapp which is in the middle of being updated by another Tomcat instance resulting in the Tomcat instance going into a bad state. There are few more race conditions like these which eventually result in Tomcat responding to incoming requests with 404.

### wardeploy
This commit builds on top of the existing zipdeploy mechanism to perform unzipping of WAR files outside the wwwwroot directory and deploys the expanded WAR file to the wwwwroot/webapps/ROOT directory which is monitored by Tomcat. After the deployment is complete wardeploy touches the wwwroot/webapps/ROOT/WEB-INF/web.xml file which causes Tomcat to reload the web app (equivalent to touching web.config). The race conditions with scaled out Java web apps are avoided as WAR file is not directly published to the wwwroot/webapps directory.

### DIFF
--- a/Kudu.Services/Deployment/PushDeploymentController.cs
+++ b/Kudu.Services/Deployment/PushDeploymentController.cs
@@ -19,8 +19,8 @@ namespace Kudu.Services.Deployment
 {
     public class PushDeploymentController : ApiController
     {
-        private const string DefaultDeployer = "Zip-Push";
-        private const string DefaultMessage = "Created via zip push deployment";
+        private const string DefaultDeployer = "Push-Deployer";
+        private const string DefaultMessage = "Created via a push deployment";
 
         private readonly IEnvironment _environment;
         private readonly IFetchDeploymentManager _deploymentManager;
@@ -50,17 +50,6 @@ namespace Kudu.Services.Deployment
         {
             using (_tracer.Step("ZipPushDeploy"))
             {
-                var zipFileName = Path.ChangeExtension(Path.GetRandomFileName(), "zip");
-                var zipFilePath = Path.Combine(_environment.ZipTempPath, zipFileName);
-
-                using (_tracer.Step("Writing zip file to {0}", zipFilePath))
-                {
-                    using (var file = FileSystemHelpers.CreateFile(zipFilePath))
-                    {
-                        await request.Content.CopyToAsync(file);
-                    }
-                }
-
                 var deploymentInfo = new ZipDeploymentInfo(_environment, _traceFactory)
                 {
                     AllowDeploymentWhileScmDisabled = true,
@@ -68,7 +57,6 @@ namespace Kudu.Services.Deployment
                     IsContinuous = false,
                     AllowDeferredDeployment = false,
                     IsReusable = false,
-                    RepositoryUrl = zipFilePath,
                     TargetChangeset = DeploymentManager.CreateTemporaryChangeSet(message: "Deploying from pushed zip file"),
                     CommitId = null,
                     RepositoryType = RepositoryType.None,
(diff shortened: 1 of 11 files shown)


shrishrirang:
@davidebbo @jvano @nickwalkmsft for CR. 

Btw, I can see ~80 failing functional tests without any of my changes. Is this a known issue or something wrong with my setup? I have followed the instructions from https://github.com/projectkudu/kudu/wiki/Getting-started

davidebbo:
It's not expected in the sense that it all passes on the CI, but setting up everything correctly is pretty difficult due to the many dependencies. Don't worry too much about it if you have high confidence that it's not related to your change. The CI is there to catch things if something does break.

For this change, since you're building on top of zip push depoy, @nickwalkmsft is the best person to talk to and review (he added that a few months ago).

Kudu.Services/Deployment/PushDeploymentController.cs
```diff
@@ -50,26 +50,48 @@ public class PushDeploymentController : ApiController
...
+            [FromUri] string deployer = DefaultDeployer,
+            [FromUri] string message = DefaultMessage)
+        {
+            using (_tracer.Step("WarPushDeploy"))
+            {
                 var deploymentInfo = new ZipDeploymentInfo(_environment, _traceFactory)
                 {
                     AllowDeploymentWhileScmDisabled = true,
                     Deployer = deployer,
+                    TargetPath = @"webapps\ROOT",
```
nickwalkmsft: Cross-plat: Use Path.Combine, or Path.DirectorySeparatorChar if you want to concat strings yourself

Kudu.Services/Deployment/PushDeploymentController.cs
```diff
@@ -50,26 +50,48 @@ public class PushDeploymentController : ApiController
...
+            HttpRequestMessage request,
+            [FromUri] bool isAsync = false,
+            [FromUri] string author = null,
+            [FromUri] string authorEmail = null,
+            [FromUri] string deployer = DefaultDeployer,
+            [FromUri] string message = DefaultMessage)
+        {
+            using (_tracer.Step("WarPushDeploy"))
+            {
                 var deploymentInfo = new ZipDeploymentInfo(_environment, _traceFactory)
```
nickwalkmsft: Did you consider DoFullBuildByDefault = false? Do WAR file contents potentially have anything in them that could trigger undesired deployment logic intended for another app stack?

shrishrirang: WAR files are expected to contain a fully built payload that simply needs to be unzipped without any additional (build) steps, so `DoFullBuildByDefault = false` looks like a correct choice here. 

Regarding the second question - theoretically, the WAR file can contain a `.deploy` file that could trigger an unintended action. We can address this possibility later, if needed, but it is a corner case that should be ok to ignore at this point.

Kudu.Services/Deployment/PushDeploymentController.cs
shrishrirang: Good point, will update the PR!

shrishrirang:
Thanks for the review @nickwalkmsft! Do you recommend any specific manual validation? Does the CI have coverage for Linux scenarios?

@suwatch, let me know if you have any comments.

PR #2685, opened 2018-02-02
App Settings using IIS AppPool environmentVariables

duckblaster:
Not sure if SiteManager is the right place to put this.

### DIFF
--- a/Kudu.SiteManagement/SiteManager.cs
+++ b/Kudu.SiteManagement/SiteManager.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Collections.Specialized;
 using System.Configuration;
 using System.Diagnostics;
 using System.IO;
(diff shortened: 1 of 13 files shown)


davidebbo:
@duckblaster sorry, I just noticed this PR. I sthis still up to date and worth reviewing? If so, we will get to it soon!

duckblaster:
I haven't had a chance to fully test it, but it compiles fine and I don't see any reason it wouldn't work.
It might be worth adding checks so it doesn't run on older versions of IIS, but the try catch blocks handle that.

duckblaster:
It seems to be working fine as far as I can tell

Kudu.SiteManagement/SiteManager.cs
```diff
@@ -600,5 +601,66 @@ private ServerManager GetServerManager()
...
+            try
+            {
+                using (var iis = GetServerManager())
+                {
+                    var appPool = iis.ApplicationPools[GetAppPool(applicationName)];
+                    var collection = appPool.GetCollection("environmentVariables");
+                    var addElement = collection.CreateElement("add");
+                    addElement["name"] = key;
+                    addElement["value"] = value;
+                    collection.Add(addElement);
```
davidebbo: This blows up if you try to change the value of an existing variable. e.g. it throws:

    Cannot add duplicate collection entry of type 'add' with unique key attribute 'name' set to 'FOO'

Simplest fix is probably to just call `RemoveAppSetting(applicationName, key);` as the first line of the `using`.

Kudu.SiteManagement/SiteManager.cs
```diff
@@ -600,5 +601,66 @@ private ServerManager GetServerManager()
...
+                    iis.CommitChanges();
+                }
+            }
+            catch { }
```
davidebbo: It's best not to swallow all exceptions like this. e.g. this made it harder to find the 'duplicate' issue above. Best to just let the code blow up when that happens. Same comment for other places that swallow all exceptions.

Issue #2709, opened 2018-02-28
Continuous deployment via GitHub fails on Nuget package restore, however succeeds when published via VS 2017


ahmelsayed:
_From @MrMikeFloyd on February 27, 2018 12:3_

Hey Functions Team,

I'm experiencing issues deploying Azure Functions using the GitHub Webhook. Publishing my Function to the portal from VisualStudio 2017 works fine. I followed [this guideline from the docs](https://docs.microsoft.com/en-us/azure/azure-functions/functions-continuous-deployment).

## Steps undertaken

1. Created a function in the portal
2. Downloaded its app content (via "**Download App content**" -> "**Content and Visual Studio Content**")
3. Opened in VS, made adjustments to the code
4. Added to a Git Repo, committed/pushed to branch for development
5. Added GitHub Repo/branch to functions app via "**Deployment options**" as denoted [in the docs](https://docs.microsoft.com/en-us/azure/azure-functions/functions-continuous-deployment)
6. Hit "**sync**"

## Symptoms

Deployment ends up in the "failed" state after a short while:

![deployment_details](https://user-images.githubusercontent.com/14937417/36726882-8aa4ad26-1bbb-11e8-8436-2ce0e6772136.png)

The log file for generating the deployment script shows the following:

```
Using cached version of deployment script (command: 'azure -y --no-dot-deployment -r "D:\home\site\repository" -o "D:\home\site\deployments\tools" --functionApp "D:\home\site\repository\wwwroot\functionnamexyz-dev.csproj" --no-solution').
```

The deployment command log shows the following:

```
Command: "D:\home\site\deployments\tools\deploy.cmd"
Handling function App deployment with Msbuild.
Using Msbuild from 'D:\Program Files (x86)\MSBuild-15.3.409.57025\MSBuild\15.0\Bin'.
Input file does not exist: D:\home\site\repository\undefined.
Failed exitCode=1, command=nuget.exe restore "D:\home\site\repository\undefined" -MSBuildPath "D:\Program Files (x86)\MSBuild-15.3.409.57025\MSBuild\15.0\Bin"
An error has occurred during web site deployment.
Input file does not exist: D:\home\site\repository\undefined.\r\nD:\Program Files (x86)\SiteExtensions\Kudu\71.10217.3259\bin\Scripts\starter.cmd "D:\home\site\deployments\tools\deploy.cmd"
```

[...]

PR #2710, opened 2018-02-28
update csproj for DownloadFunctions API

watashiSHUN:
image
removing unnecessary references since they are included in the new sdk
adding `TODOs` for furture changes, this commit is for fixing https://github.com/projectkudu/kudu/issues/2709

### DIFF
--- a/Kudu.Core/Functions/FunctionManager.cs
+++ b/Kudu.Core/Functions/FunctionManager.cs
@@ -476,9 +476,9 @@ namespace Kudu.Core.Functions
         /// <param name="projectName">the {projectName}.csproj</param>
         private static ZipArchiveEntry AddCsprojFile(ZipArchive zip, IEnumerable<ZipArchiveEntry> files, string projectName)
         {
-            const string microsoftAzureWebJobsVersion = "2.1.0-beta1";
-            const string microsoftAzureWebJobsExtensionsHttpVersion = "1.0.0-beta1";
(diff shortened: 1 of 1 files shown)


Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -476,9 +476,9 @@ public void CreateArchive(ZipArchive zip, bool includeAppSettings = false, bool
         /// <param name="projectName">the {projectName}.csproj</param>
         private static ZipArchiveEntry AddCsprojFile(ZipArchive zip, IEnumerable<ZipArchiveEntry> files, string projectName)
         {
-            const string microsoftAzureWebJobsVersion = "2.1.0-beta1";
-            const string microsoftAzureWebJobsExtensionsHttpVersion = "1.0.0-beta1";
-            const string microsoftNETSdkFunctionsVersion = "1.0.0-alpha6";
+            // TODO different template for functionv2
+            // TODO download git repository when read only
```
ahmelsayed: what does this TODO mean?

watashiSHUN: when user has selected a deployment options (ie, `git push` a csharp function created in VS)
this functionality will return `bin/` and append a `csproj` instead of returning an edible project ( resides under `site\repository`)

ahmelsayed: Got it. That makes sense

PR #2718, opened 2018-03-14
wardeploy improvements to support custom app names

shrishrirang:
- This change allows deploying war files to web apps other than ROOT.
- Posting war file contents to /api/wardeploy?name=myappname will create the myappname webapp by deploying the war contents to wwwroot/webapps/<appname> directory. If no app name is specified, ROOT is inferred.

### DIFF
--- a/Kudu.Core/Deployment/DeploymentManager.cs
+++ b/Kudu.Core/Deployment/DeploymentManager.cs
@@ -624,6 +624,7 @@ namespace Kudu.Core.Deployment
                 {
                     NextManifestFilePath = GetDeploymentManifestPath(id),
                     PreviousManifestFilePath = GetActiveDeploymentManifestPath(),
+                    IgnoreManifest = deploymentInfo.CleanupTargetDirectory,
                     Tracer = tracer,
                     Logger = logger,
                     GlobalLogger = _globalLogger,
--- a/Kudu.Core/Deployment/DeploymentManager.cs
+++ b/Kudu.Core/Deployment/DeploymentManager.cs
@@ -624,7 +624,11 @@ namespace Kudu.Core.Deployment
                 {
                     NextManifestFilePath = GetDeploymentManifestPath(id),
                     PreviousManifestFilePath = GetActiveDeploymentManifestPath(),
-                    IgnoreManifest = deploymentInfo.CleanupTargetDirectory,
+                    IgnoreManifest = deploymentInfo.CleanupTargetDirectory, // Ignoring the manifest will cause kudusync to delete sub-directories / files
+                                                                            // in the destination directory that are not present in the source directory,
+                                                                            // without checking the manifest to see if the file was copied over to the destination
+                                                                            // during a previous kudusync operation. This effectively performs a clean deployment
+                                                                            // from the source to the destination directory.
                     Tracer = tracer,
                     Logger = logger,
                     GlobalLogger = _globalLogger,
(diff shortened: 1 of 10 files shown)


shrishrirang:
@davidebbo again as Nick is oof. This PR uses https://github.com/projectkudu/KuduSync/pull/34 and https://github.com/projectkudu/KuduScript/pull/94 to perform clean deployment by ignoring the manifest when performing wardeploy. It does this by setting `CleanupTargetDirectory = true`.

Kudu.Core/Deployment/DeploymentManager.cs
```diff
@@ -624,6 +624,7 @@ private DeployResult GetResult(string id, string activeDeploymentId, bool isDepl
                 {
                     NextManifestFilePath = GetDeploymentManifestPath(id),
                     PreviousManifestFilePath = GetActiveDeploymentManifestPath(),
+                    IgnoreManifest = deploymentInfo.CleanupTargetDirectory,
```
davidebbo: It's not instantly clear to me how ignoring the manifest results in cleaning up the target folder. A comment would be useful here.

shrishrirang: Done.

davidebbo: Thanks, I guess I had misunderstood/forgotten when `IgnoreManifest` meant, but now I understand.

shrishrirang:
Thanks for the review!

davidebbo:
@shrishrirang darn, it looks like this completely breaks Kudu git deployments on Windows. I haven't looked at why yet.

Kudu.Core/Deployment/DeploymentManager.cs
```diff
@@ -624,6 +624,11 @@ private DeployResult GetResult(string id, string activeDeploymentId, bool isDepl
                 {
                     NextManifestFilePath = GetDeploymentManifestPath(id),
                     PreviousManifestFilePath = GetActiveDeploymentManifestPath(),
+                    IgnoreManifest = deploymentInfo.CleanupTargetDirectory, // Ignoring the manifest will cause kudusync to delete sub-directories / files
```
davidebbo: Problem is that `deploymentInfo` can be null, which crashes here.

shrishrirang:
Aah! I'll share a PR for the fix.

davidebbo:
I just pushed a fix to just check for null. 79c21de2a24c8425bbb00c1103a17f2ddbbc0a37.

shrishrirang:
Thanks! :)

davidebbo:
Now the only test that failed is TestSimpleWarDeploymentPerformsCleanDeployment. Is that supposed to run at all on Windows?

shrishrirang:
Yes, it is supposed to run successfully. That tests tests the wardeploy performs a clean deployment, which is one of the changes introduced by this PR. It did work fine on my machine. Let me take a look.

davidebbo:
Seems the file-before-wardeployment doesn't get deleted.

shrishrirang:
Ok, let me check why.

davidebbo:
Doesn't it need the new kuduscript that you sent a PR for? https://github.com/projectkudu/KuduScript/pull/94

shrishrirang:
Oh yes, that is precisely the reason!
In my mind everything was supposed to work fine because the `IGNORE_MANIFEST` environment variable would have no effect until kuduscript changes are in. Didn't realize I already added a test and that would fail. Fix for this is to just comment the test for now. Once I update `updateNodeModules.cmd`, it needs to be enabled again. Should I push that change or you have it already ready to push? :)

davidebbo:
It's ok if this one test fails for a day or two.

So in what scenario exactly will IgnoreManifest be set? Only the WAR scenario on Linux? Because if it's more general, then technically all the generated scripts should support it, as they all call kudusync.

PR #2720, opened 2018-03-21
[WIP] Support Run-From-Zip in ZipDeploy

ahmelsayed:
This is still a work in progress, but it works. If you would like to try it here is a private site extension build
[Kudu.zip](https://github.com/projectkudu/kudu/files/1834130/Kudu.zip)

To use run from zip:

1. Set `WEBSITE_USE_ZIP = 1`
2. `curl -X POST -u <user> --data-binary @<zipfile> https://{sitename}.scm.azurewebsites.net/api/zipdeploy`

Remaining work:
- [ ] Add unit tests
- [ ] Add an e2e deployment test
- [x] Run some more manual validations
- [x] handle some error cases

### DIFF
--- a/Kudu.Contracts/Deployment/DeploymentInfoBase.cs
+++ b/Kudu.Contracts/Deployment/DeploymentInfoBase.cs
@@ -61,5 +61,12 @@ namespace Kudu.Core.Deployment
         }
 
         public abstract IRepository GetRepository();
+
+        // If this is not set, sync triggers will look under d:\home\site\wwwroot
+        // for functionsRoot. Otherwise it'll use this path for that
+        // This is used in Run-From-Zip deployments where the content of wwwroot
+        // won't update until after a process restart. Therefore, we copy the needed
+        // files into a separate folders and run sync triggers from there.
+        public string SyncFunctionsTriggersPath { get; set; } = null;
     }
 }
--- a/Kudu.Contracts/Settings/DeploymentSettingsExtension.cs
+++ b/Kudu.Contracts/Settings/DeploymentSettingsExtension.cs
@@ -249,5 +249,17 @@ namespace Kudu.Contracts.Settings
             // returning true by default here as an indicator of generally expected behavior
             return value == null || StringUtils.IsTrueLike(value);
         }
+
+        public static bool RunFromLocalZip(this IDeploymentSettingsManager settings)
+        {
+            return settings.GetValue(SettingsKeys.RunFromZip) == "1";
+        }
+
+        public static bool RunFromRemoteZip(this IDeploymentSettingsManager settings)
+        {
+            var value = settings.GetValue(SettingsKeys.RunFromZip);
+
+            return value != null && value.StartsWith("http", StringComparison.OrdinalIgnoreCase);
+        }
     }
 }
--- a/Kudu.Services/Deployment/PushDeploymentController.cs
+++ b/Kudu.Services/Deployment/PushDeploymentController.cs
@@ -1,19 +1,20 @@
 ﻿using System;
+using System.Globalization;
+using System.IO;
 using System.IO.Compression;
+using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Threading.Tasks;
 using System.Web.Http;
+using Kudu.Contracts.Settings;
 using Kudu.Contracts.Tracing;
+using Kudu.Core;
 using Kudu.Core.Deployment;
 using Kudu.Core.Infrastructure;
+using Kudu.Core.SourceControl;
 using Kudu.Core.Tracing;
 using Kudu.Services.Infrastructure;
-using System.IO;
-using Kudu.Core.SourceControl;
-using System.Globalization;
-using Kudu.Core;
-using System.Linq;
 
 namespace Kudu.Services.Deployment
 {
@@ -26,17 +27,20 @@ namespace Kudu.Services.Deployment
         private readonly IFetchDeploymentManager _deploymentManager;
         private readonly ITracer _tracer;
         private readonly ITraceFactory _traceFactory;
+        private readonly IDeploymentSettingsManager _settings;
 
         public PushDeploymentController(
             IEnvironment environment,
             IFetchDeploymentManager deploymentManager,
             ITracer tracer,
-            ITraceFactory traceFactory)
+            ITraceFactory traceFactory,
+            IDeploymentSettingsManager settings)
         {
             _environment = environment;
             _deploymentManager = deploymentManager;
             _tracer = tracer;
             _traceFactory = traceFactory;
+            _settings = settings;
         }
 
         [HttpPost]
@@ -59,11 +63,18 @@ namespace Kudu.Services.Deployment
                     TargetChangeset = DeploymentManager.CreateTemporaryChangeSet(message: "Deploying from pushed zip file"),
                     CommitId = null,
                     RepositoryType = RepositoryType.None,
-                    Fetch = LocalZipFetch,
+                    Fetch = LocalZipHandler,
                     DoFullBuildByDefault = false,
                     Author = author,
                     AuthorEmail = authorEmail,
-                    Message = message
+                    Message = message,
+                    // This is used if the deployment is Run-From-Zip
+                    // the name of the deployed file in D:\home\data\SitePackages\{name}.zip is the 
+                    // timestamp in the format yyyMMddHHmmss. 
+                    ZipName = $"{DateTime.UtcNow.ToString("yyyyMMddHHmmss")}.zip",
+                    // This is also for Run-From-Zip where we need to extract the triggers
+                    // for post deployment sync triggers.
+                    SyncFunctionsTriggersPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName())
                 };
 
                 return await PushDeployAsync(deploymentInfo, isAsync);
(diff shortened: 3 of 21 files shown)


Kudu.Contracts/Settings/DeploymentSettingsExtension.cs
```diff
@@ -249,5 +249,10 @@ public static bool DoBuildDuringDeployment(this IDeploymentSettingsManager setti
             // returning true by default here as an indicator of generally expected behavior
             return value == null || StringUtils.IsTrueLike(value);
         }
+
+        public static bool RunFromLocalZip(this IDeploymentSettingsManager settings)
+        {
+            return settings.GetValue(SettingsKeys.RunFromZip) == "1";
```
ahmelsayed: We should probably make the deployment fail if this is set to a URL?

Kudu.Services/Deployment/PushDeploymentController.cs
```diff
@@ -222,6 +237,50 @@ public class PushDeploymentController : ApiController
...
 
+        private async Task LocalZipFetchAndDeploy(IRepository repository, DeploymentInfoBase deploymentInfo, ILogger logger, ITracer tracer)
+        {
+            var zipDeploymentInfo = (ZipDeploymentInfo)deploymentInfo;
+
+            // For this kind of deployment, RepositoryUrl is a local path to the temp zip.
+            var sourceZipFile = zipDeploymentInfo.RepositoryUrl;
+            // the name of the deployed file in D:\home\data\SitePackages\{name}.zip is the 
+            // timestamp in the format yyyMMddHHmmss. 
+            var fileName = $"{DateTime.UtcNow.ToString("yyyyMMddHHmmss")}.zip";
```
ahmelsayed: as the comment is saying. the name of the deployed zip is a timestamp. Any other preference?

Kudu.Contracts/Settings/DeploymentSettingsExtension.cs
davidebbo: Agreed, that would not make sense.

Kudu.Services/Deployment/PushDeploymentController.cs
```diff
@@ -222,6 +237,50 @@ public class PushDeploymentController : ApiController
...
+                filePath,
+                sizeInMb);
+
+            logger.Log(message);
+
+            using (tracer.Step(message))
+            {
+                // If extractTargetDirectory already exists, rename it so we can delete it concurrently with
+                // the unzip (along with any other junk in the folder)
+                var cleanTask = Task.Run(() => DeleteFilesAndDirsExcept(sourceZipFile, null, tracer));
```
ahmelsayed: tbh, I just copied this from @nickwalkmsft's existing logic in the other `LocalZipFetch` method. I'm not sure why we need to explicitly clean up temp here. 

The code flow in this PR is a bit odd though. From the original design from Nick, this class `PushDeploymentController ` handles saving the zip into a temp file under `d:\local` then the Deployment manager calls a delegate defined in this class that does the deployment (unzip).

This change plugs into the same existing code where this class will save the zip under a temp folder, then the deployment manager will call `LocalZipFetchAndDeploy` which just copies it to the right place. 

I'm thinking of making this controller write the file in the right place and have the deployment manager no-op the deployment, but not sure if that impacts anything else. Mainly thinking of deployment logs that come from the deployment manager object. I'll test it to see what that does

Kudu.Services/Deployment/PushDeploymentController.cs
```diff
@@ -222,6 +237,50 @@ public class PushDeploymentController : ApiController
...
+                var extractTask = Task.Run(() =>
+                {
+                    FileSystemHelpers.CopyFile(sourceZipFile, filePath, overwrite: true);
```
davidebbo: Ideally, we would save it directly there without it ever being in temp. But I guess that would make the logic more complex?

Interestingly, at runtime, this zip will end up getting copied back somewhere into d:\local.

Let's make sure that the `sourceZipFile` gets deleted here at the end of deployment.

Kudu.Services/Deployment/PushDeploymentController.cs
davidebbo: That's a bit aggressive, but I guess that's ok to get started. We could get fancier later and keep the last 'n' zips, or some other cleanup policy.

Kudu.Services/Deployment/PushDeploymentController.cs
```diff
@@ -222,6 +237,50 @@ public class PushDeploymentController : ApiController
...
+                    FileSystemHelpers.WriteAllTextToFile(Path.Combine(_environment.SitePackagesPath, Constants.SiteVersionTxt), fileName);
+                });
+
+                await Task.WhenAll(cleanTask, extractTask);
+            }
+
+            // Needed in order for repository.GetChangeSet() to work.
+            // Similar to what OneDriveHelper and DropBoxHelper do.
+            repository.Commit(zipDeploymentInfo.Message, zipDeploymentInfo.Author, zipDeploymentInfo.AuthorEmail);
```
davidebbo: I don't remember how that all works. I thought in typical zipdeploy scenarios there was no repo at all. What is this committing to?

Kudu.Services/Deployment/PushDeploymentController.cs
davidebbo: I think that's fine.

Kudu.Services/Deployment/PushDeploymentController.cs
davidebbo: Right, that would seem a bit better, unless it means adding way more complexity. It's not terrible to have it in TEMP during transition as long as it gets cleaned up.

Kudu.Services/Deployment/PushDeploymentController.cs
ahmelsayed: You're right. It's not needed here. In regular ZipDeploy while there is no repo, apparently we're creating a NullRepository with 1 commit as above because the rest of the code expects a ChangeSet() for various logs and checks and it was probably easier to add this instead of wiring up all the deployment code to handle no repo scenarios. 

However because we will skip the actual deploy step for run-from-zip, we don't need this here

ahmelsayed:
Okay I updated the PR with the feedback. The file is written directly into `D:\home\data\SitePackages\{timestamp}.zip`

Files needed for sync triggers are extracted from the zip into `D:\local\temp\{randomDir}\` and sync trigger is executed from there. 
The extraction operation is very fast. On my machine it took ~90 msecs to extract 10 functions from a zip archive containing 120k files (a 160 MB zip). It also took 900 msec to load the archive from disk. On Azure Files it took 20 seconds to load the archive. 

Finally `siteversion.txt` is updated with the deployment zip name.

Kudu.Contracts/Deployment/DeploymentInfoBase.cs
```diff
@@ -61,5 +61,12 @@ public bool IsValid()
         }
 
         public abstract IRepository GetRepository();
+
+        // If this is not set, sync triggers will look under d:\home\site\wwwroot
+        // for functionsRoot. Otherwise it'll use this path for that
+        // This is used in Run-From-Zip deployments where the contnent of wwwroot
+        // won't update until after a process restart. Therefore, we copy the needed
+        // files into a seperate folders and run sync triggeers from there.
+        public string SyncFunctionsTrigersPath { get; set; } = null;
```
davidebbo: Property spelling: Triger -> Trigger

Kudu.Contracts/Deployment/DeploymentInfoBase.cs
```diff
@@ -61,5 +61,12 @@ public bool IsValid()
         }
 
         public abstract IRepository GetRepository();
+
+        // If this is not set, sync triggers will look under d:\home\site\wwwroot
+        // for functionsRoot. Otherwise it'll use this path for that
+        // This is used in Run-From-Zip deployments where the contnent of wwwroot
```
davidebbo: Typo: 'content'

Kudu.Contracts/Deployment/DeploymentInfoBase.cs
```diff
@@ -61,5 +61,12 @@ public bool IsValid()
...
+        // won't update until after a process restart. Therefore, we copy the needed
+        // files into a seperate folders and run sync triggeers from there.
```
davidebbo: Typos: 'separate' & 'triggers'

Kudu.Services/Deployment/PushDeploymentController.cs
```diff
@@ -171,6 +193,72 @@ public class PushDeploymentController : ApiController
...
+            // Tested Azure Files share with a zip containing 120k files (160 MBs)
+            // takes 20 seconds to load. On my machine it takes 900 msec.
+            using (var zip = ZipFile.OpenRead(Path.Combine(_environment.SitePackagesPath, zipDeploymentInfo.ZipName)))
+            {
+                var entries = zip.Entries
+                    // Only select host.json, proxies.json, or top level directories.
+                    // Tested with a zip containing 120k files, and this took 90 msec
+                    // on my machine.
+                    .Where(e => e.FullName.Equals(Constants.FunctionsHostConfigFile, StringComparison.OrdinalIgnoreCase) ||
+                                e.FullName.Equals(Constants.ProxyConfigFile, StringComparison.OrdinalIgnoreCase) ||
```
davidebbo: Shouldn't there be a test for host.json as well?

Kudu.Contracts/Deployment/DeploymentInfoBase.cs
ahmelsayed: reminded me I need to install a spell checker :)

Kudu.Services/Deployment/PushDeploymentController.cs
ahmelsayed: as mentioned, the line above is the one for `host.json` (`Constants.FunctionsHostConfigFile == "host.json"`)

PR #2722, opened 2018-03-22
allow case insensitivity for durableTask related propertyies.  

suwatch:
See https://github.com/Azure/azure-functions-durable-extension/issues/111


suwatch:
include @cgillum @tohling from durable task.

PR #2725, opened 2018-03-25
Make BaseJobRunner get the path from the Job object

davidebbo:
Currently it computes it itself from Environment, breaking encapsulation. This is a small step to be able to have multiple folders.

### DIFF
--- a/Kudu.Core/Jobs/BaseJobRunner.cs
+++ b/Kudu.Core/Jobs/BaseJobRunner.cs
@@ -29,18 +29,18 @@ namespace Kudu.Core.Jobs
         private string _inPlaceWorkingDirectory;
         private Dictionary<string, FileInfoBase> _cachedSourceDirectoryFileMap;
 
-        protected BaseJobRunner(string jobName, string jobsTypePath, IEnvironment environment,
+        protected BaseJobRunner(JobBase job, string jobsTypePath, IEnvironment environment,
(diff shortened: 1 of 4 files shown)


ahmelsayed:
:shipit:

Kudu.Core/Jobs/BaseJobRunner.cs
```diff
@@ -29,18 +29,18 @@ public abstract class BaseJobRunner
...
             Environment = environment;
             Settings = settings;
-            JobName = jobName;
+            JobName = job.Name;
             _analytics = analytics;
 
-            JobBinariesPath = Path.Combine(Environment.JobsBinariesPath, jobsTypePath, jobName);
-            JobTempPath = Path.Combine(Environment.TempPath, Constants.JobsPath, jobsTypePath, jobName);
-            JobDataPath = Path.Combine(Environment.DataPath, Constants.JobsPath, jobsTypePath, jobName);
+            JobBinariesPath = job.JobBinariesRootPath;
```
davidebbo: @ahmelsayed turns out this broken things (caught by `CreateAndDeleteJobSucceeds` test). The reason is that you can have the following structure:

    App_Data
      jobs
        triggered
          job1
            inner
              run.cmd

In this case, `job.JobBinariesRootPath` points to `.../job1/inner`. But the old code was pointing to `.../job1`. This then causes things to blow up.

I think it may be best to just revert my change and rethink in the context of your bigger 'dual WebJobs folder' change. Let's discuss.

PR #2728, opened 2018-03-27
Cron DST fix

mathewc:
See https://github.com/Azure/azure-webjobs-sdk-extensions/pull/395 for details.


Kudu.Core.Test/Jobs/ScheduleFacts.cs
```diff
@@ -16,7 +16,7 @@ public class ScheduleFacts
         [InlineData("* * * 3 * *", "00:00:00 1/1/2015", "2.00:00:00", false)]
         [InlineData("* * * * 2 *", "00:00:00 1/1/2015", "31.00:00:00", false)]
         [InlineData("* * * * 3 *", "00:00:00 1/1/2015", "59.00:00:00", false)]
-        [InlineData("* * * * 4 *", "00:00:00 1/1/2015", "90.00:00:00", false)]
+        [InlineData("* * * * 4 *", "00:00:00 1/1/2015", "89.23:00:00", false)]  // DST transition on March 11th
```
mathewc: This existing test was actually incorrect. In this case it's configured to run every 4th month, so from January 1st to April 1st is normally 90 days, but March 11th is the switch to DST so the interval should be 23 not 24 hours.

PR #2732, opened 2018-03-29
Support webjobs in secondary location site/jobs

ahmelsayed:
This PR adds 2 `ITriggeredJobsManager` and 2 `IContinuousJobsManager` objects. 

Each manager is responsible for 1 directory of webjobs

`PrimaryManager => D:\home\site\wwwroot\App_Data\jobs`
`SecondaryManager => D:\home\site\jobs`

`JobsController`:

- List operation return a combined view
- Write operations (Create, Update, Delete)
if app is Run-From-Zip, use secondary, otherwise use primary
- Get operations look in both managers. Primary takes precedence.

### DIFF
--- a/Kudu.Core/Jobs/AggregrateJobsManagerBase.cs
+++ b/Kudu.Core/Jobs/AggregrateJobsManagerBase.cs
@@ -1,0 +1,110 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Kudu.Contracts.Jobs;
+using Kudu.Contracts.Settings;
+
+namespace Kudu.Core.Jobs
+{
+    public abstract class AggregrateJobsManagerBase<TJob> where TJob : JobBase, new()
+    {
+        protected JobsManagerBase<TJob> PrimaryJobManager { get; private set; }
+        protected JobsManagerBase<TJob> SecondaryJobManager { get; private set; }
+        private readonly IDeploymentSettingsManager _settings;
+
+        protected AggregrateJobsManagerBase(JobsManagerBase<TJob> primaryManager, Func<IEnumerable<string>, JobsManagerBase<TJob>> secondaryManagerFactory, IDeploymentSettingsManager settings)
+        {
+            PrimaryJobManager = primaryManager;
+            // pass the list of primary job names so the second manager can excluded them
+            SecondaryJobManager = secondaryManagerFactory(PrimaryJobManager.ListJobs(forceRefreshCache: false).Select(j => j.Name));
+            _settings = settings;
+        }
+
+        public void DeleteJob(string jobName)
+        {
+            if (_settings.RunFromZip())
+            {
+                SecondaryJobManager.DeleteJob(jobName);
+            }
+            else
+            {
+                // Make sure to delete the job from both managers if it exists.
+                // Calling DeleteJob on a non-existing job is a no-op.
+                PrimaryJobManager.DeleteJob(jobName);
+                SecondaryJobManager.DeleteJob(jobName);
+            }
+        }
+
+        public void CleanupDeletedJobs()
+        {
+            PrimaryJobManager.CleanupDeletedJobs();
+            SecondaryJobManager.CleanupDeletedJobs();
+        }
+
+        public TJob CreateOrReplaceJobFromFileStream(Stream scriptFileStream, string jobName, string scriptFileName)
+            => GetWriteJobManagerForJob(jobName).CreateOrReplaceJobFromFileStream(scriptFileStream, jobName, scriptFileName);
+
+        public TJob CreateOrReplaceJobFromZipStream(Stream zipStream, string jobName)
+            => GetWriteJobManagerForJob(jobName).CreateOrReplaceJobFromZipStream(zipStream, jobName);
+
+        public TJob GetJob(string jobName) 
+            => PrimaryJobManager.HasJob(jobName) ? PrimaryJobManager.GetJob(jobName) : SecondaryJobManager.GetJob(jobName);
+
+        public JobSettings GetJobSettings(string jobName)
+            => PrimaryJobManager.HasJob(jobName) ? PrimaryJobManager.GetJobSettings(jobName) : SecondaryJobManager.GetJobSettings(jobName);
+
+        public bool HasJob(string jobName)
+            => PrimaryJobManager.HasJob(jobName) || SecondaryJobManager.HasJob(jobName);
+
+        public IEnumerable<TJob> ListJobs(bool forceRefreshCache)
+            => PrimaryJobManager.ListJobs(forceRefreshCache).Concat(SecondaryJobManager.ListJobs(forceRefreshCache));
+
+        public void RegisterExtraEventHandlerForFileChange(Action<string> action)
+        {
+            PrimaryJobManager.RegisterExtraEventHandlerForFileChange(action);
+            SecondaryJobManager.RegisterExtraEventHandlerForFileChange(action);
+        }
+
+        public void SetJobSettings(string jobName, JobSettings jobSettings)
+            => GetWriteJobManagerForJob(jobName).SetJobSettings(jobName, jobSettings);
+
+        // Always sync webjobs brought in by site extensions
+        // into the writable location.
+        public void SyncExternalJobs(string sourcePath, string sourceName)
+            => WritableJobManager.SyncExternalJobs(sourcePath, sourceName);
+
+        // Always sync webjobs brought in by site extensions
+        // into the writable location.
+        public void CleanupExternalJobs(string sourceName)
+            => WritableJobManager.CleanupExternalJobs(sourceName);
+
+        // Writable manager is secondary if run from zip, and primary otherwise.
+        private JobsManagerBase<TJob> WritableJobManager => _settings.RunFromZip()
+            ? SecondaryJobManager
+            : PrimaryJobManager;
+
+        // This checks both run from zip, and Primary.HasJob()
+        // The point is that if we are in run from zip, we always use the secondary.
+        // Otherwise, we use the manager where the job exists. This can be either the primary or the secondary.
+        protected JobsManagerBase<TJob> GetWriteJobManagerForJob(string jobName)
+        {
+            if (_settings.RunFromZip())
+            {
+                return SecondaryJobManager;
+            }
+            else if (PrimaryJobManager.HasJob(jobName))
+            {
+                return PrimaryJobManager;
+            }
+            else if (SecondaryJobManager.HasJob(jobName))
+            {
+                return SecondaryJobManager;
+            }
+            else
+            {
+                return PrimaryJobManager;
+            }
+        }
+    }
+}
--- a/Kudu.Core/Jobs/BaseJobRunner.cs
+++ b/Kudu.Core/Jobs/BaseJobRunner.cs
@@ -29,7 +29,7 @@ namespace Kudu.Core.Jobs
         private string _inPlaceWorkingDirectory;
         private Dictionary<string, FileInfoBase> _cachedSourceDirectoryFileMap;
 
-        protected BaseJobRunner(string jobName, string jobsTypePath, IEnvironment environment,
+        protected BaseJobRunner(string jobName, string jobsTypePath, string basePath, IEnvironment environment,
             IDeploymentSettingsManager settings, ITraceFactory traceFactory, IAnalytics analytics)
         {
             TraceFactory = traceFactory;
@@ -38,7 +38,8 @@ namespace Kudu.Core.Jobs
             JobName = jobName;
             _analytics = analytics;
 
-            JobBinariesPath = Path.Combine(Environment.JobsBinariesPath, jobsTypePath, jobName);
+            JobBinariesPath = Path.Combine(basePath, jobName);
+
             JobTempPath = Path.Combine(Environment.TempPath, Constants.JobsPath, jobsTypePath, jobName);
             JobDataPath = Path.Combine(Environment.DataPath, Constants.JobsPath, jobsTypePath, jobName);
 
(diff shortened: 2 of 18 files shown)


Kudu.Services/Jobs/JobsController.cs
```diff
@@ -68,7 +74,7 @@ public HttpResponseMessage EnableContinuousJob(string jobName)
         {
             try
             {
-                _continuousJobsManager.EnableJob(jobName);
+                _continuousJobsManagers.GetWriteManager(_settings).EnableJob(jobName);
```
ahmelsayed: the code in this controller is error prone, but each API needs to get a manager depending on the API scenario. List APIs combine both, write APIs need to call `GetWriteManager`, other APIs call `GetManagerForJob()` which will look into the primary manager first, then the secondary

Kudu.Core/Extensions/JobManagerBaseIReadOnlyCollectionExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+                ? colleciton.GetSecondaryManager() as IContinuousJobsManager
+                : colleciton.GetPrimaryManager() as IContinuousJobsManager;
+        }
+
+        public static ITriggeredJobsManager GetWriteManager(this IReadOnlyCollection<ITriggeredJobsManager> colleciton, IDeploymentSettingsManager settings)
+        {
+            return settings.RunFromLocalZip() || settings.RunFromRemoteZip()
+                ? colleciton.GetSecondaryManager() as ITriggeredJobsManager
+                : colleciton.GetPrimaryManager() as ITriggeredJobsManager;
+        }
```
ahmelsayed: C# generics are not cooperating with me :) I'm not sure what the right syntax for this should be. The problem is that 

```cs
interface ITriggeredJobsManager : IJobsManager<TriggeredJob> { }
interface IContinuousJobsManager : IJobsManager<ContinuousJob> { }
```

Conceptually what I want is 

```cs
public static T GetWriteManager<T>(this IReadOnlyCollection<T> colleciton) where T : ITriggeredJobsManager, IContinuousJobsManager 
```

but because the interfaces inherit from an interface that takes a concrete type the definition won't compile because `m.BasePathType` is ambiguous between `IJobsManager<TriggeredJob>` and ` IJobsManager<ContinuousJob>`

davidebbo:
Playing with it a bit, the list job API is not doing the right thing because `JobsManagerBase.JobListCache` is static. So both managers end up hitting the same cache.

ahmelsayed:
oh I didn't notice that was a static. I was looking at it and wasn't sure why it's not updating.

ahmelsayed:
I can either make it not static, and have a cache per object (which seems like the right way of doing it) However, many random places in the code seem to be calling clearCache() so I'll have to wire that up too.

Alternatively, we can have a static cache per `WebJobsBasePathType`.

davidebbo:
First is cleaner, but second is probably less disruptive, as it keeps ClearCache static.

Kudu.Core/Extensions/JobManagerBaseIReadOnlyCollectionExtensions.cs
davidebbo: Feels like something that should be solvable, but I was not able to figure this out after playing with it for a bit.

Kudu.Core/Extensions/JobManagerBaseIReadOnlyCollectionExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public static IContinuousJobsManager GetWriteManager(this IReadOnlyCollection<IContinuousJobsManager> colleciton, IDeploymentSettingsManager settings)
+        {
+            return settings.RunFromLocalZip() || settings.RunFromRemoteZip()
+                ? colleciton.GetSecondaryManager() as IContinuousJobsManager
+                : colleciton.GetPrimaryManager() as IContinuousJobsManager;
+        }
+
+        public static ITriggeredJobsManager GetWriteManager(this IReadOnlyCollection<ITriggeredJobsManager> colleciton, IDeploymentSettingsManager settings)
```
davidebbo: You have `colleciton` typo in a bunch of places.

ahmelsayed:
yes I agree. bf8af84812492ac3b1e19ab6af5625e7c52d83d5 adds a cache per `WebJobsBasePathType`

Kudu.Core/Extensions/JobManagerBaseIReadOnlyCollectionExtensions.cs
ahmelsayed: I know. I tried a bunch of stuff too and couldn't get it to work. I spent long enough on it then just gave up

```cs
public static T GetWriteManager<T, U>(this IReadOnlyCollection<IJobsManager<U>> c) 
    where U : JobBase, new()
```

```cs
public static T GetWriteManager<T, U>(this IReadOnlyCollection<IJobsManager<U>> c) 
    where U : TriggeredJob, ContinuousJob
```

```cs
public static T GetWriteManager<T>(this IReadOnlyCollection<T> c) 
    where T : ITriggeredJobsManager, IContinuousJobsManager 
```

Maybe we can create a wrapper type instead of an `IReadOnlyCollection`? something like

```cs
class WebJobManagerContainer
{
    public IContinuousJobsManager PrimaryContinuousManager { get; set; }
    public IContinuousJobsManager SecondaryContinuousManager { get; set; }

    public ITriggeredJobsManager PrimaryTriggeredManager { get; set; }
    public ITriggeredJobsManager SecondaryTriggeredManager { get; set; }
}
```

I guess that'll make it more explicit when compared to just a list.

Kudu.Core/Extensions/JobManagerBaseIReadOnlyCollectionExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+                ? collection.GetPrimaryManager() as ITriggeredJobsManager
+                : collection.GetSecondaryManager() as ITriggeredJobsManager;
+        }
+
+        public static IContinuousJobsManager GetManagerForJob(this IReadOnlyCollection<IContinuousJobsManager> collection, string jobName)
+        {
+            return collection.GetPrimaryManager().HasJob(jobName)
+                ? collection.GetPrimaryManager() as IContinuousJobsManager
+                : collection.GetSecondaryManager() as IContinuousJobsManager;
+        }
```
suwatch: Have you considered an AggregrateContinuousJobsManager : IContinuousJobsManager?   This job managers will wrap the collection of JobManagers and delegate call to them as neccessary.

ahmelsayed: Thank you, that's a much better idea. I updated the PR.

ahmelsayed:
So I tried to have them log into different location which turned out to be too complicated because there are many places in the code where the path is build manually using `Path.Combine()` and constants. Also I kept having to update tests that also assume a location for logs and after changing all that (about 17 files or so) it still wasn't working and I didn't investigate why. 


3528d94d61ab979fe9d992097a6301a15ed43595 instead doesn't allow for jobs to have the same names. 

Jobs in the primary location are loaded first, then jobs with the same name from the secondary location are ignored. I think that's a cleaner approach in general since the APIs will also work as expected.

davidebbo:
@ahmelsayed great, I like that much better. I will do some ad hoc testing!

Kudu.Core/Jobs/BaseJobRunner.cs
```diff
@@ -38,9 +39,12 @@ public abstract class BaseJobRunner
             JobName = jobName;
             _analytics = analytics;
 
-            JobBinariesPath = Path.Combine(Environment.JobsBinariesPath, jobsTypePath, jobName);
-            JobTempPath = Path.Combine(Environment.TempPath, Constants.JobsPath, jobsTypePath, jobName);
-            JobDataPath = Path.Combine(Environment.DataPath, Constants.JobsPath, jobsTypePath, jobName);
+            JobBinariesPath = Path.Combine(basePath, jobName);
+
+            // add a 10 char long hash of the jobBinaries location to the log folder path
+            var hash = JobBinariesPath.GetMD5Hash(10);
```
davidebbo: Why is this new hash necessary? It's breaking test `InPlaceJobRunsInPlace`, which checks that the path it runs from contains `App_Data\jobs\triggered\`. But note that there is already a random component to the temp path, e.g. `TEMP\jobs\triggered\joba\ege02v04.osz`.

More generally, I'm guessing you had added that when both jobs by the same name were allowed to exist, but that it may no longer be relevant now that you enforce uniqueness?

ahmelsayed: your guess is correct. I forgot to undo this change. Just pushed a fix for it. also a change for the delete logic. Now it's

```cs
if (_settings.RunFromLocalZip() || _settings.RunFromRemoteZip())
{
     WriteJobManager.DeleteJob(jobName);
}
else if (PrimaryJobManager.HasJob(jobName))
{
    PrimaryJobManager.DeleteJob(jobName);
}
else
{
    SecondaryJobManager.DeleteJob(jobName);
}
```

Kudu.Core/Jobs/AggregrateJobsManagerBase.cs
```diff
@@ -0,0 +1,81 @@
...
+        protected AggregrateJobsManagerBase(Func<JobsManagerBase<TJob>> primaryManagerAction, Func<IEnumerable<string>, JobsManagerBase<TJob>> secondaryManagerAction, IDeploymentSettingsManager settings)
+        {
+            PrimaryJobManager = primaryManagerAction();
+            SecondaryJobManager = secondaryManagerAction(PrimaryJobManager.ListJobs(false).Select(j => j.Name));
+            WriteJobManager = settings.RunFromLocalZip() || settings.RunFromRemoteZip() ? SecondaryJobManager : PrimaryJobManager;
+            _settings = settings;
+        }
+
+        public void DeleteJob(string jobName)
+            => _settings.RunFromLocalZip() || _settings.RunFromRemoteZip()
```
davidebbo: Semi-colons are not optional in C# ;)

Kudu.Core/Jobs/AggregrateJobsManagerBase.cs
```diff
@@ -0,0 +1,94 @@
...
+{
+    public abstract class AggregrateJobsManagerBase<TJob> where TJob : JobBase, new()
+    {
+        protected JobsManagerBase<TJob> PrimaryJobManager { get; private set; }
+        protected JobsManagerBase<TJob> SecondaryJobManager { get; private set; }
+        protected JobsManagerBase<TJob> WriteJobManager { get; private set; }
+
+        private readonly IDeploymentSettingsManager _settings;
+
+        protected AggregrateJobsManagerBase(Func<JobsManagerBase<TJob>> primaryManagerAction, Func<IEnumerable<string>, JobsManagerBase<TJob>> secondaryManagerAction, IDeploymentSettingsManager settings)
```
davidebbo: 'Factory' might be a bit clearer than 'Action'

Kudu.Core/Jobs/AggregrateJobsManagerBase.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            PrimaryJobManager = primaryManagerAction();
+            SecondaryJobManager = secondaryManagerAction(PrimaryJobManager.ListJobs(false).Select(j => j.Name));
```
davidebbo: Confusing line could use a comment. e.g.

    // pass the list of primary job names so the second manager can excluded them

Also, use `forceRefreshCache: false` instead of just `false`.

Kudu.Core/Jobs/AggregrateJobsManagerBase.cs
```diff
@@ -0,0 +1,94 @@
...
+            WriteJobManager = settings.RunFromLocalZip() || settings.RunFromRemoteZip() ? SecondaryJobManager : PrimaryJobManager;
+            _settings = settings;
+        }
+
+        public void DeleteJob(string jobName)
+        {
+            if (_settings.RunFromLocalZip() || _settings.RunFromRemoteZip())
```
davidebbo: Suggestion, use a wrapper `RunFromZip()` helper for cases where you need either.

Kudu.Core/Jobs/AggregrateJobsManagerBase.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
+                WriteJobManager.DeleteJob(jobName);
+            }
+            else if (PrimaryJobManager.HasJob(jobName))
+            {
+                PrimaryJobManager.DeleteJob(jobName);
+            }
+            else
+            {
+                SecondaryJobManager.DeleteJob(jobName);
```
davidebbo: Do this even if it was in Primary.

Kudu.Core/Jobs/AggregrateJobsManagerBase.cs
ahmelsayed: done

Kudu.Core/Jobs/AggregrateJobsManagerBase.cs
ahmelsayed: done

Kudu.Core/Jobs/AggregrateJobsManagerBase.cs
ahmelsayed: done

Kudu.Core/Jobs/AggregrateJobsManagerBase.cs
ahmelsayed: done

ahmelsayed:
actually this logic 

```
return _settings.RunFromZip() || !PrimaryJobManager.HasJob(jobName)
    ? SecondaryJobManager
    : PrimaryJobManager;
```

is not right. It the job doesn't exist in both, we should use the primary since this will be used in CreateOrUpdate

ahmelsayed:
This is more correct for what we need 

```cs
if (_settings.RunFromZip())
{
    return SecondaryJobManager;
}
else if (PrimaryJobManager.HasJob(jobName))
{
    return PrimaryJobManager;
}
else if (SecondaryJobManager.HasJob(jobName))
{
    return SecondaryJobManager;
}
else
{
    return PrimaryJobManager;
}
```

PR #2734, opened 2018-03-31
Honor execludeList in cleanup of old webjobs. Also fix a typo

ahmelsayed:


ahmelsayed:
actually now that I think of it, this is not sufficient. Primary will still clean up secondary webjobs it doesn't know about

PR #2735, opened 2018-03-31
Implement CleanupDeletedJobs in the aggregate manager

ahmelsayed:
I'll wait until the private CI finishes first. This change shouldn't have any impact on the CI run because the test cases never populate the secondary manager, so the previous change to use the excludeList should be enough for test cases.


davidebbo:
I think this new change looks reasonable. There needs to be only one cleanup pass that knows about everything.

PR #2737, opened 2018-04-03
Fixed clippy img in IE

Kappa013:
Without max height clippy image has height of over 100px in IE

image

### DIFF
--- a/Kudu.Web/Content/Site.css
+++ b/Kudu.Web/Content/Site.css
@@ -114,4 +114,5 @@ a.site-item {
 
 .btn-clipboard img {
   max-width: 13px;
+  max-height: 13px;
 }


Kappa013:
Giving the SVG a max height results in a proper height in IE. By setting the max-height to 15px, the same icon size is held in IE/Chrome compared to the current icon in Chrome.

davidebbo:
I'm not seeing this in Chrome, Edge, or IE. e.g. in IE, I get:

image

Unless I'm looking at the wrong UI?

Kappa013:
It happens on Windows Server 2012 R2, I think it has IE 11 pre-installed. Although I'm not 100% sure about the IE version, and whether you guys support it.

davidebbo:
Change looks harmless enough, so let's go ahead and take it.

PR #2749, opened 2018-04-24
Durable Functions - Support storage connection string in host.json

tohling:
Adding support for reading storage connection string from host.json. Customers of Durable Function can now configure their Durable Functions to use a separate storage connection string for better perf. Sample configuration is as follows, where **OrchestrationConnection** is an app setting:

```
{
    "durableTask" : {
        "connection" : "OrchestrationConnection"
    }
}
```

### DIFF
--- a/Common/Constants.cs
+++ b/Common/Constants.cs
@@ -120,6 +120,7 @@ namespace Kudu
         public const string DynamicSku = "Dynamic";
         public const string AzureWebJobsSecretStorageType = "AzureWebJobsSecretStorageType";
         public const string HubName = "HubName";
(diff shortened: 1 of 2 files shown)


ahmelsayed:
@tohling for v2 apps, we are in the process of moving sync triggers to the runtime. Can you make a similar change to the runtime [`SyncTriggers`](https://github.com/Azure/azure-functions-host/blob/6e4b2eb8e5c3a5b3402a621beef44b7d6d629fd9/src/WebJobs.Script.WebHost/Management/WebFunctionsManager.cs#L169)

PR #2753, opened 2018-04-25
Durable Functions: Modified storage connection name

tohling:
Existing Durable Functions customer is using the following name for the storage connection string property. This change is to ensure that adding support for reading the connection string does not break existing customers.

```
{
    "durableTask" : {
        "azureStorageConnectionStringName" : "OrchestrationConnection"
    }
}
```


davidebbo:
I don't really understand the change, probably because I don't know enough about durable. You make it sound like `azureStorageConnectionStringName` is what one specific customer is using. Or did you mean it is a general documented property name?

tohling:
@davidebbo, yes some of our customers are already using this name and we don't want to break those customers.

PR #2755, opened 2018-04-26
Don't include nuget.org when using custom feed

davidebbo:
This fixes a regression from when we didn't use nuget.org.

### DIFF
--- a/Kudu.Contracts/Settings/DeploymentSettingsExtension.cs
+++ b/Kudu.Contracts/Settings/DeploymentSettingsExtension.cs
@@ -211,9 +211,10 @@ namespace Kudu.Contracts.Settings
             return defaultValue;
         }
 
-        public static string GetSiteExtensionRemoteUrl(this IDeploymentSettingsManager settings)
+        public static string GetSiteExtensionRemoteUrl(this IDeploymentSettingsManager settings, out bool isDefault)
         {
             string value = settings.GetValue(SettingsKeys.SiteExtensionsFeedUrl);
+            isDefault = String.IsNullOrEmpty(value);
             return !String.IsNullOrEmpty(value) ? value : DefaultSiteExtensionFeedUrl;
         }
 
--- a/Kudu.Core/SiteExtensions/SiteExtensionManager.cs
+++ b/Kudu.Core/SiteExtensions/SiteExtensionManager.cs
@@ -758,8 +758,15 @@ namespace Kudu.Core.SiteExtensions
             }
             else
             {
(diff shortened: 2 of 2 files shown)


davidebbo:
@EricSten-MSFT please take a look.

Kudu.Contracts/Settings/DeploymentSettingsExtension.cs
```diff
@@ -211,9 +211,10 @@ private static TimeSpan GetTimeSpan(IDeploymentSettingsManager settings, string
             return defaultValue;
         }
 
-        public static string GetSiteExtensionRemoteUrl(this IDeploymentSettingsManager settings)
+        public static string GetSiteExtensionRemoteUrl(this IDeploymentSettingsManager settings, out bool isDefault)
         {
             string value = settings.GetValue(SettingsKeys.SiteExtensionsFeedUrl);
+            isDefault = String.IsNullOrEmpty(value);
             return !String.IsNullOrEmpty(value) ? value : DefaultSiteExtensionFeedUrl;
```
EricSten-MSFT: Instead of calling ```String.IsNullOrEmpty(value)``` twice, why not just use ```!isDefault``` on the line with the ternary operator?

davidebbo: Good point, I'll change that

Kudu.Core/SiteExtensions/SiteExtensionManager.cs
```diff
@@ -758,8 +758,15 @@ private IEnumerable<SourceRepository> GetRemoteRepositories(string feedUrl)
             }
             else
             {
-                repos.Add(GetSourceRepository(DeploymentSettingsExtension.NuGetSiteExtensionFeedUrl));
-                repos.Add(GetSourceRepository(_settings.GetSiteExtensionRemoteUrl()));
+                // The remote feed URL can either be the default siteextensions.net, or some user override via
+                // SCM_SITEEXTENSIONS_FEED_URL. We only want to add nuget.org to the list if the user
+                // is *not* overriding the feed URL
+                string remoteUrl = _settings.GetSiteExtensionRemoteUrl(out bool isDefault);
```
EricSten-MSFT: Does C# allow you to declare a variable as you pass it into a function, and it has scope outside the call to the function?  Wow...I've learned something new today!

davidebbo: Yeah, it's a fairly recent addition. Very convenient!

davidebbo: e.g. see https://www.codeproject.com/Tips/1175809/Csharp-New-Inline-Out-Variables

PR #2766, opened 2018-05-07
fail deployment when function version mismatch

watashiSHUN:
new behavior compared to vsdeploy:
image
image

this PR is still in progress:
1. add deployment tests
2. currently, if user updated appsettings and do a `git push`, they will see `"Everything up-to-date"` instead of a rebuild of their function app...this is because kudu logic is only triggered when there's a change in repository, not in appsettings

### DIFF
--- a/Kudu.Core/Deployment/Generator/FunctionMsbuildBuilder.cs
+++ b/Kudu.Core/Deployment/Generator/FunctionMsbuildBuilder.cs
@@ -1,12 +1,15 @@
 ﻿using Kudu.Contracts.Settings;
+using Kudu.Core.Infrastructure;
 
 namespace Kudu.Core.Deployment.Generator
 {
     class FunctionMsbuildBuilder : MicrosoftSiteBuilder
     {
+        // we want to fail early (before running msbuild) if the deployment environment mismatch with the target framework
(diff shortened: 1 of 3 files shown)


watashiSHUN:
changes in VShelper is just some refactor, avoid writing duplicate code to retrieve property from csproj, here is a quick summary:

I want to aggregate these lines:
1. https://github.com/projectkudu/kudu/blob/50c9627e0965409b375f5a9ac3f8225c64532da9/Kudu.Core/Infrastructure/VsHelper.cs#L74-L79
2. https://github.com/projectkudu/kudu/blob/50c9627e0965409b375f5a9ac3f8225c64532da9/Kudu.Core/Infrastructure/VsHelper.cs#L97-L108
3. To determine which function version the app is targeting, I need to get `<AzureFunctionsVersion>v2</AzureFunctionsVersion>` from `<PropertyGroup>`

and lastly this code is no longer being used, should be removed with https://github.com/projectkudu/kudu/commit/59561993c49d5935ce91ca062d800a6e57b1ad5f:
https://github.com/projectkudu/kudu/blob/50c9627e0965409b375f5a9ac3f8225c64532da9/Kudu.Core/Infrastructure/VsHelper.cs#L124-L140

PR #2770, opened 2018-05-18
remove code property from function testdata

watashiSHUN:
to be 100% safe, I need to check whether the function is httptrigger, but I don't see a easy way to do it
instead we are looking for the following
```
"code": {
     "name": "code",
     "value": "gibberishxyz..."
}
```

### DIFF
--- a/Kudu.Core/Functions/FunctionManager.cs
+++ b/Kudu.Core/Functions/FunctionManager.cs
@@ -417,8 +417,41 @@ namespace Kudu.Core.Functions
                 }
             }
 
-            return await FileSystemHelpers.ReadAllTextFromFileAsync(testDataFilePath);
+            // TODO use Rabin-Karp
+            var fileContentString = await FileSystemHelpers.ReadAllTextFromFileAsync(testDataFilePath);
+            // if it contains the following json property, need to remove them from the file
+            // "code": {
+            //    "name": "code",
+            //    "value": "gibberishxyz..."
+            // }
+            JObject testDataToAdjust = null;
+            if (fileContentString.Contains("\"code\":"))
+            {
+                try
+                {
+                    var testData = JObject.Parse(fileContentString);
+                    var codeProperty = testData["code"];
+                    if ("code".Equals((string)codeProperty["name"], StringComparison.OrdinalIgnoreCase) && (string)codeProperty["value"] != null)
+                    {
+                        testDataToAdjust = testData;
+                    }
+                }
+                catch (Exception)
+                {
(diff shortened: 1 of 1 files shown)


Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -417,8 +417,41 @@ private string GetFunctionLogPath(string name)
...
+                catch (Exception)
+                {
+                    // failed to retrieve code property from fileContentString, NOOP
+                }
+            }
+
+            if (testDataToAdjust != null)
+            {
+                // TODO logging, since we are touching user data
+                testDataToAdjust.Property("code").Remove();
```
davidebbo: Any reason to not do that directly under the `if ("code".Equals(...))` and avoid the extra `testDataToAdjust`?

Kudu.Core/Functions/FunctionManager.cs
```diff
@@ -417,8 +417,41 @@ private string GetFunctionLogPath(string name)
                 }
             }
 
-            return await FileSystemHelpers.ReadAllTextFromFileAsync(testDataFilePath);
+            // TODO use Rabin-Karp
```
davidebbo: I would expect string.Contains() to already use a decently optimized implementation. Not worth getting fancier here.

Kudu.Core/Functions/FunctionManager.cs
watashiSHUN:
```
testDataToAdjust.Property("code").Remove();
// update the fileContentString
fileContentString = testDataToAdjust.ToString(Formatting.None);
await FileSystemHelpers.WriteAllTextToFileAsync(testDataFilePath, fileContentString);
```
Are you talking about putting this whole chunk in here?
```
if ("code".Equals((string)codeProperty["name"], StringComparison.OrdinalIgnoreCase) && (string)codeProperty["value"] != null)
{
	testDataToAdjust = testData;
}
```
I don't want to do that because I am ignoring any failures from deserialization and parsing, they simply implies that testData needs no further adjustments
```
catch (Exception)
{
	// failed to retrieve code property from fileContentString, NOOP
}
```
And I don't want to swallow any exceptions after the format is matched, ie IO exceptions during rewrite

davidebbo: Ok fair enough.

watashiSHUN:
https://github.com/projectkudu/kudu/commit/6f0570d7a36eb958944c30c17920e35f099321f4 merged

PR #2817, opened 2018-07-19
Supporting Docker CI webhook for Windows Containers

vinisoto:
Leveraging what already exists for App Service for Linux to allow Docker continuous integration for the upcoming Azure App Service Windows Containers.


jvano:
@suwatch is added to the review. #Closed

PR #2821, opened 2018-07-30
Fixed URL returned when using custom routes in /api/functions

ubikusss:
All started with the below issue:
https://github.com/Microsoft/vscode-azurefunctions/issues/484

It went back to Kudu and the problem was that if you were using a custom route it was not taken into consideration.


jvano:
Hi

Thanks,

Joaquin Vano
Azure App Service

PR #2827, opened 2018-08-09
Allow the master key to be retrieved via ARM template / listkeys

davidebbo:


davidebbo:
It's not the right change after all.

PR #2832, opened 2018-08-16
Deprecating Support Portal and created a Landing page

puneetg1983:
Removing the redirect for mawssupport.trafficmanager.net and created a landing page that helps customers find the Diagnostic tools easily.

image


davidebbo:
We ended up backing this one out, and @puneetg1983 will change it to host image in CDN instead of embedded.

PR #2837, opened 2018-08-24
Add ability to remove old zip files

ankitkumarr:
Added code to remove old zip files (default 5) from the file-system.

### DIFF
--- a/Kudu.Contracts/Settings/SettingsKeys.cs
+++ b/Kudu.Contracts/Settings/SettingsKeys.cs
@@ -45,5 +45,6 @@
         public const string DoBuildDuringDeployment = "SCM_DO_BUILD_DURING_DEPLOYMENT";
         public const string RunFromZipOld = "WEBSITE_RUN_FROM_ZIP";  // Old name, will eventually go away
         public const string RunFromZip = "WEBSITE_RUN_FROM_PACKAGE";
+        public const string MaxZipPackageCount = "SCM_MAX_ZIP_PACKAGE_COUNT";
     }
 }
--- a/Kudu.Core/Deployment/DeploymentHelper.cs
+++ b/Kudu.Core/Deployment/DeploymentHelper.cs
@@ -2,6 +2,8 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using Kudu.Core.Tracing;
+using Kudu.Contracts.Tracing;
 using Kudu.Core.Infrastructure;
 using Kudu.Core.SourceControl;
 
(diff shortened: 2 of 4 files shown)


Kudu.Core/Deployment/DeploymentHelper.cs
```diff
@@ -49,5 +51,46 @@ public static bool IsDefaultWebRootContent(string webroot)
...
             return false;
         }
+
+        public static void PurgeZipsIfNecessary(string sitePackagesPath, ITracer tracer)
+        {
+            int DEFAULT_PURGE_VAL = 5;
+            int MIN_PURGE_VAL = 1;
+
+            var totalAllowedZips = DEFAULT_PURGE_VAL;
+            if (System.Environment.GetEnvironmentVariable("PURGE_VAL") != null)
```
davidebbo: - Use SCM_ naming pattern for consistency with other Kudu settings
- Define a constant in `SettingsKeys.cs`
- Use `IDeploymentSettingsManager` to get to the value

e.g. search for `SettingsKeys.DisableSubmodules` for an example that uses this pattern.

Kudu.Core/Deployment/DeploymentHelper.cs
```diff
@@ -49,5 +51,46 @@ public static bool IsDefaultWebRootContent(string webroot)
...
+                {
+                    tracer.TraceError(ex, "Exception thrown while retrieving Environment variable value. Invalid value set in the Environment variable.");
+                    totalAllowedZips = DEFAULT_PURGE_VAL;
+                }
+            }
+            IEnumerable<string> zipFiles = FileSystemHelpers.GetFiles(sitePackagesPath, "*.zip");
+            if (zipFiles.Count() > totalAllowedZips)
+            {
+                // Create a list of files and their last mod date in a keypair and sort in descending order.
+                var fileModDateList = new List<KeyValuePair<string, DateTime>>(zipFiles.Select(fileName => new KeyValuePair<string, DateTime>(fileName, FileSystemHelpers.GetLastWriteTimeUtc(fileName))));
```
davidebbo: This code can probably be simplified by doing an OrderBy on the collection, and avoid the need for a temporary list to do the sorting.

ankitkumarr: You are right! I will update it right away. Thanks!

davidebbo: You can also do `.Skip(totalAllowedZips)` on the enumerable, and then delete everything that's left (assuming descending sort).

Kudu.Core/Deployment/DeploymentHelper.cs
ankitkumarr: I made the update, but please let me know if there's a better place to keep the constants for default and min values `DEFAULT_ALLOWED_ZIPS` and `MIN_ALLOWED_ZIPS`. Thanks again for the feedback.

Kudu.Core/Deployment/DeploymentHelper.cs
ankitkumarr: Cool, I didn't know `.Skip(int)` existed. Thanks!

Kudu.Contracts/Settings/SettingsKeys.cs
```diff
@@ -45,5 +45,6 @@ public static class SettingsKeys
         public const string DoBuildDuringDeployment = "SCM_DO_BUILD_DURING_DEPLOYMENT";
         public const string RunFromZipOld = "WEBSITE_RUN_FROM_ZIP";  // Old name, will eventually go away
         public const string RunFromZip = "WEBSITE_RUN_FROM_PACKAGE";
+        public const string PurgeZipValue = "SCM_PURGE_ZIP_VAL";
```
davidebbo: `SCM_PURGE_ZIP_VAL` does not convey well the meaning. I would call it `SCM_MAX_ZIP_PACKAGE_COUNT`.

PR #2845, opened 2018-08-30
Update kudu to use new kuduscript with python3.6 support

JennyLawrance:


davidebbo:
@JennyLawrance Do you need this deployed on Windows, or Linux only?

JennyLawrance:
@davidebbo Linux only.

PR #2857, opened 2018-09-07
Fixes #2844 - catch unauth exceptions for r/o mode

### LINKED ISSUE #2844: Kudu should work (as much it could) with R/O file system

ankitkumarr:
Catch the Unauthorized exceptions to ensure that kudu works with R/O file system.

### DIFF
--- a/Kudu.Core/Environment.cs
+++ b/Kudu.Core/Environment.cs
@@ -151,7 +151,7 @@ namespace Kudu.Core
         {
             get
             {
-                return FileSystemHelpers.EnsureDirectory(_repositoryPath);
+                return FileSystemHelpers.EnsureDirectoryIgnoreAccessExceptions(_repositoryPath);
             }
             set
             {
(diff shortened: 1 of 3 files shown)


ankitkumarr:
This change ensures that kudu works when `deployments`, `locks`, `LogFiles`, and `site/repository` (all the directories kudu creates) directories are not present in the kudu user home. This enables kudu to start up properly when in r/o mode and unable to create these directories.

Kudu.Core/Environment.cs
```diff
@@ -151,7 +151,15 @@ public string RepositoryPath
         {
             get
             {
-                return FileSystemHelpers.EnsureDirectory(_repositoryPath);
```
davidebbo: I think it may make sense to add a FileSystemHelpers.EnsureDirectoryIgnoreAccessExceptions() helper, and then just call that from all 4 places.

ankitkumarr: That makes sense. Thanks, I will update this PR.

PR #2863, opened 2018-09-18
Set right permissions on the deploy.sh on Linux.

rramachand21:
Set right permissions on the deploy.sh on Linux.

### DIFF
--- a/Kudu.Core/Deployment/Generator/CustomBuilder.cs
+++ b/Kudu.Core/Deployment/Generator/CustomBuilder.cs
@@ -1,6 +1,8 @@
-﻿using System;
+using System;
+using System.IO;
 using System.Threading.Tasks;
 using Kudu.Contracts.Settings;
+using Kudu.Core.Helpers;
 
 namespace Kudu.Core.Deployment.Generator
 {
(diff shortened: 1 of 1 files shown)


Kudu.Core/Deployment/Generator/CustomBuilder.cs
```diff
@@ -16,11 +18,27 @@ public CustomBuilder(IEnvironment environment, IDeploymentSettingsManager settin
...
+                    {
+                        string finalCommandPath = Path.GetFullPath(Path.Combine(RepositoryPath, commandFullPath));
+                        if (File.Exists(finalCommandPath))
+                        {
+                            commandFullPath = finalCommandPath;
+                        }
+                    }
+
+                    context.Logger.Log("Setting execute permissions for " + commandFullPath);
+                    PermissionHelper.Chmod("ugo+x", commandFullPath, Environment, DeploymentSettings, context.Logger);
```
davidebbo: If it's a relative path and the file doesn't exist, won't you end up trying to `chmod` on that relative path here?

JennyLawrance: Should we still do a file.Exists check before the chmod?

rramachand21: Sure, this is all within the kudu container .. if the file doesnt exist, it will throw an error in the Kudu logs and VSCode displays that error which is good.

rramachand21: Regardless of my comment above, I will make sure that the command is under the Repository path if the command is relative path.

Kudu.Core/Deployment/Generator/CustomBuilder.cs
```diff
@@ -16,11 +18,34 @@ public CustomBuilder(IEnvironment environment, IDeploymentSettingsManager settin
 
         public override Task Build(DeploymentContext context)
         {
+            string commandFullPath = _command;
```
davidebbo: This var could be scoped under `if (!OSDetector.IsOnWindows())`

PR #2867, opened 2018-09-20
Add support for zipdeploy using ARM templates

ankitkumarr:
UPDATE: I did test with ARM and was able to successfully deploy.

--------------------------------------------------------------------------------------------------------------
I did not directly test using an ARM template yet. I did test the change using a JSON like this-
```
{
      "apiVersion": "2016-03-01",
      "location": "[parameters('location')]",
      "type": "Microsoft.Web/sites/Extensions",
      "properties": {
        "packageUri": "<url>.zip"
      }
}
```

Assuming that this is the format of the JSON that will be sent by the ARM. Specifically, this part-

```
{
      "properties": {
        "packageUri": "<url>.zip"
      }
}

### CODE REVIEW

**Comment by ankitkumarr** [comment]
I would like to point out this piece of code in `PushDeploymentController` in this PR ,

```
case FetchDeploymentRequestResult.RanSynchronously:
                    if (ArmUtils.IsArmRequest(Request) && isRequestJSON)
                    {
                        DeployResult dr = new DeployResult();
                        response = Request.CreateResponse(HttpStatusCode.OK, ArmUtils.AddEnvelopeOnArmRequest(dr, Request));
                    }
                    else
                    {
                        response.StatusCode = HttpStatusCode.OK;
                    }
                    break;
```
Only if the deployment was successful, I am sending an ARM specific response by creating a blank `DeployResult`. I wasn't sure if this is the best response to be sent. Please let me know if there's a better response that ARM could use. Currently, this response does result in ARM identifying that the resource creation was successful.
Also, if there's some kind of conflict or a badrequest by ARM, the response will only contain an appropriate status code. 

Let me know if any other behavior would be better suitable for ARM.

File: Kudu.Core/Deployment/DeploymentHelper.cs [hunk 1]
```diff
@@ -75,5 +77,31 @@ public static void PurgeZipsIfNecessary(string sitePackagesPath, ITracer tracer,
...
+                        return response.Content;
+                    }
+                    catch (Exception ex)
+                    {
+                        if (--retries <= 0)
+                        {
+                            tracer.TraceError(ex, "Could not make a successful GET request to {0}", uri.AbsoluteUri);
+                            throw ex;
+                        }
+                    }
```
suwatch: sleep like 200 ms before retry

suwatch:
Once we expose this as ARM template, we need to follow ARM async pattern.    Please see PUT SiteExtension for example.

Kudu.Services/Deployment/PushDeploymentController.cs
```diff
@@ -178,7 +220,15 @@ private async Task<HttpResponseMessage> PushDeployAsync(ZipDeploymentInfo deploy
...
+                    if (ArmUtils.IsArmRequest(Request) && isRequestJSON)
+                    {
+                        DeployResult dr = new DeployResult();
+                        response = Request.CreateResponse(HttpStatusCode.OK, ArmUtils.AddEnvelopeOnArmRequest(dr, Request));
+                    }
+                    else
+                    {
+                        response.StatusCode = HttpStatusCode.OK;
+                    }
                     break;
```
suwatch: See PUT SiteExtension for async pattern.

Kudu.Core/Deployment/DeploymentHelper.cs
```diff
@@ -75,5 +78,45 @@ public static void PurgeZipsIfNecessary(string sitePackagesPath, ITracer tracer,
...
+                        return response.Content;
+                    }
+                    catch (Exception ex)
+                    {
+                        if (--retries <= 0)
+                        {
+                            tracer.TraceError(ex, "Could not make a successful GET request to {0}", uri.AbsoluteUri);
+                            throw ex;
+                        }
+                        Thread.Sleep(delay);
```
ahmelsayed: use `await Task.Delay(delay);` instead of `Thread.Sleep` since the latter will block a thread for the sleep.

Kudu.Services/Deployment/DeploymentController.cs
```diff
@@ -451,7 +451,33 @@ public HttpResponseMessage GetResult(string id)
...
+                            responseMessage = Request.CreateResponse(HttpStatusCode.BadRequest, ArmUtils.AddEnvelopeOnArmRequest(result, Request));
+                            break;
+                        case DeployStatus.Success:
+                            responseMessage = Request.CreateResponse(HttpStatusCode.OK, ArmUtils.AddEnvelopeOnArmRequest(result, Request));
+                            break;
+                        default:
+                            responseMessage = Request.CreateResponse(HttpStatusCode.BadRequest, ArmUtils.AddEnvelopeOnArmRequest(result, Request));
+                            break;
+                    }
+                }
```
ahmelsayed: I'm thinking about this. It seems like it'll be a breaking change for any client also listing app specific deployments through ARM. 

Like I think the portal has a deployment list and drill down UI, that will start failing for failed deployments. I think maybe adding a `"provisioningState": "Failed"` property into the response might be a better option.

You could update the `DeployResult` object to have a `provisioningState` field that's only populated in ARM. And then instead of setting the response statusCode, set that filed to `Failed`, `InProgress` or `Succeeded`

Kudu.Services/Deployment/PushDeploymentController.cs
```diff
@@ -127,36 +130,72 @@ public class PushDeploymentController : ApiController
...
+            using (_tracer.Step("Reading the zip URL from the request JSON"))
             {
-                await WriteSitePackageZip(deploymentInfo, _tracer);
+                try
+                {
+                    return requestObject.Value<JObject>("properties").Value<string>("packageUri");
+                }
+                catch (Exception ex)
+                {
+                    _tracer.TraceError(ex, "Error reading the URL from the JSON");
```
ahmelsayed: Maybe also log `requestObject.ToString()` so if you needed to go back and look at it you know what the request content actually was.

ahmelsayed:
Can you also make sure to try a regular `CURL -XPOST --data-binary ....` but with `isAsync=true` with some large file and check that succeeds? I'm not too sure from the code, but just wanna make sure that we don't return a response before reading the whole request content on async zip scenario.

ahmelsayed:
The change looks good to me. @suwatch and @davidebbo if you don't have any further comments, we can go ahead and merge this change into kudu

Kudu.Services.Web/App_Start/NinjectServices.cs
```diff
@@ -450,7 +450,7 @@ public static void RegisterRoutes(IKernel kernel, RouteCollection routes)
             routes.MapHttpRouteDual("zip-put-files", "zip/{*path}", new { controller = "Zip", action = "PutItem" }, new { verb = new HttpMethodConstraint("PUT") });
 
             // Zip push deployment
-            routes.MapHttpRoute("zip-push-deploy", "api/zipdeploy", new { controller = "PushDeployment", action = "ZipPushDeploy" }, new { verb = new HttpMethodConstraint("POST") });
+            routes.MapHttpRouteDual("zip-push-deploy", "zipdeploy", new { controller = "PushDeployment", action = "ZipPushDeploy" }, new { verb = new HttpMethodConstraint("POST", "PUT") });
```
davidebbo: Do we really want both routes to do both things? When it comes to routes, having more ways of doing the same thing is not necessarily a good thing.

Kudu.Services/Deployment/PushDeploymentController.cs
```diff
@@ -127,36 +130,89 @@ public class PushDeploymentController : ApiController
...
-                await WriteSitePackageZip(deploymentInfo, _tracer);
+                try
+                {
+                    deploymentInfo.ZipURL = await GetZipURLFromJSON();
+                }
+                catch (Exception)
+                {
+                    var badResponse = Request.CreateResponse();
+                    badResponse.StatusCode = HttpStatusCode.BadRequest;
+                    return badResponse;
```
davidebbo: Won't that cause a generic 400 that gives no indication about the missing URL in the json?

Kudu.Services/Deployment/PushDeploymentController.cs
```diff
@@ -127,36 +130,89 @@ public class PushDeploymentController : ApiController
...
+            using (_tracer.Step("Reading the zip URL from the request JSON"))
+            {
+                try
+                {
+                    return requestObject.Value<JObject>("properties").Value<string>("packageUri");
+                }
+                catch (Exception ex)
+                {
+                    _tracer.TraceError(ex, "Error reading the URL from the JSON {0}", requestObject.ToString());
+                    throw ex;
```
davidebbo: Just `throw`.

Kudu.Services.Web/App_Start/NinjectServices.cs
ahmelsayed: we can't break `/api/zipdeploy`, are you saying we should leave `/api/zipdeploy` to only handle binary stream and have `/zipdeploy` handle only JSON? this change makes both `/zipdeploy` and `/api/zipdeploy` handles both scenarios.

davidebbo: > are you saying we should leave `/api/zipdeploy` to only handle binary stream and have `/zipdeploy` handle only JSON?

Yes, that's what I meant. But not a big thing, I'm fine either way on this one.

Kudu.Services/Deployment/PushDeploymentController.cs
ankitkumarr: @davidebbo I can update to send an error message back in JSON. I am thinking of adding a check if it's an ARM request then we can send an ARM error response. But, if it's a generic JSON request, we send a simple JSON with the error.

Kudu.Core/Deployment/DeploymentHelper.cs
```diff
@@ -75,5 +77,41 @@ public static void PurgeZipsIfNecessary(string sitePackagesPath, ITracer tracer,
                 }
             }
         }
+
+        public static async Task<HttpContent> PerformGetAsync(HttpClient client, Uri uri, ITracer tracer)
+        {
+            int retries = 3;
+            int delay = 200;
+            HttpResponseMessage response = null;
+            while (true)
```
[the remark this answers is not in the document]
ankitkumarr: I think you are right! Apologies, I did not notice that. I will update to use the `OperationManager.AttemptAsync`.

PR #2868, opened 2018-09-20
Remove env vars and commandline when the request is coming from ARM.

ahmelsayed:


Kudu.Services/Diagnostics/ProcessController.cs
```diff
@@ -456,6 +456,12 @@ private ProcessInfo GetProcessInfo(Process process, string href, bool details =
                 SetEnvironmentInfo(info);
             }
 
+            if (ArmUtils.IsArmRequest(Request))
+            {
+                info.EnvironmentVariables = null;
+                info.CommandLine = null;
+            }
```
ahmelsayed: we could have populating these as a conditional above, but then we will lose the info for 

```csharp
processInfo.IsScmSite 
processInfo.IsWebJob 
processInfo.Description
```

which are deduced from the environment variables.

ahmelsayed:
The only other `GET` that could be problematic is the one for `/dump`, but I don't think would work over ARM, right @suwatch?

Kudu.Services/Diagnostics/ProcessController.cs
```diff
@@ -456,6 +456,12 @@ private ProcessInfo GetProcessInfo(Process process, string href, bool details =
                 SetEnvironmentInfo(info);
             }
 
+            if (ArmUtils.IsArmRequest(Request))
```
davidebbo: Why not just `if (!ArmUtils.IsArmRequest(Request))` in the code above where you set them in the first place?

ahmelsayed: I just mentioned that here https://github.com/projectkudu/kudu/pull/2868#discussion_r219310329 
 
> we could have populating these as a conditional above, but then we will lose the info for
```
processInfo.IsScmSite 
processInfo.IsWebJob 
processInfo.Description
```
>which are deduced from the environment variables.

davidebbo:
@ahmelsayed  right, no way dump would work over ARM!

PR #2871, opened 2018-09-24
Whitelist some process environment variables for ARM Reader role

ahmelsayed:


Kudu.Services/Diagnostics/ProcessController.cs
```diff
@@ -29,6 +29,16 @@ public class ProcessController : ApiController
         private readonly IEnvironment _environment;
         private readonly IDeploymentSettingsManager _settings;
 
+
+        // These settings are not secrets, and are needed by the Visual Studio Snapshot debugger.
+        // Since they are not secrets, it's okay for a Reader over ARM to get them.
+        private readonly IReadOnlyList<string> ArmWhitelistedVaiables = new []
```
ahmelsayed: I think I should call it `_armWhitelistedVariables` to match the others.

PR #2877, opened 2018-10-02
Fix: Read TaskHub from host.json for FunctionsV2

tohling:
The TaskHub used for Durable Functions is moved under the 'extensions' property in Functions V2's host.json. 

We need to add this support for Durable Functions customers using Functions V2.

### DIFF
--- a/Kudu.Core/Helpers/PostDeploymentHelper.cs
+++ b/Kudu.Core/Helpers/PostDeploymentHelper.cs
@@ -349,6 +349,13 @@ namespace Kudu.Core.Helpers
             var json = JObject.Parse(File.ReadAllText(hostConfigPath));
             JToken durableTaskValue;
 
+            // For Functions V2: the 'durableTask' property is set under the 'extensions' property.
+            JToken extensionsValue;
+            if (json.TryGetValue(Constants.Extensions, StringComparison.OrdinalIgnoreCase, out extensionsValue) && extensionsValue != null)
+            {
+                json = (JObject)extensionsValue;
+            }
+
             // we will allow case insensitivity given it is likely user hand edited
             // see https://github.com/Azure/azure-functions-durable-extension/issues/111
             if (json.TryGetValue(Constants.DurableTask, StringComparison.OrdinalIgnoreCase, out durableTaskValue) && durableTaskValue != null)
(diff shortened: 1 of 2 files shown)


Kudu.Core/Helpers/PostDeploymentHelper.cs
```diff
@@ -349,6 +349,13 @@ private static void ReadDurableConfig(string hostConfigPath, out Dictionary<stri
             var json = JObject.Parse(File.ReadAllText(hostConfigPath));
             JToken durableTaskValue;
 
+            // For Functions V2: the 'durableTask' property is set under the 'extensions' property.
+            JToken extensionsValue;
+            if (json.TryGetValue(Constants.Extensions, StringComparison.OrdinalIgnoreCase, out extensionsValue) && extensionsValue != null)
+            {
+                json = (JObject)extensionsValue;
+            }
+
```
suwatch: In corner case if both exists, we will honor V2 format.  Regression paranoid speaking, perhaps we could move this check after the V1 check was not found.

davidebbo: Note that v2 runtime blows up if you have it the old way. Not sure if v1 blows up if you have it the new way. I think this is too much of an edge case to be worth changing. If they have both in there, they probably have bigger problems.

cgillum:
I recall we had a copy of this code elsewhere, perhaps in the functions runtime (for Seabreeze?). Do we need another PR for that copy of this logic?

Kudu.Core/Helpers/PostDeploymentHelper.cs
cgillum: I just tested configuring in both locations for a V2 function app (using the latest CLI) and can confirm that the runtime **did not** blow up. I'll note this as a workaround.

mathewc:
Yes there is a copy of this logic in the runtime, copied initially from this Kudu code: https://github.com/Azure/azure-functions-host/blob/041fc38f0ccdc42e8aa867e860a85b189a4bb4de/src/WebJobs.Script.WebHost/Management/WebFunctionsManager.cs#L293

Kudu.Core/Helpers/PostDeploymentHelper.cs
davidebbo: @cgillum hopefully we can have the fixed deployed soon enough that we don't need to spread this workaround too much.

PR #2880, opened 2018-10-04
Add detectors link

michimune:
This change will allow content developers to directly go to the Diagnose and solve problems page (aka App Service Diagnostics) in the Azure portal from the 503 error page.


michimune:
@suwatch @davidebbo can you please review?

PR #2886, opened 2018-10-12
Fixes #2879 - Improve lock refresh logic

### LINKED ISSUE #2879: Improve ResetLockedStatusFile robustness when refreshing the status lock.
After each deployment, we are trying to reset the status lock (to ensure that we always have it in case we lost it due to file system issues).     The reset status lock logic first releases the lock (dispose the file handle) and then acquires it back immediately (open the file for write access).   Since this is done in a tight window (one call after another) and given the nature of remote file system, I suspected we could run into issues where acquiring it could fail due race with previous call to release file handle.   Statistically per logs, it can happen 0.16% of all job refreshes.

The outcome was job refresh failed leaving the old job running – till next change.   

The improvement is, instead of releasing and acquiring a new one, we could just try to acquire it first.   If failed with UnauthorizedAccessException, then it means we are still have the lock and return.  This should be most common cases.   If we can acquire it (mean we lost it), we will dispose the old handle and assign the new handle to it.    This will avoid release and acquire on every calls.

Internal Ref: https://icm.ad.msft.net/imp/v3/incidents/details/84978825/home

ankitkumarr:


Kudu.Core/Jobs/ContinuousJobLogger.cs
```diff
@@ -26,21 +26,29 @@ public ContinuousJobLogger(string jobName, IEnvironment environment, ITraceFacto
...
-                if (_lockedStatusFile != null)
-                {
-                    _lockedStatusFile.Dispose();
-                }
+                resetLockedStatusFile = File.Open(GetStatusFilePath(), FileMode.OpenOrCreate, FileAccess.Read, FileShare.ReadWrite);
+            }
+            catch (UnauthorizedAccessException)
+            {
+                // This most likely means we still have the lock
+                return;
```
suwatch: throw if _lockedStatusFile is null.

ankitkumarr: Yeah, sorry I missed that. Updating right now.

ahmelsayed: I don't really know what the context of Suwat comment above is, but I don't see a throw if `_lockedStatusFile` is null

ankitkumarr: Right, so @suwatch initially pointed out that if we get an `UnauthorizedAccessException` and the `_lockedStatusFile` is null, that would be a problem.
Then, I discussed with Suwat offline that the lock we are acquiring is not exclusive (it allows sharing for read and write) and we came to the conclusion that we should not hit the case of `UnauthorizedAccessException` and should just make a simple change to acquire a new lock and replace the old one.

PR #2891, opened 2018-10-18
Update runtime API to return core runtime and SDK versions. Fixes #2890

### LINKED ISSUE #2890: Extend /api/diagnostics/runtime to return information about installed .NET Core runtimes
Similar to how we return Node versions, the API should return .NET Core versions.

We need to figure out the schema that makes sense for .NET Core.

davidebbo:


Kudu.Contracts/Diagnostics/DotNetCoreSharedFrameworksInfo.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Newtonsoft.Json;
+using System.Collections.Generic;
+
+namespace Kudu.Services.Diagnostics
+{
+    public class DotNetCoreSharedFrameworksInfo
+    {
+        [JsonProperty(PropertyName = "Microsoft.NetCore.App")]
```
ahmelsayed: nit: in #2890 this is specified as `NETCore` not `NetCore`

davidebbo: Problem is that the serializer ignores the casing altogether and makes everything lower case. Do you know how to control this?

davidebbo: I se, we globally set our `LowerCasePropertyNamesContractResolver()` which makes everything lowercase. Maybe best to keep things this way for consistency with the rest of the kudu API.

davidfowl:
How long until this gets deployed?

davidebbo:
@davidfowl you can try it now in West US 2. Just go to the Kudu root and click 'Runtime Versions'.

Though I do see one bug: if the site is 32 bit, then we end up showing the 32 bit data for both `dotnetcore32` and `dotnetcore64`. The reason is that on 32 bit processes, `Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles)` actually evaluates to `Program Files (x86)`. Seems the way to get `Program Files` regardless of process bitness is `Environment.GetEnvironmentVariable("ProgramW6432")`.

Though this does make me wonder if the API should only return .NET Core info that matches the site bitness, instead of returning both.

davidfowl:
Yea, maybe we just want a "dotnet" since it's per site and you can only be a single architecture. Would there be weirdness with out of proc ANCM if we filtered dotnet based on the selected architecture?

davidebbo:
Yes, it's tricky. The problem with out of proc is that the user can select 32 bit or 64 bit in the Web App configuration, but it has no effect whatsoever on the bitness of the Core process (I think out of proc will always be 32 bit only?). However, it *does* affect the scm site, and hence it will end up returning the 64 bit data, which could then be a mismatch with Core. This could then result in bogus errors from VS telling the user they can't publish when in fact they can.

davidfowl:
OK so maybe the best bet is to always return data for both 64 and 32 bit in any configuration. That way it always works.

davidfowl:
@davidebbo how do we make this work for linux as well?

ahmelsayed:
on linux, the application runtime container and the kudu container's are separate. You'd be getting just what's on the kudu container. Is that what you're looking for?

davidebbo:
Clearly, it's not what they're looking for, since the goal is to determine whether publishing will result in a working app at runtime. So unfortunately, the Kudu driven approach is probably just not going to work here, and we'll need to think about alternatives for Linux. In the end, the result needs to be a function of the selected base image.

PR #2894, opened 2018-10-22
Fixes 2893: only assume json for arm requests

ankitkumarr:
I am yet to test this out extensively. But just wanted to make the PR to get any feedback/opinions.
This should fix the issue for now. But, other suggestions are- 

1) make exclusive /zipdeploy for ARM/JSON requests with PUT requests
2) Only check for JSON requests in PUT requests and keep POST how it was initially


davidebbo:
That's probably good enough. You should try reproing the issue using `curl` and passing the bad mime type, to then make sure it works with your new change.

ankitkumarr:
@davidebbo I ran a couple of tests using `curl`, so far everything looks good.

PR #2901, opened 2018-11-05
Improved portability

jairov4:
**Building in non-US based hosts**
There was a hardcoded "at " string in test code leading to assert failure when localization is set to use language different to English.

**Testing in hosts lacking C: drive**
FileSystem watching test code was hardcoding reference to "C:" drive leading to risky test failure on noisy machines where modifications to C: may happen or C drive is not present. The proposed solution is to use an empty temp directory.

**Support VS2017 Community Edition in build script**
C'mon guys we need a more inclusive society...  :)

### DIFF
--- a/Kudu.Core.Test/FileSystemHubFacts.cs
+++ b/Kudu.Core.Test/FileSystemHubFacts.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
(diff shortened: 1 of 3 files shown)


jairov4:
@suwatch can I you review this please?

suwatch:
looks good to me.  @balag0  could you merge this one as well.

PR #2902, opened 2018-11-05
Fix site creation through Kudu.Web if authentication was specified.

jairov4:
When you use `Kudu.Web` to create sites secured (you have set `<basicAuth username="user" password="password" />` in `web.config`) the flow was failing because SiteManager was not aware of using those credentials.

### DIFF
--- a/Kudu.SiteManagement/SiteManager.cs
+++ b/Kudu.SiteManagement/SiteManager.cs
@@ -4,13 +4,11 @@ using System.Configuration;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
-using System.Net.Http;
 using System.Net.NetworkInformation;
-using System.Runtime.ConstrainedExecution;
-using System.Security.Cryptography.X509Certificates;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


jairov4:
@nickwalkmsft @davidfowl may someone review this change?

Kudu.SiteManagement/SiteManager.cs
```diff
@@ -154,7 +152,8 @@ public async Task<Site> CreateSiteAsync(string applicationName)
                     await OperationManager.AttemptAsync(() => WaitForSiteAsync(serviceUrls.First()));
 
                     // Set initial ScmType state to LocalGit
-                    var settings = new RemoteDeploymentSettingsManager(serviceUrls.First() + "api/settings");
+                    var credentials = _context.Configuration.BasicAuthCredential.GetCredentials();
+                    var settings = new RemoteDeploymentSettingsManager(serviceUrls.First() + "api/settings", credentials);
```
suwatch: This changes the behavior as we will now always pass basic auth regardless of appSettings existence.  Though, I can't think how it would regress any.   

@balag0  could you merge this in?  Might need rebasing.

PR #2904, opened 2018-11-09
Allow Get processes via extension bridge for contributor requests.

balag0:


Kudu.Services/Diagnostics/ProcessController.cs
```diff
@@ -465,7 +465,7 @@ private ProcessInfo GetProcessInfo(Process process, string href, bool details =
                 info.IisProfileTimeoutInSeconds = ProfileManager.IisProfileTimeoutInSeconds;
                 SetEnvironmentInfo(info);
 
-                if (ArmUtils.IsArmRequest(Request))
+                if (ArmUtils.IsArmRequest(Request) && !ArmUtils.IsRbacContributorRequest(Request))
```
suwatch: @ahmelsayed  could you review this as well?   I would think the fix should be in the same location as [this](https://github.com/projectkudu/kudu/commit/8598919e6e1f0babb97625dea30a23e7f5d1672e) and not here.

balag0: its the same location. that code was changed by e54efffd32774bd5bf1312249e484f64384075ea

ahmelsayed: The commit Bala linked to is after the one Suwat linked to. They are both on the same location in the same file, `ProcessController.cs`. 

If the request is ARM && NOT contributer, remove env vars and commandline. I'm assuming `X-MS-CLIENT-ROLEBASED-CONTRIBUTOR` is 1 for ALL scenarios BUT reader? If that's true, it feels like it should have been `IS-READER`, but it doesn't matter.

balag0: Its the other way X-MS-CLIENT-ROLEBASED-CONTRIBUTOR will be 1 only for contributors for GET requests. It will be 0 or missing for all other scenarios.

PR #2918, opened 2019-01-08
Fix the detector deeplink

michimune:
We have an issue where the deeplink to the detector only takes to the top page of App Service Diagnostics instead of the particular detector. This change will fix the issue.


suwatch:
Look ok to me.   Please confirm before we rebase and merge in the fix.

PR #2922, opened 2019-01-22
2907:When accessing javascript library or other static contents, make sure it fall back to local script/contents if CDN is not available

exc3eed:
When accessing javascript library or other static contents, make sure it fall back to local script/contents if CDN is not available


suwatch:
@ahmelsayed   lots of javascript changes - mind reviewing..

ahmelsayed:
if you're just adding all these files to the repo, then why even have the CDN option? It might make things marginally faster but at the expense of an external dependency. My main reason of using a CDN and not committing these files in the repo is because they are not really editable. The right way would be to have the build system bring in these files. If we're not doing that and we're okay with committing these files here then I think we should just remove the dependency on the CDN. @suwatch  what do you think?

exc3eed:
Thanks for the comment.
Falling back to local static content is just for the scenario where cdn doesn’t work (like disconnected onprem environment). If cdn works which is most of the cases (like public azure and onprem with internet connection) we should keep cdn option. 
Also disconnected scenario is very rare. Why should we completely remove CDN option?I think cdn option is better way to maintain static contents for azure.


---
In reply to: [457711265](https://github.com/projectkudu/kudu/pull/2922#issuecomment-457711265) [](ancestors = 457711265)

ahmelsayed:
I prefer CDN as well, but having both seems a bit redundant with unneeded complexity. If we just use the local versions of these files, the changes will be just the urls rather than adding additional checks everywhere. I just don't know how much a CDN would be buying us in this case. I don't have a strong opinion about it, the changes just seem messy to me and any update would have to keep all that in mind.

Kudu.Services.Web/WebHooks.cshtml
```diff
@@ -50,7 +50,15 @@
     </table>
 </div>
 
+<script>
+    var knockoutJsLoaded = false;
+</script>
 <script src="//ajax.aspnetcdn.com/ajax/knockout/knockout-2.2.1.js"></script>
```
ahmelsayed: do you need to set `knockoutJsLoaded` here?

exc3eed:
Ok makes sense. I am removing cdn option.

---
In reply to: [457775781](https://github.com/projectkudu/kudu/pull/2922#issuecomment-457775781) [](ancestors = 457775781)

Kudu.Services.Web/WebHooks.cshtml
exc3eed: Yes. thanks for noticing this. As we are planning to use local dependency I will just change the src = <local path>

---
In reply to: [251172743](https://github.com/projectkudu/kudu/pull/2922#discussion_r251172743) [](ancestors = 251172743)

Kudu.Services.Web/Kudu.Services.Web.csproj
```diff
@@ -18,6 +18,7 @@
     <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
     <TargetFrameworkProfile />
     <UseGlobalApplicationHostFile />
+    <Use64BitIISExpress />
```
ahmelsayed: remove

PR #2932, opened 2019-02-26
Update detector URLs

michimune:
There is another change to the detector deeplink URL.


michimune:
Found an issue. We will revisit it later.

Issue #2953, opened 2019-03-21
Node version is not updated from iisnode.yml and iisnode.yml is never updated on windows


Nickman87:
After noticing that successive deploys on a windows instance failed, because an anciend node version was used, I started to investigate the `selectNodeVersion` script in kudu.

This lead me to the following part:
https://github.com/projectkudu/kudu/blob/a13592e6654585d5c2ee5c6a05fa39fa812ebb84/Kudu.Core/Scripts/selectNodeVersion.js#L336

And I discovered the following issues:

1. The `nodeVersion` variable is never updated when an iisnode.yml file is found, causing the whole script to fallback to the environment defined version (which is undefined for us so fallback to node v 0.10.40 is seen)
2. The iisnode.yml file is generated based on our package.json file, but is never again updated in the future should we change versions

The only solution now for us is to remove the iisnode.yml file every deploy to make sure everything is up to date...

Nickman87:
Created a PR to tackle this issue. If more work needs to be done on this, please let me know!

jvano:
Hi

If the problem persists and is related to running it on Azure App Service, please open a support incident in Azure:
https://learn.microsoft.com/en-us/azure/azure-portal/supportability/how-to-create-azure-support-request

This way we can better track and assist you on this case

Thanks,

Joaquin Vano
Azure App Service

PR #2954, opened 2019-03-21
iisnode.yml automation

Nickman87:
Update iisnode.yml according to package.json every deploy.

Should solve https://github.com/projectkudu/kudu/issues/2953


jvano:
Hi

Thanks,

Joaquin Vano
Azure App Service

PR #2962, opened 2019-04-10
Remove disabled function filter for settriggers

ankitkumarr:
Related to https://github.com/Azure/Azure-Functions/issues/1111 [kudu work-item]

We don't want to filter out disabled functions before making the /settriggers call.

I left out the filter for "excluded", because I don't think it's supported anymore and we probably want to leave it as is.


ankitkumarr:
Approved by @suwatch offline.

PR #2971, opened 2019-05-16
typo cleanup

nschonni:
Ran cSpell across the code base. I lumped this into one PR, but it could be split into individual ones if prefferred.
I also avoided anything that had `public` access so it shouldn't include breaking changes. There were a few of those, that'll I'll submit separately if this lands.


suwatch:
Really appreciate the fix.  Mind rebase to one commit to keep it clean.   We can rebase ourselves as well if you prefer.

nschonni:
@suwatch I think it's better that you squash/merge if that's your process. I'd like to keep it separate to make it easier to split up or remove particular changes if requested during the review

nschonni:
Oh, I think you pulled this in through 787c893a9336beb498252bb2f90a06a95763f9e9 so I'll close this

PR #2980, opened 2019-05-20
fix: continus -> continuos

nschonni:
Maybe breaking since it is in trace message


jvano:
Hi

Thanks,

Joaquin Vano
Azure App Service

jvano:

jvano (approved):

PR #2995, opened 2019-06-20
Add units for log stream timeout message

JasonFreeberg:
Reminding users that the log stream timeout is in seconds.


JasonFreeberg:
@suwatch, this is a minor language edit. Mind if I merge?

PR #3049, opened 2019-10-03
SyncTriggers on Docker WebHook

mathewc:
Issue runtime SyncTriggers when a Docker image update webhook is received.

I'll need to make the same change to KuduLite as well. Working on testing E2E now by hot swapping Kudu binaries in my private stamp and verifying runtime sync is invoked.

### DIFF
--- a/Kudu.Core/Helpers/EnvironmentHelper.cs
+++ b/Kudu.Core/Helpers/EnvironmentHelper.cs
@@ -27,18 +27,6 @@ namespace Kudu.Core.Helpers
         // Is this a Windows Containers site?
         public static bool IsWindowsContainers()
         {
-            // OLD WAY: This can be removed after ANT84
-            string xenonVarString = System.Environment.GetEnvironmentVariable("XENON");
-            int xenonVarValue;
-            if (xenonVarString != null && int.TryParse(xenonVarString, out xenonVarValue))
-            {
-                if (xenonVarValue == 1)
-                {
-                    return true;
-                }
-            }
-
-            // NEW WAY
             string isolation = System.Environment.GetEnvironmentVariable("WEBSITE_ISOLATION");
             return isolation == "hyperv" || isolation == "process";
         }
--- a/Kudu.Core/Helpers/FunctionsHelper.cs
+++ b/Kudu.Core/Helpers/FunctionsHelper.cs
@@ -1,0 +1,72 @@
+﻿using System;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Kudu.Contracts.Tracing;
+using Kudu.Core.Tracing;
+
+namespace Kudu.Core.Helpers
+{
+    public static class FunctionsHelper
+    {
+        private static HttpClient _httpClient;
+
+        internal static HttpClient HttpClient
+        {
+            get
+            {
+                if (_httpClient == null)
+                {
+                    _httpClient = new HttpClient();
+                }
+                return _httpClient;
+            }
+            set
+            {
+                _httpClient = value;
+            }
+        }
+
+        public static async Task SyncTriggersAsync(ITracer tracer)
+        {
+            string uri = GetRuntimeUri("admin/host/synctriggers");
+
+            var request = new HttpRequestMessage(HttpMethod.Post, uri);
+            await SendRuntimeRequestAsync(request, tracer);
+        }
+
+        private static async Task<HttpResponseMessage> SendRuntimeRequestAsync(HttpRequestMessage request, ITracer tracer)
+        {
+            try
+            {
+                var requestId = Guid.NewGuid().ToString();
+                tracer.Trace($"Issuing request to runtime (Method: {request.Method}, Uri: {request.RequestUri}, RequestId: {requestId})");
+
+                request.Headers.UserAgent.Add(PostDeploymentHelper.UserAgent.Value);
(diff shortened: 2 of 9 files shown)


Kudu.Core/Helpers/EnvironmentHelper.cs
```diff
@@ -27,18 +27,6 @@ public static string NormalizeBinPath(string binPath)
         // Is this a Windows Containers site?
         public static bool IsWindowsContainers()
         {
-            // OLD WAY: This can be removed after ANT84
```
mathewc: Saw this old code in here - can be removed now?

Kudu.Core/Helpers/FunctionsHelper.cs
```diff
@@ -0,0 +1,72 @@
...
+                tracer.TraceError(ex, "Request failed");
+                throw;
+            }
+        }
+
+        private static string GetRuntimeUri(string path)
+        {
+            // form the runtime host name from our own SCM host
+            // by removing the scm segement
+            var host = PostDeploymentHelper.HttpHost;
```
mathewc: Is there a better/safer way to get the corresponding runtime URL?

Kudu.Services/Docker/DockerController.cs
```diff
@@ -50,5 +53,32 @@ public HttpResponseMessage ReceiveHook()
...
+                    // 1) No containers are currently running - in this case the sync will start a container with the new image,
+                    //    and the sync will be up to date.
+                    // 2) One or more containers are currently running - in this case the container restart will cause them to
+                    //    restart, and the background sync triggers performed by the runtime will do the sync.
+                    // We do this on a background thread so we don't block the webhook on it. The sync is best effort.
+                    await FunctionsHelper.SyncTriggersAsync(bgTracer);
+                }
+                catch (Exception ex)
+                {
+                    bgTracer.TraceError(ex);
```
mathewc: Chose to do this in the background. Initially I thought it might be good to await the call as part of the webhook - that way if the sync fails, the caller might retry.

Kudu.Core/Helpers/EnvironmentHelper.cs
suwatch: Please remove if no upstream usage.  thank.

Kudu.Core/Helpers/FunctionsHelper.cs
suwatch: Unfortunately no.

Kudu.Core/Helpers/FunctionsHelper.cs
```diff
@@ -0,0 +1,72 @@
...
+            host = host.Replace(".scm", "");
+
```
suwatch: host.Replace(".scm.", ".") to be more specific.

Kudu.Services/Docker/DockerController.cs
suwatch: The current syntrigger failed deployment if sync failed.  This would divert from that, no?

Kudu.Core/Helpers/EnvironmentHelper.cs
mathewc: Joaquin added this originally - I'll confirm with him that this can be removed

Kudu.Services/Docker/DockerController.cs
mathewc: Good point. I went back and forth on this... I'll switch to awaiting this

Kudu.Core/Helpers/EnvironmentHelper.cs
jvano: Yes, that can be removed

---
In reply to: [331590375](https://github.com/projectkudu/kudu/pull/3049#discussion_r331590375) [](ancestors = 331590375)

jvano:
BTW @shrishrirang is making some adjustments to the Docker Webhooks, see PR: https://github.com/projectkudu/kudu/pull/3050

Kudu.Services/Docker/DockerController.cs
```diff
@@ -25,7 +26,7 @@ public DockerController(ITraceFactory traceFactory, IDeploymentSettingsManager s
         }
 
         [HttpPost]
-        public HttpResponseMessage ReceiveHook()
+        public async Task<HttpResponseMessage> ReceiveHook()
```
shrishrirang: >ReceiveHook [](start = 47, length = 11)

@mathewc@microsoft.com : The frontend intercepts /docker/hook requests. Do you see this being called?

mathewc: I will be doing e2e tests before committing, but my understanding from talking with Jenny is that the front end intercepts but also forwards.

shrishrirang: @sanchitmehta can confirm this. But my understanding is that the intercepted requests will not be forwarded to the worker. Also, I believe all this code (DockerController.cs) will go away soon as the new implementation on the front-end will be the default behavior going forward.

mathewc: I was unaware of that. I'll sync up with you guys offline - if this is truly an obsolete hook implementation, then I'll need to do this in the front end rather than Kudu.

mathewc:
Closing this because FE now handles these requests and doesn't forward them to Kudu. We'll need to do this in the FE.

PR #3050, opened 2019-10-04
Trigger a recycle after wardeploy

shrishrirang:
- Current implementation of /wardeploy triggers a recycle of container, only on Xenon and Linux. This recycle is performed using the restartTrigger.txt file.
- This commit adds support to /wardeploy for triggering a recycle of classic Windows web apps in addition to just Xenon and Linux, if the app setting WEBSITE_RECYCLE_PREVIEW_ENABLED is defined. Besides, the recyle will be performed using the /api/app/restart API.
- The new behavior is hidden behind a WEBSITE_RECYCLE_PREVIEW_ENABLED flag, which is added only temporarily. This flag will eventually be removed and the new behavior will become the default.
- Also, this commit removes the "container/webhook" API, as it is not used anymore.


shrishrirang:
@jvano, @suwatch, @sanchitmehta for CR.

Kudu.Core/Deployment/DeploymentManager.cs
```diff
@@ -234,10 +237,24 @@ public void Delete(string id)
...
+                            tracer.Trace("Triggering recycle (preview mode enabled).");
+
+                            await RestartMainSiteAsync(deploymentInfo);
+                        }
+                        else
+                        {
+                            logger.Log("Triggering recycle.");
+                            tracer.Trace("Triggering recycle.");
+
+                            DockerContainerRestartTrigger.RequestContainerRestart(_environment, RestartTriggerReason, tracer);
```
jvano: >Docker [](start = 28, length = 6)

I think we should try not to use the word 'Docker' and just call it ContainerRestartTrigger.

jvano:
>     public static class DockerContainerRestartTrigger

This will become obsolete, shouldn't we change the name and put a comment on when it can be removed?

---
Refers to: Kudu.Core/Infrastructure/DockerContainerRestartTrigger.cs:14 in d7b4edc. [](commit_id = d7b4edc832ef508c62c071c019c0b8f018fdd521, deletion_comment = False)

shrishrirang:
Clarification: Currently, zipdeploy/wardeploy perform a recycle only for Linux and Windows Containers. For wardeploy, we want to trigger a recycle on Windows too. However, we don't want to do this for Windows classic apps (to avoid changing the behavior of existing deployment flows). Introduced `ShouldRestartAfterDeployment` for this purpose, which is set only by wardeploy.

Kudu.Core/Deployment/DeploymentManager.cs
```diff
@@ -327,6 +344,32 @@ public static ChangeSet CreateTemporaryChangeSet(string authorName = null, strin
             };
         }
 
+        private async Task RestartMainSiteAsync(DeploymentInfoBase deploymentInfo)
```
sanchitmehta: I believe this is fire and forget but if this request returns a non success code, should this fail the deployment ?

shrishrirang: Yeah, I was thinking of the same. I pushed a change to return a failure. There's no good way to ensure that the operation is atomic though.

Kudu.Core/Deployment/DeploymentManager.cs
shrishrirang: Leaving this as-is, as discussed.

shrishrirang: Leaving this unchanged, as discussed.

---
In reply to: [331619685](https://github.com/projectkudu/kudu/pull/3050#discussion_r331619685) [](ancestors = 331619685)

Kudu.Core/Deployment/DeploymentManager.cs
shrishrirang: Yeah, I was thinking of the same. I pushed a change to return a failure. There's no good way to ensure that the operation is atomic though.

---
In reply to: [332693077](https://github.com/projectkudu/kudu/pull/3050#discussion_r332693077) [](ancestors = 332693077)

shrishrirang:
>     public static class DockerContainerRestartTrigger

Added a warning, but leaving the name unchanged as discussed.

---
In reply to: [538503095](https://github.com/projectkudu/kudu/pull/3050#issuecomment-538503095) [](ancestors = 538503095)

---
Refers to: Kudu.Core/Infrastructure/DockerContainerRestartTrigger.cs:14 in 843e32c. [](commit_id = 843e32c02a79fa532f122fdf8942fa049f036aef, deletion_comment = False)

Kudu.Core/Deployment/DeploymentManager.cs
```diff
@@ -234,10 +240,24 @@ public void Delete(string id)
...
+                            tracer.Trace("Triggering recycle (preview mode enabled).");
+
+                            await RestartMainSiteAsync(deploymentInfo, logger);
+                        }
+                        else
+                        {
+                            logger.Log("Triggering recycle.");
+                            tracer.Trace("Triggering recycle.");
+
+                            DockerContainerRestartTrigger.RequestContainerRestart(_environment, RestartTriggerReason, tracer);
```
suwatch: this should go to PostDeploymentHelper .. https://github.com/projectkudu/kudu/blob/master/Kudu.Core/Helpers/PostDeploymentHelper.cs#L109-L116.

You can see that we perform many operation and this should be one of them.

Kudu.Contracts/Settings/SettingsKeys.cs
```diff
@@ -42,10 +42,11 @@ public static class SettingsKeys
         public const string TouchWebConfigAfterDeployment = "SCM_TOUCH_WEBCONFIG_AFTER_DEPLOYMENT";
         public const string MaxRandomDelayInSec = "SCM_MAX_RANDOM_START_DELAY";
         public const string DockerCiEnabled = "DOCKER_ENABLE_CI";
-        public const string LinuxRestartAppContainerAfterDeployment = "SCM_RESTART_APP_CONTAINER_AFTER_DEPLOYMENT";
+        public const string RestartAppContainerAfterDeployment = "SCM_RESTART_APP_CONTAINER_AFTER_DEPLOYMENT";
```
suwatch: I would love to make this a more generic feature - not just for container.   It might be useful for MSDeploy or Windows scenario.

shrishrirang: Just to be sure I understand you correctly - You want to re-purpose the app setting to extend its scope to classic Windows apps too, while keeping the name unchanged? That would be a change in behavior for classic Windows apps as the default value of the app setting is assumed to be true. Is that fine?

suwatch: I thought it is off by-default - hence I'd love you to rename to be more generic to support any kind of windows applications as well.   Could we do this instead?

- Rename to more generic (not container specific).
- Lacking of such appsettings - for Windows, default is off and Container, default is on.

shrishrirang: I think, given that we will have to continue supporting SCM_RESTART_APP_CONTAINER_AFTER_DEPLOYMENT (container specific app-setting) for backward compatibility, adding a generic SCM_RESTART_APP_AFTER_DEPLOYMENT would be too confusing. Especially, given that the defaults would be different, it could be confusing what the expected behavior is for different combinations of the following: both app settings defined with conflicting values, one defined and one not defined, etc cases. 

I am thinking of the following.
- Re-purpose SCM_RESTART_APP_CONTAINER_AFTER_DEPLOYMENT (old app-setting) without renaming to work for Windows too. This means, the default behavior for Windows will change, but I guess a restart after a new deployment is the expected behavior anyway. If a particular app does not want this behavior, SCM_RESTART_APP_CONTAINER_AFTER_DEPLOYMENT can be defined explicitly to false.
- SCM_RESTART_APP_CONTAINER_AFTER_DEPLOYMENT would be a little misleading (because the name says 'container') if it is used for Windows apps too, but may be we can live with it.

What do you think?

---
In reply to: [334778351](https://github.com/projectkudu/kudu/pull/3050#discussion_r334778351) [](ancestors = 334778351)

suwatch: Sounds reasonable.   Let me know if you have squash and rebase - I can merge it in for you

PR #3069, opened 2019-10-23
Fix the IsLoopback check

shrishrirang:
- IsLoopback() is an incorrect check. It will be true if a local proxy (on the same machine as the Kudu server) forwards requests to Kudu as it would be a local connection.
- Instead, checking for the Host header is a better check for mocking APIs


shrishrirang:
@jvano @suwatch

jvano:
it looks good to me

:shipit:

PR #3087, opened 2019-12-06
Fix misleading message about the termination of the log streaming session

jvano:
In case we need to recycle the process that host kudu, the current message for the termination is accurate in the context of just kudu but if kudu is a companion of the real app, the message can be misleading into which app actually was terminated. 

This change is just to make the message more precise about what is happening in the context of a log streaming session.


jvano:
@suwatch

PR #3089, opened 2019-12-06
Append WebJobType with /SDK to analytics

Hazhzeng:
### Background
When a webjob is developed by Azure.WebJobs.SDK, we will mark the webjob data directory with **webjobssdk.marker**.

However, when analytics tools writing jobType to Kusto during webjob initialization, creation and deletion, the reported jobType is missing the expected **/SDK** flag.

This change can be used by the webjobs detector to raise warning when visual studio deploys webjobs in an unexpected mode.
(e.g. timer trigger is deployed in triggered mode, but it should use continuous mode instead).

### Fix
This PR is intended to fix JobType developed with SDK in Kudu table. 
Changing **continuous** to **continuous/SDK**
Changing **triggered** to **triggered/SDK**

### DIFF
--- a/Kudu.Core.Test/Jobs/ContinuousJobRunnerFacts.cs
+++ b/Kudu.Core.Test/Jobs/ContinuousJobRunnerFacts.cs
@@ -94,6 +94,12 @@ namespace Kudu.Core.Test.Jobs
         }
 
         [Fact]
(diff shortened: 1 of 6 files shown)


Hazhzeng:
cc: @paulbatum

paulbatum:
Looks good to me.

PR #3101, opened 2019-12-19
Clearing app_offline.htm that may be leftover

ankitkumarr:
I need to test this in a private site.
Also need to fix tests.

### DIFF
--- a/Kudu.Core/Helpers/PostDeploymentHelper.cs
+++ b/Kudu.Core/Helpers/PostDeploymentHelper.cs
@@ -11,6 +11,7 @@ using System.Reflection;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using Kudu.Contracts.Infrastructure;
 using Kudu.Contracts.Settings;
 using Kudu.Contracts.Tracing;
 using Kudu.Core.Deployment;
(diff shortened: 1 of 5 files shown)


Kudu.Core/Helpers/PostDeploymentHelper.cs
```diff
@@ -740,5 +741,33 @@ public static async Task UpdateSiteVersion(ZipDeploymentInfo deploymentInfo, IEn
...
+            var appOfflineFile = Path.Combine(environment.WebRootPath, Constants.AppOfflineFileName);
+            if (FileSystemHelpers.FileExists(appOfflineFile))
+            {
+                if (deploymentLock.IsHeld)
+                {
+                    tracer.Trace($"Deployment lock is held, will not remove {Constants.AppOfflineFileName}");
+                    return;
+                }
+
+                try
```
suwatch: I would try/catch the whole thing to be saved.

ankitkumarr: Good point! I meant that have the callers use "SafeOperation". But catching the whole thing and logging is probably a better option.

Kudu.Core/Helpers/PostDeploymentHelper.cs
```diff
@@ -740,5 +741,33 @@ public static async Task UpdateSiteVersion(ZipDeploymentInfo deploymentInfo, IEn
...
+                if (FileSystemHelpers.FileExists(appOfflineFile))
+                {
+                    if (deploymentLock.IsHeld)
+                    {
+                        tracer.Trace($"Deployment lock is held, will not remove {Constants.AppOfflineFileName}");
+                        return;
+                    }
+
+                    var appOfflineContent = OperationManager.Attempt(() => FileSystemHelpers.ReadAllText(appOfflineFile));
+                    if (appOfflineContent.Contains(Constants.AppOfflineKuduContent))
```
ankitkumarr: can make this ignore case. Not a big deal.

suwatch:
merged commit 91aa9708

Issue #3113, opened 2020-01-28
Forward SyncTriggers requests to Functions Runtime


mathewc:
The Kudu SyncTriggers logic [here](https://github.com/projectkudu/kudu/blob/91aa97089b2f4cab997883d2ce902507ddd7e0ec/Kudu.Core/Helpers/PostDeploymentHelper.cs#L133) doesn't do the extended sync that the Functions runtime does. Specifically, it omits the cached secrets/function info used by other parts of the system for performance/coldstart optimization. Eventually the background sync triggers performed by the Functions runtime will update the cache, so this isn't a huge issue. However, the more we can ensure a full sync is done, the better the system will perform. 

We should update Kudu to forward the request to the Functions runtime. Initially we can try that, and if that request fails, fall back to what we do now.

PR #3185, opened 2020-07-17
update sync triggers to forward to functions runtime

ankitkumarr:
Fixes https://github.com/projectkudu/kudu/issues/3113

Testing in progress.

### DIFF
--- a/Kudu.Core/Helpers/PostDeploymentHelper.cs
+++ b/Kudu.Core/Helpers/PostDeploymentHelper.cs
@@ -135,7 +135,6 @@ namespace Kudu.Core.Helpers
         /// It is written to require least dependencies but framework assemblies.
         /// Caller is responsible for synchronization.
         /// </summary>
-        /// <param name="siteName">WEBSITE_SITE_NAME env</param>
         /// <param name="kind">MSDeploy, ZipDeploy, Git, ..</param>
         /// <param name="requestId">for correlation</param>
         /// <param name="status">Success or fail</param>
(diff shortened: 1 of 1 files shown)


Kudu.Core/Helpers/PostDeploymentHelper.cs
```diff
@@ -479,12 +517,18 @@ private static void WriteAutoSwapOngoing()
             }
         }
 
-        // Throws on failure
-        private static async Task PostAsync(string path, string requestId, string content = null)
+        private static async Task PostKuduAsync(string path, string requestId, string content = null)
```
suwatch, line 520: I would have an overload taking host name (then you can pass scm or main site hostname).   We should preserve the ipaddess semantic since it was there to solve DNS issue.

ankitkumarr:
@suwatch, good point. I will update it right away.

Kudu.Core/Helpers/PostDeploymentHelper.cs
```diff
@@ -163,6 +162,44 @@ public static async Task SyncFunctionsTriggers(string requestId, TraceListener t
...
+
+            // this couples with sync function triggers
+            await SyncLogicAppJson(requestId, tracer);
+        }
+
+        private static async Task<bool> FunctionsRuntimeSyncTriggers(string requestId)
+        {
+            try
+            {
+                var scmUrl = IsLocalHost ? HttpAuthority : HttpHost;
```
suwatch, line 180: this should read scmHostName not scmUrl.

Kudu.Core/Helpers/PostDeploymentHelper.cs
```diff
@@ -163,6 +162,44 @@ public static async Task SyncFunctionsTriggers(string requestId, TraceListener t
...
+
+                var scmSplit = scmUrl.Split('.');
+
+                if (!(scmSplit.Length > 1 && scmSplit[1].Equals("scm", StringComparison.OrdinalIgnoreCase)))
+                {
+                    return false;
+                }
+
+                var functionsSiteUrl = string.Join(".", scmSplit.Where((el, idx) => idx != 1));
```
suwatch, line 189: functionSiteHostName not Url.

Kudu.Core/Helpers/PostDeploymentHelper.cs
mathewc: Or perhaps "hostOrAuthority" as it is in the other location in this file where this code is present

PR #3203, opened 2020-08-26
OneDeploy API - /api/publish

shrishrirang:
- Adds a new /api/publish API for Java deployment scenarios
- /api/publish primarily supports Java scenarios currently, but can be easily extended for additional scenarios
- Overview of new changes
  - OneDeploy reuses bulk of the code from PushDeploymentController and DeploymentManager
  - OneDeployBuilder is a new OneDeploy specific site builder, OneDeployFetch is a OneDeploy specific Fetch handler
  - WatchedFileEnabled/RestartAllowed are additional knobs introduced in DeploymentInfoBase.cs for OneDeploy specific functionality
  - The implementation inspects the WEBSITE_STACK environment variable for certain scenarios. This variable is injected by the App Service runtime
  - OneDeploy uses the new restart API (in other words, it does not rely on touching any file for triggering a recycle)
- Summary of changes to existing code:
  - A bunch of variables were renamed for readability purposes
  - ZipDeploymentInfo is now ArtifactDeploymentInfo (generalized) and is now used by OneDeploy for non-zip artifacts as well
  - Helper functions from ZipDeploymentTests.cs are moved to DeploymentHelperTests.cs/TestFile.cs so that OneDeployTests can use the same helpers
- Subsequent commit will make enhancements to the way OneDeploy handles lib, startup files, etc to improve the defaults

### DIFF
--- a/Common/Constants.cs
+++ b/Common/Constants.cs
@@ -8,8 +8,8 @@ namespace Kudu
         public const string WebRoot = "wwwroot";
         public const string MappedSite = "/_app";
         public const string RepositoryPath = "repository";
-        public const string ZipTempPath = "zipdeploy";
-        public const string ZipExtractPath = "extracted";
+        public const string ZipTempDirectoryName = "zipdeploy";
+        public const string ArtifactStagingDirectoryName = "extracted";
 
         public const string LockPath = "locks";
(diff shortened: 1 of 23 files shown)


jvano:
Looks good! :shipit:

PR #3213, opened 2020-09-23
Add external deployment ID

jjk2673:
In zipdeploy/wardeploy, check incoming headers for a fixed deployment ID, and use it if present.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/projectkudu/kudu?pullRequest=3213) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/projectkudu/kudu?pullRequest=3213) before we can accept your contribution.<br/><br/>:x: jjk2673 [sign now](https://cla.dotnetfoundation.org/projectkudu/kudu?pullRequest=3213)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/projectkudu/kudu?pullRequest=3213) it.</sub>

Kudu.Core/Deployment/DeploymentManager.cs
```diff
@@ -183,6 +183,20 @@ public void Delete(string id)
                     }
                 }
 
+                if (!string.IsNullOrEmpty(deploymentInfo.FixedDeploymentId))
+                {
+                    changeSet = new ChangeSet(
```
shrishrirang, line 188: Should we not create the original `ChangeSet` with the appropriate Deployment ID instead of a GUID?

Kudu.Services/Deployment/PushDeploymentController.cs
```diff
@@ -66,6 +68,14 @@ public class PushDeploymentController : ApiController
...
             using (_tracer.Step("ZipPushDeploy"))
             {
+                string deploymentId = null;
+                IEnumerable<string> idValues;
+
+                if (Request.Headers.TryGetValues(Constants.ScmDeploymentIdHeader, out idValues) && idValues.Count() > 0)
+                {
+                    deploymentId = idValues.ElementAt(0);
+                }
+
```
shrishrirang, line 78: Would be good to have a helper method for doing this like the StatusCode400() method in this class.

Kudu.Core/Deployment/DeploymentManager.cs
jjk2673: (let's continue this conversation on the kudulite thread)

Kudu.Services/Deployment/PushDeploymentController.cs
jjk2673: Agreed.

jjk2673: Taken care of.

Kudu.Core/Deployment/FetchDeploymentManager.cs
```diff
@@ -213,6 +223,14 @@ public async Task PerformDeployment(DeploymentInfoBase deploymentInfo, IDisposab
                                 deploymentInfo: deploymentInfo,
                                 needFileUpdate: deploySpecificCommitId,
                                 fullBuildByDefault: deploymentInfo.DoFullBuildByDefault);
+
+                            if (updateStatusObj != null)
+                            {
+                                // Since build is success we need to force restart on windows
+                                // Directly set this instead of Constants.BuildSuccessful
+                                updateStatusObj.DeploymentStatus = Constants.PostBuildRestartRequired;
```
shrishrirang, line 231: We should do this from RestartMainSiteIfNeeded() by substituting old logic. But because Windows flow is not fully implemented, hide it behind an app setting for now.

@jvano @JasonFreeberg  - FYI

JasonFreeberg: Notes from 10/12 meeting concerning the above:

  - Adjust the hook to check for the custom deployment ID for code apps. If the custom deployment ID exists, then we don't do the restart callback
  - In that case we know that the client is using the new deployment flow, and the restart will be handled at controller

Kudu.Core/Deployment/DeploymentManager.cs
```diff
@@ -707,7 +707,12 @@ private DeployResult GetResult(string id, string activeDeploymentId, bool isDepl
                     {
                         await builder.Build(context);
                         builder.PostBuild(context);
-                        await RestartMainSiteIfNeeded(tracer, logger, deploymentInfo);
+
+                        // Don't request a second container restart if it is already being restarted as part of the PushDeploy workflow
+                        if (string.IsNullOrEmpty(deploymentInfo.DeploymentId))
+                        {
+                            await RestartMainSiteIfNeeded(tracer, logger, deploymentInfo);
+                        }
```
shrishrirang, line 715: Would be better to add this check for DeploymentId along with other checks inside RestartmainSiteIfNeeded. That way RestartmainSiteIfNeeded can be called unconditionally and as the name says - the site will be restarted only if needed.

Also, you may want to consider posting the restart operation to the FE from within RestartMainSiteIfNeeded. RestartmainSiteIfNeeded already has 2 ways of requesting a restart. One way is using a filesystem based notification. Second is using the /api/restart API. The new one being introduced is the third way and unless there is a good reason it would be good to do it from within RestartMainSiteIfNeeded. That way callers can call RestartMainSiteIfNeeded without having to worry about the details. Also - this keeps it consistent with KuduLite.

shrishrirang: Also, make sure that /api/publish?restart=false does not trigger a restart (neither the old way or new way of restarting)

suwatch:
Please squash to one commit and rebase with latest master - let me know when done.

jvano:

Issue #3216, opened 2020-09-28
Change SCM_HTTPCLIENT_TIMEOUT for Mooncake to 5 mins


suwatch:
Mooncake env had long latency issue downloading from nuget.org leading to siteextension installation failure.  We should have a better default for Mooncake env (eg. 5 mins).

jvano:
Hi

If the problem persists and is related to running it on Azure App Service, please open a support incident in Azure:
https://learn.microsoft.com/en-us/azure/azure-portal/supportability/how-to-create-azure-support-request

This way we can better track and assist you on this case

Thanks,

Joaquin Vano
Azure App Service

PR #3222, opened 2020-09-30
Change mooncake httpclient timeout to 5mins

purva-vasudeo:
Changing setting SCM_HTTPCLIENT_TIMEOUT to 5min for Mooncake.

Issue - https://github.com/projectkudu/kudu/issues/3216


Kudu.Contracts/Settings/DeploymentSettingsExtension.cs
```diff
@@ -75,7 +76,26 @@ public static TimeSpan GetLogStreamTimeout(this IDeploymentSettingsManager setti
...
+            if (IsMoonCake())
+            {
+                // Update default timeout for SettingsKeys.HttpClientTimeout for Mooncake
+                defaultTimeout = MoonCakeDefaultHttpClientTimeout;
+            }
+
+            return GetTimeSpan(settings, SettingsKeys.HttpClientTimeout, defaultTimeout);
+        }
+
+        private static bool IsMoonCake()
```
suwatch, line 90: As a habit, make this a public and add unit test.   It served as documentation and what to expect.  I would make an API to take string  eg. IsMoonCake(string currentStampName).   This allows unit testing w/o munging with env.

purva-vasudeo: Changed the method to be test friendly. also added a Fact test for that method.

PR #3236, opened 2020-11-10
Add additional traces to help identify spurious failures

shrishrirang:


Kudu.TestHarness/SitePool.cs
```diff
@@ -87,7 +87,23 @@ private static async Task<ApplicationManager> CreateApplicationInternal(int site
...
+                catch (Exception ex)
+                {
+                    TestTracer.Trace("Error nuking wwwroot: {0}", ex);
+                    
+                    // Not all tests care about starting with a clean wwwroot directory.
+                    // So ignore the exception and continue.
+                }
+
+                wwwrootContent = (await appManager.VfsManager.ListAsync("site/wwwroot")).Select(x => x.Name).ToList();
+                TestTracer.Trace("wwwroot content after nuking it: {0}", string.Join(",", wwwrootContent));
```
suwatch, line 106: Could you make this as part of OneDeploy test instead.  I don't want to slow down other tests that are already working fine w/o this?

shrishrirang: This PR is not changing the existing behavior much. It just makes 2 extra vfs calls to inspect content of wwwroot. Shouldn't add any significant overhead. Do you see any issues?

suwatch: my point is .. it will do this for all 100s of tests we have and they never have issue with reusing site.   since this is OneDeploy only issue, i suggested to move this check to the OneDeploy test instead.

shrishrirang: You have a good point. I made the changes in my repo to restrict the change to just the OneDeploy tests. But I am thinking again - it won't help what I am trying to do here.

Here's some context: This PR doesn't really fix the OneDeploy test flakiness. It just collects additional logs to understand the cause of flakiness. I know that some test is creating a directory called 'app' and for some reason deleting 'app' fails subsequently. Capturing detailed traces for all tests will help understand which test is causing this undesired side-effect so that we can go and fix the issue appropriately in a follow-up PR.

How about we live with the slowness for a week and then revert this PR? That will be enough time to capture the necessary info. Does that sound reasonable?

suwatch: Fair enough .. let try that.

PR #3289, opened 2021-04-15
Removed check for installed file of MSBuild

purva-vasudeo:


Kudu.Core/Infrastructure/PathUtils/PathWindowsUtility.cs
```diff
@@ -159,9 +159,7 @@ internal override string ResolveMSBuild1670Dir()
                     var versionStr = Path.GetFileName(msBuild);
                     if (SemanticVersion.TryParse(versionStr, out SemanticVersion currVersion))
                     {
-                        string installedFilePath = Path.Combine(programFiles, "MSBuilds", $"{versionStr}.installed");
-                        if (FileSystemHelpers.FileExists(installedFilePath) &&
-                            (latestMsBuild == null || currVersion.CompareTo(latestMsBuild) > 0))
+                        if (latestMsBuild == null || currVersion.CompareTo(latestMsBuild) > 0)
```
suwatch, line 162: the flip side is we may using partially installed MSBuild.   Can we scope to ignore the check for just specific version?

PR #3297, opened 2021-05-12
show a banner when using Run From Package

ankitkumarr:


Kudu.Core/Helpers/EnvironmentHelper.cs
```diff
@@ -31,11 +32,17 @@ public static bool IsWindowsContainers()
             return isolation == "hyperv" || isolation == "process";
         }
 
+        public static bool IsRunFromPackage()
+        {
+            string runFromPackage = System.Environment.GetEnvironmentVariable("WEBSITE_RUN_FROM_PACKAGE");
+            return !string.IsNullOrEmpty(runFromPackage) && !StringUtils.IsFalseLike(runFromPackage);
```
suwatch, line 38: also OR with Uri.TryParse - it could be URI format as well.

ankitkumarr: Sounds good. Right now, it checks if `WEBSITE_RUN_FROM_PACKAGE` exists and is not set to `0` or `false`, then we assume it's set to a correct value.
But I can add an explicit URL check as you suggest.

ankitkumarr:
I will update the PR with screenshots of how this looks in kudu shortly.

ankitkumarr:
I took the text from Portal. This text is also what Azure Portal shows when trying to modify files.
## Picture from main page

kudurfp2



## Picture from Console
Kudu-rfp1

suwatch:
Looks great - could you squash it and we will rebase and merge?

PR #3360, opened 2022-04-12
Adds streaming support for zipDeploy

eliaslopezgt:
This PR fixes streaming for ZipDeploy by using `HttpCompletionOption.ResponseHeadersRead` on HttpClient's `GetAsync`.
from https://www.stevejgordon.co.uk/using-httpcompletionoption-responseheadersread-to-improve-httpclient-performance-dotnet.

> The main benefit is for performance. When using this option, we avoid the intermediate MemoryStream buffer, instead of getting the content directly from the stream exposed on the Socket. This avoids unnecessary allocations which is a goal in highly optimised situations.

This is what we are currently observing since we are not doing Streaming properly, we read the entire file when we do `EnsureSuccessStatusCode` and it reflects in the following ways:

| Size | Error |
|------|-------|
|Small      | No error.      |
|0.5 gb| Task Cancelled Error (HttpClient exceeded 100 sec timeout)|
|Less than max buffer size of 2147483647     | System.OutOfMemory Exception (Instance doesn't have enough memory)      |
|More than max buffer size of 2147483647           |     Cannot write more bytes to the buffer than the configured maximum buffer size: 2147483647  |

This PR fixes all three cases.

### DIFF
--- a/Kudu.Core/Deployment/DeploymentHelper.cs
+++ b/Kudu.Core/Deployment/DeploymentHelper.cs
@@ -9,6 +9,7 @@ using Kudu.Core.Infrastructure;
 using Kudu.Core.SourceControl;
 using System.Net.Http;
 using System.Threading.Tasks;
+using Kudu.Core.Settings;
 
 namespace Kudu.Core.Deployment
 {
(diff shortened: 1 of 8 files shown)


suwatch:
Looks good - let me know if it is ready for merge and I will merge it.  Thanks!

PR #3368, opened 2022-05-09
Added Validation API for Zip Deploy

patelchandni:
Clients can call this API after updating app settings and before zip deploy call to check:
1. Mismatched language to Azure app.
2. (Windows .NET only) bitness check - deploying 64-bit code to 32 bit app.
3. Check if package URL is accessible from kudu for addressing deployments with networking issues.
4. Client tools will convert their flags into app settings and set those flags. Then call this API to check if the settings are propagated correctly before deployment.
5. Logical conflicts like RFP = SDBDD = 1, RFP = URL, and recommend RFP = 1.


Kudu.Services.Web/App_Start/NinjectServices.cs
```diff
@@ -476,6 +476,7 @@ public static void RegisterRoutes(IKernel kernel, RouteCollection routes)
             // Zip push deployment
             routes.MapHttpRouteDual("zip-push-deploy", "zipdeploy", new { controller = "PushDeployment", action = "ZipPushDeploy" }, new { verb = new HttpMethodConstraint("POST", "PUT") });
             routes.MapHttpRoute("zip-war-deploy", "api/wardeploy", new { controller = "PushDeployment", action = "WarPushDeploy" }, new { verb = new HttpMethodConstraint("POST") });
+            routes.MapHttpRouteDual("validate-zip-deploy", "validatezipdeploy", new { controller = "PushDeployment", action = "ValidateZipDeploy" }, new { verb = new HttpMethodConstraint("GET") });
```
suwatch, line 479: This should be POST as we are performing validation.

suwatch: Also make sure we can call this from ARM. the route should be api/zipdeploy/validate.  Also no need to Map..Dual - that is for old api.;   New api should be just MapHttpRoute.

patelchandni: Thank you for reviewing. 
1. I have tested the API with and without JSON request. I can't test the ARM template locally, so I'll test it on private stamp after adding extension, like I did last time. I'll definitely test it thoroughly there to see if any other changes will be required in the ARM layer for it to work. 
2. I have made the above changes. Are there any other changes required?

Kudu.Services.Web/App_Start/NinjectServices.cs
```diff
@@ -475,8 +475,9 @@ public static void RegisterRoutes(IKernel kernel, RouteCollection routes)
 
             // Zip push deployment
             routes.MapHttpRouteDual("zip-push-deploy", "zipdeploy", new { controller = "PushDeployment", action = "ZipPushDeploy" }, new { verb = new HttpMethodConstraint("POST", "PUT") });
+            routes.MapHttpRoute("validate-zip-deploy", "zipdeploy/validate", new { controller = "PushDeployment", action = "ValidateZipDeploy" }, new { verb = new HttpMethodConstraint("POST") });
```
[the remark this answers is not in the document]
patelchandni: @suwatch added MapHttpRouteDual. Also, added couple tests for validation API.

Kudu.Services/Deployment/PushDeploymentController.cs
```diff
@@ -852,5 +836,272 @@ private string GetCorrelationId(HttpRequestMessage request)
...
+                    pathmsg = string.Concat(isfunction, "ZipDeploy. Extract zip.");
+                }
+            }
+            _tracer.Trace(pathmsg);
+            return pathmsg;
+        }
+
+        [HttpPost]
+        [HttpPut]
+        public async Task<HttpResponseMessage> ValidateZipDeploy(
```
suwatch, line 874: this should only support POST - not PUT (it does not update resource).

Kudu.Services.Web/App_Start/NinjectServices.cs
```diff
@@ -475,8 +475,9 @@ public static void RegisterRoutes(IKernel kernel, RouteCollection routes)
 
             // Zip push deployment
             routes.MapHttpRouteDual("zip-push-deploy", "zipdeploy", new { controller = "PushDeployment", action = "ZipPushDeploy" }, new { verb = new HttpMethodConstraint("POST", "PUT") });
+            routes.MapHttpRouteDual("validate-zip-deploy", "zipdeploy/validate", new { controller = "PushDeployment", action = "ValidateZipDeploy" }, new { verb = new HttpMethodConstraint("POST", "PUT") });
```
suwatch, line 478: POST only.

Kudu.Services.Web/App_Start/NinjectServices.cs
```diff
@@ -475,8 +475,9 @@ public static void RegisterRoutes(IKernel kernel, RouteCollection routes)
...
             routes.MapHttpRoute("zip-war-deploy", "api/wardeploy", new { controller = "PushDeployment", action = "WarPushDeploy" }, new { verb = new HttpMethodConstraint("POST") });
-
+
```
suwatch, line 480: remove spaces

Kudu.Services/Deployment/PushDeploymentController.cs
```diff
@@ -852,5 +836,272 @@ private string GetCorrelationId(HttpRequestMessage request)
...
+                if (StringUtils.IsTrueLike(remotebuild))
+                {
+                    pathmsg = string.Concat(isfunction, "ZipDeploy. Extract zip. Remote build.");
+                }
+                else
+                {
+                    pathmsg = string.Concat(isfunction, "ZipDeploy. Extract zip.");
+                }
+            }
+            _tracer.Trace(pathmsg);
```
suwatch, line 868: do we have full control of pathmsg?  to be safe, _tracer.Trace("{0}", pathmsg)

Kudu.Services/Deployment/PushDeploymentController.cs
```diff
@@ -87,32 +92,11 @@ public class PushDeploymentController : ApiController
...
-                    {
-                        deploymentInfo.DeploymentPath = "ZipDeploy. Extract zip. Remote build.";
-                    }
-                    else
-                    {
-                        deploymentInfo.DeploymentPath = "ZipDeploy. Extract zip.";
-                    }
-                }
-                _tracer.Trace(deploymentInfo.DeploymentPath);
+                deploymentInfo.DeploymentPath = ZipDeployPathInfo();
```
suwatch, line 99: GetZipDeployPathInfo()

PR #3379, opened 2022-06-27
MSBuild version map added for Dotnet 7

shpraka:
Added the MSBuild version mapping for dotnet 7.

### DIFF
--- a/Kudu.Core/Infrastructure/PathUtils/PathWindowsUtility.cs
+++ b/Kudu.Core/Infrastructure/PathUtils/PathWindowsUtility.cs
@@ -137,6 +137,9 @@ namespace Kudu.Core.Infrastructure
             if (VsHelper.IsDotNet31Version(targetFramework) && ScmHostingConfigurations.UseMSBuild167ForDotnet31)
             {
                 probPaths.Add(Path.Combine(programFiles, "MSBuilds", "16.7.0", "MSBuild", "Current", "Bin"));
+            } else if (VsHelper.IsDotNet7Version(targetFramework))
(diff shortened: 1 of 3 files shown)


Kudu.Core/Infrastructure/PathUtils/PathWindowsUtility.cs
```diff
@@ -137,6 +137,9 @@ internal override string ResolveMSBuild16Dir(string targetFramework)
             if (VsHelper.IsDotNet31Version(targetFramework) && ScmHostingConfigurations.UseMSBuild167ForDotnet31)
             {
                 probPaths.Add(Path.Combine(programFiles, "MSBuilds", "16.7.0", "MSBuild", "Current", "Bin"));
+            } else if (VsHelper.IsDotNet7Version(targetFramework))
+            {
+                probPaths.Add(Path.Combine(programFiles, "MSBuilds", "17.1.0", "MSBuild", "Current", "Bin"));
```
suwatch, line 142: should we write code to pick latest MSBuilds 17.x  - in this case, 17.2.0?   My thinking is - 1) this allows build to pick latest MSBuild 17.x automatically - 2) this allows us to remove old MSBuild like 17.0 or 17.1 w/o worrying about it being hardcoded in Kudu code.

To pick latest 17.x - make sure to sort by SematicVersion and ensure installed marker exists.

shpraka: That's a good point. Can we make that change in separate PR?

PR #3383, opened 2022-07-14
Deploy static files to any directory under /home

dannysongg:
Static files now can be deployed to any directory under /home. Static files cannot be deployed with type=clean to protect user from wiping the entire /home directory. Removing type=lib and type=script as they both just function the same way type=static does.


suwatch:
unless existing, please add a test to cover it

dannysongg:
@suwatch I have edited the existing tests to cover the new behavior

PR #3405, opened 2022-10-07
Patelchandni/ziptoperror

patelchandni:
Only for function app with run from package setup, if we get "Offset to Central Directory cannot be held in an Int64" error, then adding a retry to download zip file again and extract triggers.

### DIFF
--- a/Kudu.Services/ByteRanges/HttpRequestMessageExtensions.cs
+++ b/Kudu.Services/ByteRanges/HttpRequestMessageExtensions.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.ComponentModel;
 using System.Diagnostics.CodeAnalysis;
+using System.IO;
 using System.Net;
 using System.Net.Http;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


suwatch:
please rebase and squash to one commit once ready for review.

patelchandni:
Adding @eliaslopezgt for review as well.

Kudu.Services/ByteRanges/HttpRequestMessageExtensions.cs
```diff
@@ -41,34 +42,48 @@ public static HttpResponseMessage CreateErrorResponse(this HttpRequestMessage re
             return rangeNotSatisfiableResponse;
         }
 
-        public static async Task<long> CopyToAsync(this HttpContent content, string filePath, ITracer tracer)
+        public static async Task<long> CopyToAsync(this HttpContent content, string filePath, ITracer tracer, bool leaveOpen = false)
```
eliaslopezgt, line 45: Consider creating a new method, and based on a feature flag switch between them.

Kudu.Services/Deployment/PushDeploymentController.cs
```diff
@@ -384,161 +384,172 @@ private async Task<HttpResponseMessage> PushDeployAsync(ArtifactDeploymentInfo d
...
 
-            return response;
+                return response;
+            }
+            finally
+            {
+                if (isRequestJSON != true && artifactType == ArtifactType.Zip && _settings.RunFromLocalZip()) 
+                {
+                    var src = await content.ReadAsStreamAsync();
+                    src.Dispose();
```
eliaslopezgt, line 550: null check?

eliaslopezgt (changes requested):
👍 to test this with some comments.

Kudu.Services/Deployment/PushDeploymentController.cs
patelchandni: added check

Kudu.Services/ByteRanges/HttpRequestMessageExtensions.cs
patelchandni: added new method that is only called for function app with run from package path

PR #3410, opened 2022-10-18
Backup stream for deployment retry and check disk space in validation

patelchandni:
1. Storing backup MemoryStream to copy zip again. This is needed for async deployments.
2. Checking local disk space in validation API. 

Use this link to compare all changes together (this PR + my last merged PR): 
https://github.com/projectkudu/kudu/compare/ed6e9c4..3fe359d

### DIFF
--- a/Kudu.Services/Deployment/PushDeploymentController.cs
+++ b/Kudu.Services/Deployment/PushDeploymentController.cs
@@ -37,6 +37,8 @@ namespace Kudu.Services.Deployment
         private const string CorrelationID = "x-ms-correlation-id";
         private const string ZipDeployValidationError = "ZipDeploy Validation ERROR";
         private const string ZipDeployValidationWarning = "ZipDeploy Validation WARNING";
+        //to store the stream especially for async calls so that we can copy content later on
+        private MemoryStream _backupStream;
 
         private readonly IEnvironment _environment;
         private readonly IFetchDeploymentManager _deploymentManager;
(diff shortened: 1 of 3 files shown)


Kudu.Services/Deployment/PushDeploymentController.cs
```diff
@@ -846,14 +854,28 @@ private async Task WriteSitePackageZip(ArtifactDeploymentInfo zipDeploymentInfo,
...
+
+            // Make sure D:\home\data\SitePackages exists
+            FileSystemHelpers.EnsureDirectory(_environment.SitePackagesPath);
+
+            using (tracer.Step("Saving request content to {0}", filePath))
+            {
+                if (_backupStream != null)
                 {
-                    await content.CopyToAsync(filePath, tracer);
+                    _backupStream.Dispose();
```
suwatch, line 875: Should set to null - to avoid leaving non null disposed memstream.

Kudu.Services/Deployment/PushDeploymentController.cs
```diff
@@ -611,38 +607,50 @@ private async Task LocalZipHandler(IRepository repository, DeploymentInfoBase de
...
-                    {
-                        await WriteSitePackageZip(zipDeploymentInfo, _tracer, Request.Content);
+                        if (!string.IsNullOrEmpty(zipDeploymentInfo.RemoteURL))
+                        {
+                            var content = await DeploymentHelper.GetArtifactContentFromURLAsync(zipDeploymentInfo, tracer);
+                            await content.CopyToAsync(filePath, tracer);
+                        }
+                        else
+                        {
+                            await Request.Content.CopyFromBackupAsync(filePath, tracer, _backupStream);
```
suwatch, line 635: I would do using here and set to null.   CopyFromBackupAsync routine should not dispose the object passing to it as it does not own it.

PR #3430, opened 2023-01-25
Add SCM config to use latest msbuild for 14 and 15

shpraka:
Add SCM config to use latest msbuild for MSbuild 14 and 15.

- UseLatestMSBuild16InsteadOfMSBuild14
- UseLatestMSBuild16InsteadOfMSBuild15

### DIFF
--- a/Kudu.Core/Infrastructure/PathUtils/PathWindowsUtility.cs
+++ b/Kudu.Core/Infrastructure/PathUtils/PathWindowsUtility.cs
@@ -108,6 +108,11 @@ namespace Kudu.Core.Infrastructure
 
         internal override string ResolveMSBuild15Dir()
         {
+            if (ScmHostingConfigurations.UseLatestMSBuild16InsteadOfMSBuild15)
+            {
+                return ResolveLatestMSBuild16Dir();
+            }
+
             string programFiles = SystemEnvironment.GetFolderPath(SystemEnvironment.SpecialFolder.ProgramFilesX86);
             List<string> probPaths = new List<string>
             {
@@ -140,7 +145,7 @@ namespace Kudu.Core.Infrastructure
             }
             else if (VsHelper.IsDotNet7Version(targetFramework))
             {
-                // Using ResolveMSBuild1670Dir as it's picking the latest MSBuild version.
+                // Using ResolveLatestMSBuildDir as it's picking the latest MSBuild version.
                 probPaths.Add(ResolveLatestMSBuildDir());
             }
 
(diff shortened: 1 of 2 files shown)


Kudu.Core/Infrastructure/PathUtils/PathWindowsUtility.cs
```diff
@@ -108,6 +108,11 @@ internal override string ResolveNpmJsPath()
 
         internal override string ResolveMSBuild15Dir()
         {
+            if (ScmHostingConfigurations.UseLatestMSBuild)
+            {
+                return Path.Combine(ResolveLatestMSBuildDir());
```
suwatch, line 113: I would pick and pin to one MSBuild version.   This ensures one deterministic version and, more importantly, no impact/regression if we ship new MSBuild for those scenario.

shpraka: yeah that makes sense. 

How about different scm config for each version something like: UseLatestMSBuildForMSBUILD14 and UseLatestMSBuildforMSBUILD15?

shpraka: Updated the PR.

Kudu.Core/Infrastructure/PathUtils/PathWindowsUtility.cs
```diff
@@ -108,6 +108,11 @@ internal override string ResolveNpmJsPath()
 
         internal override string ResolveMSBuild15Dir()
         {
+            if (ScmHostingConfigurations.UseLatestMSBuildInsteadOfMSBuild15)
+            {
+                return Path.Combine(ResolveLatestMSBuildDir());
```
suwatch, line 113: Let say we ship new MSBuild tomorrow, will ResolveLatestMSBuildDir return that new MSBuild version?   If so, it means that every time we ship new MSBuild, we risk of breaking many apps, correct?   To avoid this, we should nominate one MSBuild (say latest one as off today) to be used instead of MSBuild 14 and 15.    Shipping new MSBuild would not impact that.  Thought?

shpraka: That's a good point. 

I agree it will increase the risk of breaking apps because of compatibility issues or something like that. However by pinning to one version, we will tie our hands to maintaining that particular version while adding new MSBuild versions as new dotnet versions will require new MSBuild versions. 

Since MSBuild size keeps increasing, it would contribute to RU size as well. 

Thoughts?

shpraka: Updated the PR and pointed to latest version of MSBuild 16.

PR #3432, opened 2023-01-30
User/finvamp1/update case branches

FinVamp1:
Adding Additional changes for the RBAC Paths

### DIFF
--- a/Kudu.Core.Test/TraceModuleFacts.cs
+++ b/Kudu.Core.Test/TraceModuleFacts.cs
@@ -56,6 +56,8 @@ namespace Kudu.Core.Test
         [InlineData("/api/continuouswebjobs/1/dump", false)]
         [InlineData("/api/functions", true)]
         [InlineData("/api/functions/", true)]
+        [InlineData("/api/functions/config", true)]
+        [InlineData("/api/functions/config/", true)]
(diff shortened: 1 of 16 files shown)


Kudu.Services.Web/Tracing/TraceModule.cs
```diff
@@ -8,6 +8,7 @@
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Web;
+using System.Web.DynamicData;
```
mathewc, line 11: What is this new using statement for?

Kudu.Core.Test/TraceModuleFacts.cs
```diff
@@ -69,6 +69,10 @@ public class TraceModuleFacts
         [InlineData("/API/functions/ADMIN/mAsterkEy/key", false)]
         [InlineData("/api/FUNCTIONS/ADMIN/token", false)]
         [InlineData("/API/FUNCtions/admin/TOKEN?key=1", false)]
+        [InlineData("api/functions/admin/download", false)]
```
mathewc, line 72: Let's add api/functions/config for completeness here. It'll return true.

mathewc (approved):

PR #3436, opened 2023-02-09
Fix typo in GitHub Issue Template

SeanKilleen:
Happened to see it while submitting an issue, figured I'd contribute the fix.


jvano:
Hi

Thanks,

Joaquin Vano
Azure App Service

SeanKilleen:
@jvano I'm not sure your reply was intended here? This is a  typo fix, I don't believe there should be any support ticket required.

jvano:

jvano (approved):

PR #3450, opened 2023-03-16
Adding Safe Directory Git Config

shpraka:


Kudu.Services/GitServer/Helpers.cs
```diff
@@ -59,5 +60,67 @@ public static void PktFlush(this Stream response)
...
+            process.BeginErrorReadLine();
+
+            // waiting for 10 seconds
+            process.WaitForExit(10000);
+
+            if(process.ExitCode != 0 || error.Length != 0)
+            {
+                throw new Exception($"Process exited with {process.ExitCode} and error: {error}");
+            }
+            if (!string.IsNullOrWhiteSpace(output) && output.IndexOf('*') == -1)
```
surenderssm, line 98: what is the significance of the * here?

shpraka: This is for all the directories.

shpraka:
@suwatch your changes looks more clean. should we add that then? Also I couldn't understand what you meant by userprofiles scenario? Using %HOME%/.gitconfig should suffice right?

suwatch:
It needs refinement and more testing.   For instance, if customer already have the git config, how do we check that settings already there and avoid checking the next time (along the same lines as installation txt for hint).   we need to test when user profile is enabled since it may not be on this location.

Kudu.Services/GitServer/Helpers.cs
suwatch: output.IndexOf('*') < 0

PR #3454, opened 2023-03-28
support site jwt token format

suwatch:


Kudu.Core/Helpers/SiteTokenHelper.cs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        // SiteTokenIssuingMode
+        // 0: (default) add both x-ms-site-restricted-token and x-ms-site-token
+        // 1: add x-ms-site-restricted-token only
+        // 2: add x-ms-site-token only
+        public static bool ShouldAddSiteRestrictedToken() => 2 != ScmHostingConfigurations.SiteTokenIssuingMode;
+        public static bool ShouldAddSiteToken() => 1 != ScmHostingConfigurations.SiteTokenIssuingMode;
+
+        private static byte[] GetWebSiteAuthEncryptionKey()
```
mathewc, line 67: Rather than copy this code from https://github.com/projectkudu/kudu/blob/98ad238b860f81a4cb4e3419c8785a58ba68e661/Kudu.Core/Helpers/SimpleWebTokenHelper.cs#L57, can we share it across both classes?

Kudu.Core/Helpers/SiteTokenHelper.cs
```diff
@@ -0,0 +1,87 @@
...
+        // Setter for negative testing purposes
+        public static string SiteTokenAudience
+        {
+            get => _siteTokenAudience ?? "https://appservice.core.azurewebsites.net";
+            set => _siteTokenAudience = value;
+        }
+
+        public static string SiteTokenIssuer
+        {
+            get => _siteTokenIssuer ?? "https://appservice.core.azurewebsites.net";
```
mathewc, line 35: Should the issuer be https://<site>.scm.azurewebsites.net?

suwatch: Fixed.

Kudu.Core/Helpers/SiteTokenHelper.cs
suwatch: We will get rid off the other class eventually.   So duplicate method for now.

PR #3465, opened 2023-06-13
Add safe directory config to `.gitconfig` on startup

sarsharma:
The latest windows-git throws this error:
`fatal: detected dubious ownership in repository at xyz/abc`

We need to run ` git config --global --add safe.directory *` to suppress this error

Instead of running command using git.exe, we can edit the `.gitconfig` file to append this section
```
[safe]
    directory = "*"
```
Also, to prevent this step on every start, we can keep a marker file to know the status of config

### DIFF
--- a/Kudu.Services.Web/App_Start/NinjectServices.cs
+++ b/Kudu.Services.Web/App_Start/NinjectServices.cs
@@ -394,6 +394,11 @@ namespace Kudu.Services.Web.App_Start
             // Temporary fix for https://github.com/npm/npm/issues/5905
             EnsureNpmGlobalDirectory();
             EnsureUserProfileDirectory();
+            
+            if(ScmHostingConfigurations.EnsureGitSafeDirectoryEnabled)
+            {
+                EnsureGitConfigFileWithSafeDirectoryConfig(environment);
+            }
         }
 
         public static class SignalRStartup
@@ -626,6 +631,23 @@ namespace Kudu.Services.Web.App_Start
         private static void RemoveTempFileFromUserDrive(IEnvironment environment)
         {
             FileSystemHelpers.DeleteDirectorySafe(Path.Combine(environment.RootPath, "data", "Temp"), ignoreErrors: true);
+        }
+
+        // add/update gitconfig file to be equivalent of running
+        // git config --global --add safe.directory "*"
+        private static void EnsureGitConfigFileWithSafeDirectoryConfig(IEnvironment environment)
+        {
+            var markerFilePath = Path.Combine(environment.RootPath, "gitsafedirectory.marker");
+            if (File.Exists(markerFilePath))
+            {
+                return;
+            }
+            OperationManager.SafeExecute(() =>
+            {
+                var gitConfigPath = Path.Combine(environment.RootPath, ".gitconfig");
+                File.AppendAllLines(gitConfigPath, new[] {"[safe]", "    directory = *" });
+                File.WriteAllText(markerFilePath, string.Empty);
+            });
         }
 
         // Perform migration tasks to deal with legacy sites that had different file layout
--- a/Kudu.Core/Settings/ScmHostingConfigurations.cs
+++ b/Kudu.Core/Settings/ScmHostingConfigurations.cs
@@ -145,6 +145,12 @@ namespace Kudu.Core.Settings
             get { return GetValue("DataCopyingTelemetryEnabled", "0") != "0"; }
         }
 
+        public static bool EnsureGitSafeDirectoryEnabled
+        {
+            // this is disabled by default
+            get { return GetValue("EnsureGitSafeDirectoryEnabled", "0") == "1"; }
+        }
+
         // SiteTokenIssuingMode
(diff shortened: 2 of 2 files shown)


kumaraksh1 (approved):
Looks good to me !!

Kudu.Services.Web/App_Start/NinjectServices.cs
```diff
@@ -628,6 +629,25 @@ private static void RemoveTempFileFromUserDrive(IEnvironment environment)
...
+        // add/update gitconfig file to be equivalent of running
+        // git config --global --add safe.directory "*"
+        private static void EnsureGitConfigFileWithSafeDirectoryConfig(IEnvironment environment)
+        {
+            OperationManager.SafeExecute(() =>
+            {
+                var gitConfigPath = Path.Combine(environment.RootPath, ".gitconfig");
+                var markerFilePath = Path.Combine(environment.RootPath, "gitsafedirectory.marker");
+
+                if (File.Exists(markerFilePath))
```
shpraka, line 641: should we put it outside SafteExecute?

Kudu.Services.Web/App_Start/NinjectServices.cs
```diff
@@ -628,6 +629,25 @@ private static void RemoveTempFileFromUserDrive(IEnvironment environment)
             FileSystemHelpers.DeleteDirectorySafe(Path.Combine(environment.RootPath, "data", "Temp"), ignoreErrors: true);
         }
 
+        // add/update gitconfig file to be equivalent of running
+        // git config --global --add safe.directory "*"
+        private static void EnsureGitConfigFileWithSafeDirectoryConfig(IEnvironment environment)
+        {
+            OperationManager.SafeExecute(() =>
+            {
+                var gitConfigPath = Path.Combine(environment.RootPath, ".gitconfig");
```
surenderssm, line 638: we can move this after checking for marker file, as gitConfigPath won't be even used if marker file exist.

surenderssm (approved):
Approved with suggestions.

Kudu.Services.Web/App_Start/NinjectServices.cs
sarsharma: moved, thanks!

Kudu.Services.Web/App_Start/NinjectServices.cs
sarsharma: updated, thanks!

Kudu.Services.Web/App_Start/NinjectServices.cs
```diff
@@ -628,6 +629,23 @@ private static void RemoveTempFileFromUserDrive(IEnvironment environment)
...
+        {
+            var markerFilePath = Path.Combine(environment.RootPath, "gitsafedirectory.marker");
+            if (File.Exists(markerFilePath))
+            {
+                return;
+            }
+            OperationManager.SafeExecute(() =>
+            {
+                var gitConfigPath = Path.Combine(environment.RootPath, ".gitconfig");
+                File.AppendAllLines(gitConfigPath, new[] {"[safe]", "    directory = *" });
```
suwatch, line 644: what if customer already have the file with [safe] defined?

Kudu.Services.Web/App_Start/NinjectServices.cs
```diff
@@ -628,6 +629,23 @@ private static void RemoveTempFileFromUserDrive(IEnvironment environment)
             FileSystemHelpers.DeleteDirectorySafe(Path.Combine(environment.RootPath, "data", "Temp"), ignoreErrors: true);
         }
 
+        // add/update gitconfig file to be equivalent of running
+        // git config --global --add safe.directory "*"
+        private static void EnsureGitConfigFileWithSafeDirectoryConfig(IEnvironment environment)
```
suwatch, line 634: for safe rollout, I would hide this under ScmConfigurations (disabled by default).

Kudu.Services.Web/App_Start/NinjectServices.cs
sarsharma: since we are appending this at the end of the file, `[safe] directory=*` takes precedence even if some other `[safe]` value is present in the gitconfig file

Kudu.Services.Web/App_Start/NinjectServices.cs
sarsharma: updated, thanks!

Kudu.Services.Web/App_Start/NinjectServices.cs
sarsharma: Do you see any potential issues with marking all directories as safe? The user has ownership of all the directories and if they have their own `[safe]` configuration added in the gitconfig, their directory list would be a subset of `*` directories, so it should not impact their usage.

suwatch: I think it is fine as long as we tested it (duplicate [safe] sections in gitconfig).   After all it will be behind off-by-default knob.

PR #3467, opened 2023-06-21
Fix OneDeploy --clean option for type zip

dannysongg:
Fixing issue where --clean option would not work for type=zip.


dannysongg:
@jvano

jvano (approved):
looks good, :shipit:

PR #3468, opened 2023-06-22
Fix TestZipDeployment

dannysongg:
Previous PR did not account for clean zip deployments with paths multiple levels deep relative to wwwroot. For example, performing a clean zip deployment to wwwroot/a/b/ would fail. Refactored logic to account for this. 

All tests passing now:
image


dannysongg:
@jvano @suwatch

PR #3477, opened 2023-07-20
Refactor logic for zip and static clean deployments 

dannysongg:
Refactoring logic to simplify the code behind clean deployments. This change will also enable a couple of scenarios that were previously not allowed.

image


dannysongg:
@jvano

jvano (approved):
It would nice to add negative test cases too but it looks good.

dannysongg:
@suwatch this is ready to merge from our end.

PR #3497, opened 2023-12-08
Add terms and conditions for the site extensions installation

SatishRanjan:
Add the following terms and conditions for the site extension installation:
https://www.nuget.org/policies/Terms


jvano (approved):
Looks good!

PR #3508, opened 2024-04-10
Provide a custom error message for concurrent profiling sessions

pharring:
In ANT 102, we introduced a limit on the number of concurrent profiling sessions. The limit is set to 1.
If you try to start a Kudu profiling session when there's already another profilng session going (e.g. the Application Insights Profiler), you'll get an error message in the Kudu portal like:

image

That error message comes from the captured stderr of VSDiagnostics.exe. The HRESULT 0xE111005E is VSHUB_E_MAX_NUMBER_OF_SESSIONS_ALREADY_REACHED which was added to the Standard Collector service specifically for this case.

This change adds detection for that HRESULT and displays a custom error message instead of the generic "Profiling process failed" message.


pharring:
Opened in wrong repo. Sorry.

PR #3509, opened 2024-04-22
Maximum length of Query String increased to 8192

shibayan:
@suwatch  When deploying from GitHub using `azure/webapps-deploy@v3`, if the commit message is long and contains multibyte strings, it will get stuck in the query string length limit and cause an error, so Kudu will relax the limit to handle this.

Since this problem does not occur with the Linux App Service, but only with the Windows App Service, the idea is to adapt to the Linux behavior.

https://github.com/Azure/webapps-deploy/issues/336


suwatch:
@SatishRanjan   could you help with this logistic of merging the change?

